PR #71, opened 2014-09-29
Fix height of burger

robertkowalski:
Do not slice off burger when you scroll up

Screenshot:
size-burger


robertkowalski:
this got fixed by the huge css / js refactor for layouts!

PR #204, opened 2014-12-16
Update full page document editor

benkeen:
Updates to the document editor page. This is what's changed:

- focus placed on editor on first load
- editor now appears full page (no margins)
- buttons now place in a full button row underneath the main header
- header and function button row are now fixed, so large docs will only scroll the editor region of the page.
- notifications moved to global notifications
- editor re-themed for consistency with other pages in Fauxton. 
- delete document confirmation prompt moved to bootstrap dialog
- ace editor now automatically scrolls into view when you click up/down and the available space in the page runs out (right now the cursor just disappears offscreen)
- resizing the window to something small will hide the "API URL" header button so there's no wrapping, and remove the text from the buttons in the button row and just show icons.

To test:
- All functionality on the document editor page (add doc, edit doc, clone, delete, view attachments). N.B. Upload attachments doesn't work locally. 
- The "<" and "cancel" links now take you back to where you were previously (all docs, specific View)
- The Delete Doc is now in a modal.
- All other Ace editors will be styled with the dark theme, like the doc editor page.

### DIFF
--- a/app/addons/documents/assets/less/documents.less
+++ b/app/addons/documents/assets/less/documents.less
@@ -58,21 +58,8 @@ button.beautify {
 
 #map-function, #reduce-function {
   width: 100%;
-  font-size: 16px;
-}
-
-#doc-actions {
-  height: 42px;
-}
-
-#editor-container {
-  height: 688px;
-  width: 100%;
-  font-size: 16px;
-}
-
-.editor-content-page {
-  padding-right: 20px;
+  font-size: 13px;
+  line-height: 22px;
 }
 
 .metadata-page {
@@ -83,8 +70,12 @@ button.string-edit {
   position: absolute;
   padding: 0;
   z-index: 1000;
-  width: 16px;
+  width: 20px;
   left: 22px;
+
+  i.icon {
+    margin-right: 0px;
+  }
 }
 
 button.string-edit[disabled] {
--- a/app/addons/documents/helpers.js
+++ b/app/addons/documents/helpers.js
@@ -1,0 +1,25 @@
+define([
+  'api'
+], function(FauxtonAPI) {
+
+  var Helpers = {};
+
+  Helpers.getPreviousPage = function (database, wasCloned) {
+    var previousPage = database.url('index'), // default to the current database's all_docs page
+        lastPages = FauxtonAPI.router.lastPages;
+
+    if (!wasCloned && lastPages.length >= 2) {
+
+      // if we came from "/new", we don't want to link the user there
+      if (/new$/.test(lastPages[1])) {
+        previousPage = lastPages[0];
+      } else {
+        previousPage = lastPages[1];
+      }
+    }
+
+    return previousPage;
+  };
+
+  return Helpers;
+});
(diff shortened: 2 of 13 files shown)


benkeen:
Hi @SoleilPhan, @seanbarclay - any word on the JSON colors for this? Thanks!

seanbarclay:
@benkeen let's try these for starters:
key: white
value: string: #72cdf4 (teal)
value: numbers: #29be9d (light blue)
value: boolean: #ff6532 (orange)

benkeen:
Thanks @kxepal! I personally don't mind the colours, but this is just a first try as @seanbarclay mentioned. Sean, what do you think of the colours & @kxepal's points? I just committed it so you can download the branch and play around with it.
__________

N.B. so I don't forget, bug found with this ticket:
1. edit a doc
2. Clone it
3. Click back. Some weirdness there.

benkeen:
@michellephung I've done some digging, but I can't think of a decent way to handle the scroll-right scenario you found. It's not like it's a show stopper, but it would be sure nice to fix. Do you mind if I don't include it with this PR, though?

seanbarclay:
The reason behind the dark code editor was to bring in some consistency and familiarity with modern developer code editors (e.g. SublimeText), which leverage a dark, high contrast palette. It helps to differentiate the functionality a bit between normal UI controls and code. The docs and the editor were designed to be dark to be consistent. However, they were also designed with a lot of functionality that hasn't been baked in yet (e.g.. table views). I designed a lot of this close to a year ago and it's just starting to make it's way in , which is a good sign... but there's still a long way to go. 

Also, a few developers that I tested with preferred the dark theme vs. a white or lighter theme. It seemed like a safe bet. 

That being said, we can definitely continue to tweak the contrast on the colors to pop or be toned back. The colors we submitted are meant to help developers differentiate between keys (white) and values such as strings (teal), numbers (light blue) and boolean (orange). I'm not opposed to shifting colors to help with contrast and readability. 

I'll take a look at lightening up that green and post an update soon. 
Cheers

kxepal:
@seanbarclay The reasons behind dark theme are clear. My point was on overall design. If you're uses dark theme for code editor, you also tries to reach two other goals: turn other editor UI into the same dark theme or minimize amount of bright panels. Same problem here: we have nice dark editor for JSON code, but our attention get distracted by large light grey panels above and below it. For view browsing things getting worse since there we have a big burger of dark-light elements (dark left global sidebar, light database sidebar, dark view results table, light pagination and tabs). Design should be consistent whenever light or dark theme is used. And when it is consistent, it looks awesome.

Just my 5 cents (:

michellephung:
@benkeen 
no worries, its just a nice to have :D

benkeen:
Squashed + rebased with master.

benkeen:
Thanks for everyone's input! The colours have been adjusted a little; the header darkened (grey not white) and the highlight row darkened for increased emphasis, as per @kxepal's suggestions.

michellephung:
+1

michellephung:
nice job on this btw. :D

benkeen:
Thanks, @michellephung :)

PR #207, opened 2014-12-18
Add sidebar to compaction page

benkeen:
This ticket adds the sidebar to the Compaction page and refactors
the code a little to start explictly stating what resources and Views
in the documents addon are intended for sharing.

Closes COUCHDB-2404


app/addons/compaction/routes.js
```diff
@@ -11,49 +11,77 @@
...
-      "database/:database/compact": "compaction"
+      'database/:database/compact': 'compaction'
     },
 
-    initialize: function(route, masterLayout, options) {
+    initialize: function (route, masterLayout, options) {
       var databaseName = options[0];
-
-      this.database = this.database || new Databases.Model({id: databaseName});
+      this.database = this.database || new Databases.Model({ id: databaseName });
```
benkeen: Hey @garrensmith, agreed that makes total sense. Do you have an example of either a mixin or inheritance done on the route objects? Not sure how to approach that & I was just wondering if something like this has been done before that I can model it on.

app/addons/documents/shared-resources.js
```diff
@@ -0,0 +1,317 @@
...
+// License for the specific language governing permissions and limitations under
+// the License.
+
+define([
+  'app',
+  'api',
+  'cloudant.pagingcollection'
+], function (app, FauxtonAPI, PagingCollection) {
+
+  var Documents = FauxtonAPI.addon();
```
benkeen: Hey @garrensmith, rats this is more complicated than I first thought. So what I wanted was for the shared resources to be the *subset* of resources in an addon intended to be used by external code. I figured if the addon was defined elsewhere and the shared resources merged into it, the return value from `shared-resources.js` would include other non-shared stuff - which is exactly what I hoped to avoid.

But now that I think about it, no matter how the code is arranged, since the object is passed around by reference, no matter where the addon is first defined, the returned object from `shared-resources.js` will contain the whole shebang. Yegads.

How about this: move the addon definition elsewhere like you suggested, and just live with the fact that the actual returned object from shared-resources contains everything. I think there's still a big win to be had explicitly grouping code by what's intended to be shared or not.

garrensmith: @benkeen yup no easy way around this. I like your idea. I agree with you, showing that its shared, shows good intent. +1

benkeen: I spent some time trying to define the addon elsewhere, but I'm not sure I can get it working.

The logical place to define the addon is in `base.js`, I think, but requireJS kind of inverts everything and makes that impossible. You need to first define the addon in the furthest leaf in the inheritance tree, rather than the root (`base.js`) because all the dependencies build upon it up bit by bit. Since `shared-resources.js` contains the basic building blocks of code - stuff needed all over the place within the addon itself - that ends up being the code that's included first, hence necessarily the place to define the addon.

garrensmith: @benkeen thats fine. Define the addon in shared. I think leave a comment there so we know why its defined there.

benkeen: Great, thanks Garren! I think this ticket is good to go in that case. +1?

app/addons/compaction/routes.js
```diff
@@ -17,29 +17,25 @@ define([
 
        // Modules
        "addons/compaction/views",
-       "addons/databases/resources"
+       "addons/databases/resources",
+  'addons/documents/shared-routes'
```
garrensmith: I think you need to fix this formatting

garrensmith:
+1 looks good.

PR #208, opened 2014-12-18
Couchdb 2483 add cors to fauxton

ocelotpotpie:
This adds CORS into fauxton in the config section, allowing users to enable and disable CORS.


app/addons/cors/templates/cors.html
```diff
@@ -0,0 +1,31 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+License for the specific language governing permissions and limitations under
+the License.
+-->
+
+<header id="cors-header">
+  <p><a href="">Cross-Origin Resource Sharing</a> (CORS) lets you connect to remote servers directly from the browser, so you can host browser-based apps on static pages and talk directly with CouchDB to load your data.</p>
```
robertkowalski: this link is empty

app/addons/cors/views.js
```diff
@@ -0,0 +1,100 @@
...
+          
+        disable_creds.save();
+      }
+    }
+  });
+  
+  CORS.Views = Views;
+
+  return CORS;
+});
```
robertkowalski: newline missing

app/addons/cors/views.js
```diff
@@ -0,0 +1,100 @@
...
+          attribute: "enable_cors",
+          value: "true"
+        });
+        
+        var enable_creds = new CORS.ConfigModel({
+          section: "cors",
+          attribute: "credentials",
+          value: "true"
+        });
+
```
robertkowalski: We are using CamelCase instead of snake_case, `enable_option` and `enable_creds` are snake-cased

robertkowalski:
It is not clear to me if we are using a cors-module with an own view or adding a config section to `/_config` as there is code for both - I think garren also mentioned that

robertkowalski:
please take a look at our [Styleguide](https://github.com/apache/couchdb-fauxton/blob/master/styleguide.md) - we are using single quotes (`'`) and prefix CSS-selectors with `js-` when they interact with JavaScript. To access view-variables we are using `this.$()` .

ocelotpotpie:
@robertkowalski Thanks for the feedback, Robert! I've pushed some changes to match the style guide as you suggested.

robertkowalski:
Cool thank you!

It is not clear to me if we are using a cors-module with an own view or adding a config section to /_config as there is code for both - I think garren also mentioned that: https://github.com/apache/couchdb-fauxton/pull/208/files#r22057673

ocelotpotpie:
@robertkowalski I cleaned up the redundant routes that garren mentioned previously.

The config section does automatically pull in all the config options for a user, so all of the CORS config options will show up in the main config area as well.

benkeen:
Hi @ocelotpotpie, I'm getting a missing file: `origin_domain_table.html` on loading the CORS page. Mind taking a quick look?

ocelotpotpie:
@benkeen Good catch! It looks like I had some weird rule in .gitignore to not commit the templates. Fixed!

benkeen:
Thanks @ocelotpotpie. Some functional problems: 

- checking "Enable CORS" then clicking save doesn't save that setting.
- Adding a new item doesn't refresh the displayed list of CORS domains
- deleting a CORS domain shows the notification that it's been deleted, but the line remains in the page.
- updating an item in the page doesn't update the row - the "Update" button remains. 

Also, two issues I mentioned in my last review are still around:
- the indentation for the sidebar isn't consistent with other pages.
- Going to the Main config page, then to the CORS page doesn't remove the "Add Option" link in the header. It should.

robertkowalski:
the testsuite is red @ocelotpotpie

garrensmith:
@ocelotpotpie could you close this ticket. CORS is merged in.

PR #210, opened 2014-12-31
Keep track of sidebar status

benkeen:
This is a simple usability change: it keeps track of whether the 
user has expanded or shrunk the sidebar so on future refreshes
it doesn't revert to the wrong state.

To test, just shrink the sidebar and refresh the page. Confirm the 
sidebar remains shrunk. Similarly with expanding the sidebar.

### DIFF
--- a/app/addons/fauxton/base.js
+++ b/app/addons/fauxton/base.js
@@ -130,8 +130,10 @@ function(app, FauxtonAPI, Components, ZeroClipboard) {
 
     toggleMenu: function(){
        var $selectorList = $('body');
+      var minimized = !$selectorList.hasClass('closeMenu');
+      this.setState(minimized);
        $selectorList.toggleClass('closeMenu');
-      FauxtonAPI.Events.trigger(FauxtonAPI.constants.EVENTS.BURGER_CLICKED);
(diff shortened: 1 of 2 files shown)


michellephung:
this works +1

benkeen:
@garrensmith, this look okay to you? Checking local storage in main.js is a little new, so I wanted to run it by you before merging.

garrensmith:
@benkeen this is a nice idea. The code needs some neatening up. Rather create an object that hides away the fact that you using localStorage to keep state. The api could be something like:

    SideBarState = {
      isMinimized: function () { .... },
      setState: function () {....}
    };

Then instead of reading the stored state and setting it in main.js rather do that in Fauxton/base.js. I think thats a better place to do that. Most likely in the module initialise method.

benkeen:
Good idea @garrensmith, will do!

benkeen:
Hey @garrensmith, I think it's better now. I didn't end up using a separate object for the state, but instead just made all the changes within the NavBar View, which I think makes sense. It sets tracks it's own state as it's toggled and uses that info when first rendered during `initialize()`. Mind taking a glance?

garrensmith:
+1 That is a nice solution.

PR #211, opened 2014-12-31
Invalid constant name bug fix

benkeen:
This just fixes an invalid constant name. It worked previously because the API published an event of name `undefined` which could be subscribed to.


michellephung:
+1 cool beans

benkeen:
Brilliant, thanks @michellephung!

benkeen:
Merged as a526645901af4f5dfd1b42086c299855f279db92

PR #212, opened 2015-01-05
scrollable sidebar

garrensmith:
Thanks to @robertkowalski this makes the sidebar scrollable. This fixes the issue when there are many design docs in the sidebar and a user cannot access them all.


robertkowalski:
+1

PR #213, opened 2015-01-05
remove sidebar class and set sidebar width in #sidebar-content

garrensmith:
Without this fix the sidebar goes across into the main section like this:

### DIFF
--- a/app/addons/activetasks/assets/less/activetasks.less
+++ b/app/addons/activetasks/assets/less/activetasks.less
@@ -45,7 +45,7 @@
   }
 }
 
-.sidebar .polling-interval {
+.nav.polling-interval {
   li.nav-header {
     padding: 9px 0px 0px 9px;
   }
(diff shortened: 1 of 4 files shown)


robertkowalski:
that fixes the issue, but we have to take care of 

```css
.sidebar .polling-interval {
```

in `activetasks.less`

robertkowalski:
example fix

```patch
diff --git a/app/addons/activetasks/assets/less/activetasks.less b/app/addons/activetasks/assets/less/activetasks.less
index 5acc4bb..68c44a9 100644
--- a/app/addons/activetasks/assets/less/activetasks.less
+++ b/app/addons/activetasks/assets/less/activetasks.less
@@ -45,7 +45,7 @@
   }
 }

-.sidebar .polling-interval {
+.nav.polling-interval {
   li.nav-header {
     padding: 9px 0px 0px 9px;
   }
```

robertkowalski:
+1

garrensmith:
merged.

PR #216, opened 2015-01-08
Adding React into Fauxton

garrensmith:
This is the building blocks for using React in Fauxton. The index editor
and the main side header link are now done in React with a flux pattern.


tasks/couchserver.js
```diff
@@ -45,11 +45,11 @@ module.exports = function (grunt) {
           accept = req.headers.accept.split(','),
           filePath;
 
-      if (setContentSecurityPolicy) {
+          /*if (setContentSecurityPolicy) {
```
robertkowalski: is this intentional?

package.json
```diff
@@ -25,28 +49,6 @@
   "author": "",
   "license": "Apache V2",
   "devDependencies": {
-    "async": "~0.2.6",
-    "grunt": "~0.4.1",
```
robertkowalski: these are all `devDependencies`, we don't need any of these to run Fauxton in production, just to prepare the build.

app/core/store.js
```diff
@@ -0,0 +1,34 @@
...
+// the License.
+
+define([
+  "backbone"
+],
+function(Backbone) {
+
+  var Store = function () {
+    this.initialize.apply(this, arguments);
+    _.bindAll(this);
```
robertkowalski: do we need bindAll?

app/addons/fauxton/components.react.js
```diff
@@ -0,0 +1,170 @@
...
+          "v. ", version
+        )
+      );
+    }
+  });
+
+  var Burger = React.createClass({displayName: "Burger",
+    render: function () {
+      return (
+        React.createElement("div", {className: "burger", onClick: this.toggleMenu},
```
robertkowalski: this seems to be a compiled file?

app/addons/documents/components.react.js
```diff
@@ -0,0 +1,465 @@
...
+  var ToggleButton = React.createClass({displayName: "ToggleButton",
+    getInitialState: function () {
+      return {
+        title: indexEditorStore.getTitle()
+      };
+    },
+
+    render: function() {
+      return (
+        React.createElement("div", {className: "dashboard-upper-menu"},
```
robertkowalski: is this a generated file?

app/addons/documents/routes-documents.js
```diff
@@ -317,9 +318,9 @@ function(app, FauxtonAPI, Documents, Changes, Index, DocEditor, Databases, Resou
 
     viewFn: function (databaseName, ddoc, viewName) {
       var params = this.createParams(),
-          urlParams = params.urlParams,
-          docParams = params.docParams,
-          decodeDdoc = decodeURIComponent(ddoc);
+      urlParams = params.urlParams,
+      docParams = params.docParams,
+      decodeDdoc = decodeURIComponent(ddoc);
```
robertkowalski: the intendation is funky from beginning of the diff until here. please also separate whitespace commits from feature-commits.

app/addons/documents/routes-documents.js
```diff
@@ -453,14 +461,14 @@ function(app, FauxtonAPI, Documents, Changes, Index, DocEditor, Databases, Resou
...
-          pageSize,
-          collection;
+      params = this.createParams(),
+      urlParams = params.urlParams,
+      docParams = params.docParams,
+      ddoc = event.ddoc,
+      defaultPageSize,
+      isLazyInit,
+      pageSize,
+      collection;
```
robertkowalski: intendation

robertkowalski:
great work, very promising! I left some comments from things that i stumbled upon at a first glance. 

The diff is hard to review (+183,439  −1,087) - can you separate everything into 4 commits (or more) after addressing the small things i found?

suggestion:

 - adding react libraries
 - setup react build
 - rewrite navbar
 - rewrite editor

I think this makes reviewing a lot easier and I can give better feedback afterwards.

app/core/store.js
garrensmith: For now this is needed because the methods get called in the dispatcher.

kxepal:
Awesome @garrensmith !

benkeen:
I wonder if we could include the jsx dependency as a grunt module? Without jsx installed it fails to build & it would nice to have npm handle all that for us. There's this guy: https://github.com/yosuke-furukawa/grunt-jsx - haven't tried it.

app/addons/fauxton/actions.js
```diff
@@ -0,0 +1,56 @@
...
+  return {
+    toggleNavbarMenu: function () {
+      FauxtonAPI.dispatch({
+        type: ActionTypes.TOGGLE_NAVBAR_MENU 
+      });
+    },
+
+    addHeaderLink: function (link) {
+      FauxtonAPI.dispatch({
+        type: 'ADD_NAVBAR_LINK',
```
benkeen: I guess this and all the other dispatch types could all move to `/addons/fauxton/actiontypes.js`.

app/addons/fauxton/components.react.jsx
```diff
@@ -0,0 +1,170 @@
...
+          <a href={link.href}>
+            <i className={link.icon + " fonticon "}></i>
+            {link.title}
+          </a>
+        </li>
+      );
+    }
+  });
+
+  var NavBar = React.createClass({
```
benkeen: This all looks awesome. Just one thing that may have got lost in the refactor was the 4 or 5px shadow line that appears at the right of the primary nav.

app/addons/fauxton/components.react.jsx
```diff
@@ -0,0 +1,170 @@
...
+              </ul>
+            </div>
+          </div>
+        </div>
+      );
+    }
+  });
+
+
+  return {
```
benkeen: [Opinion Alert!] I *really* like this pattern of having a return statement at the end of a file that does little more than outline what's being returned, e.g.

```javascript
return {
   Burger: Burger,
   Fries: Fries,
   Coffee: Coffee
};
```

I find it much easier to read at a glance what's available without having to sift through code. We don't do this in Fauxton - we just append to an object as it goes down the file then return the object. Be nice to start doing it...

benkeen:
This all looks really great - very clean! No criticisms are coming to mind. 

One thing to note: the main branch has the nav bar state saved in local storage (https://github.com/apache/couchdb-fauxton/commit/0e8b68edfacdddfcf6d777fc61c605d159fa47b9) which will need to be manually added in here.

robertkowalski:
I found two general points:

 - I am no Flux-expert but reading examples and tutorials I always see the dispatcher separated from the store (in a separate file). Why are we adding the dispatcher as a property to the stores? To me it seems that this leads to the fact the Store and Dispatcher are the same Object/Element.

 - We currently just use chais `assert` in Fauxton tests, this PR mixes `assert` and `expect` - I personally prefer assert but I am also happy if we decide to switch to `expect` - but I think we should keep it consistent and stick to one

app/addons/fauxton/components.react.jsx
garrensmith: Yes I agree. I think we should move to doing it this way going forward.

garrensmith:
@robertkowalski, @benkeen I've fixed all the issues you have mentioned and rebased again master.

robertkowalski:
Starting now functional testing:

bildschirmfoto 2015-01-13 um 14 43 04


the urls of the labels are broken, they say `href="<%-getDocUrl('map_functions')%>"`

robertkowalski:
bildschirmfoto 2015-01-13 um 14 45 12

the animation on create index is broken. click: create a new index and then and click on it

robertkowalski:
The react tests are not unmounting elements, `React.unmountComponentAtNode(mytestedCompontent);`

robertkowalski:
the added sourcefiles need license headers

app/addons/documents/actions.js
```diff
@@ -0,0 +1,164 @@
...
+      return designDocs.find(function (doc) {
+        return doc.id === designDocId;
+      }).dDocModel();
+
+    }
+  };
+
+  return {
+    //helpers are added here for use in testing actions
+    helpers: ActionHelpers,
```
robertkowalski: i have no good feeling when we start to mix production code with testcode. 

i see that the only test that uses it is a spy which get's called, that verifies that another function is called. suggestion: lets test the result, not the steps in between that we just can access by mixing in testcode in our production environment.

app/addons/documents/tests/componentsSpec.react.jsx
```diff
@@ -0,0 +1,199 @@
...
+      newView: false,
+      viewName: 'test-view',
+      designDocs: designDocs,
+      designDocId: designDoc._id
+    });
+  };
+
+  describe('View editor', function () {
+
+    describe('Toggle button', function () {
```
robertkowalski: I haven't figured out why, but this test block is not executed on a testsuite run

robertkowalski: Ah, I guess it is the naming: `componentsSpec.react.jsx`

in our test helper in our gruntfile we define a testfilename as:

```js

    mochaSetup: {
      default: {
        files: { src: helper.watchFiles(['[Ss]pec.js'], ['./app/addons/**/*[Ss]pec.js'])},
```

app/addons/documents/actions.js
garrensmith: Normally I would agree but I cannot how else to test this behaviour without exposing the helpers.

garrensmith:
@robertkowalski I've pushed a new commit with all your feedback. The only thing I haven't done is convert all the expect's to asserts.

app/addons/documents/actions.js
```diff
@@ -0,0 +1,164 @@
...
+    },
+
+    saveView: function (viewInfo, designDocs) {
+      var designDoc;
+
+      if (_.isUndefined(viewInfo.designDocId)) {
+        FauxtonAPI.addNotification({
+          msg:  "Please enter a design doc name.",
+          type: "error",
+          selector: "#define-view .errors-container",
```
robertkowalski: i just found several usages of `selector`, we are now starting to avoid using error containers for everything that has a submit-button, and start using the global notification. like mentioned yesterday in the standup.

app/addons/documents/actions.js
robertkowalski: Suggestion: just move the functions directly to the function literal, example:

```js
  return {
    createNewDesignDoc: function (id, database) {
      var designDoc = {
        _id: id,
        views: {
        }
      };

      return new Documents.Doc(designDoc, {database: database});
    },

    findDesignDoc: function (designDocs, designDocId) {
      return designDocs.find(function (doc) {
        return doc.id === designDocId;
      }).dDocModel();
    },

    toggleEditor: function () {
      FauxtonAPI.dispatch({
        type: ActionTypes.TOGGLE_EDITOR
      });
    },
```

I think you already do the same in line 142 with

```
deleteView: function (options) {
```

robertkowalski: i initially thought it would be code that is just there for the tests, but it is also used in production

benkeen: If we want to differentiate the actions from the helpers, we could maybe just group the helpers in an object. Similarly, putting the actions into an actions property is tempting too, but getting verbose.

```javascript
return {
  helpers: {
    createNewDesignDoc: function() { ... },
    findDesignDoc: function() { ... },
  },
  toggleEditor: function () { ... } ,
  selectReduceChanged: function () { ... } 
  ...
}
```

benkeen:
Small thing. An extra space is getting prepended to the map function each time it's saved. To reproduce:
1. Select any database. 
2. All Design Docs -> New View
3. Note the Map Function field has an extra space at the start
4. Fill in the fields + click save.

Every time you re-edit the index and click save, another space gets added.

benkeen:
Couple of small aesthetic things.

1. When editing the View and clicking the `Edit Index` tab to shrink it, it's smooth up until the last moment when there's a jump as the page shifts up.
2. No space between the green + red buttons? 

screen shot 2015-01-13 at 4 03 58 pm

app/addons/documents/components.react.jsx
```diff
@@ -0,0 +1,473 @@
+define([
+  "app",
```
robertkowalski: missing license header

app/addons/documents/actiontypes.js
```diff
@@ -0,0 +1,24 @@
+// use this file except in compliance with the License. You may obtain a copy of
```
robertkowalski: the license header is cut off

robertkowalski:
double click on a design-doc: sidebar/navbar is not visible and size of the editor is too small

bildschirmfoto 2015-01-14 um 13 50 45

robertkowalski:
there are still error-containers: https://github.com/apache/couchdb-fauxton/pull/216#discussion-diff-22875951

app/addons/documents/actions.js
robertkowalski: to fix, just remove the selector property

garrensmith:
@robertkowalski @benkeen I've fixed the all the issues you both have mentioned. The one thing is I cannot get the animation for the final part any smoother at this stage. The final step can be smoothened out once all tab-panel's are in react.

Could you both give it another round of testing. I think we getting close.

robertkowalski:
The editor partially overlaps the footer:

bildschirmfoto 2015-01-15 um 16 33 03

robertkowalski:
editing the map and then adding a custom reduce will "freeze" both editor windows after saving and also unformat the code: https://cloudup.com/cWQLkqlm2Mx

robertkowalski:
the testsuite is red, at least one test is failing

robertkowalski:
hi garren, this is going to be great. i found three small issues

robertkowalski:
found another one: the footer.html of the navbar subview can get deleted

michellephung:
Hey Garren, 
This is cool. I didn't run into any problems clicking around.

I looked into the failing test. I pushed a solution here:

https://github.com/michellephung/couchdb-fauxton/commit/8f415ed03bcc2d259763bece2a6687ce02bfeeef

These changes will pass the failing test.

garrensmith:
thanks @robertkowalski, I've removed the footer.

sebastianrothbucher:
I do like the idea of factoring out the Actions to make the whole thing more testable - so +1

garrensmith:
I've broken this into two smaller pull requests for easier merging. They are #233 and #235

benkeen:
Looks great. The only thing I spotted was on IE11, editing a View and clicking Save reloads the page but the Edit Index tab section has disappeared. But this is an existing bug that wasn't introduced here.

+1 from me.

garrensmith:
This is not needed anymore.

PR #217, opened 2015-01-08
Fix selenium tests for CI

robertkowalski:
After some fiddling with Jenkins I had the solution for running travis ci.

This PR enables Selenium testing on the travis VM itself, without depending on secrets or ecternal services like saucelabs.

This means that we can test every PR without maintaining a Jenkins with a whitelist for comitters! :)

The browser was switched from chrome to firefox for two reasons: less dependencies we have to maintain, nightwatch/selenium supports it epr default and more importantly travis is supporting Firefox on their VMs out of the box.

Happy testing!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,21 +6,16 @@ git:
 services:
   - couchdb
 addons:
-  sauce_connect: true
+  firefox: "34.0"
 before_install:
+  - export DISPLAY=:99.0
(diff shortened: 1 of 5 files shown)


garrensmith:
This is so awesome :rocket:  :+1:

robertkowalski:
merged!

PR #223, opened 2015-01-09
Constant var rename

benkeen:
This was an old problem that was re-introduced by a bad merge
this morning. The constant var name was just incorrect.


michellephung:
+1

PR #224, opened 2015-01-09
More Selenium fixes

robertkowalski:
Do not click on the div, click directly on the link instead


robertkowalski:
Green! :traffic_light:

benkeen:
+1 fantastic!

robertkowalski:
yay we have CI again!

PR #226, opened 2015-01-13
Moved momentjs to js/libs folder

benkeen:
I placed the moment.js lib in the wrong location in an earlier
commit. This puts it into the libs folder to be with its friends.


ocelotpotpie:
+1 because correct paths are nice.

benkeen:
haha thanks man :)

PR #227, opened 2015-01-14
Include new Ace theme in bundle

benkeen:
The new Ace editor theme needs to be explicitly included in the
generated bundle file.


robertkowalski:
question: do we then still need `assets/js/libs/ace/theme-crimson_editor.js` that's also in that file?

benkeen:
Good question. Yes - it's acting as a default right now. Down the road once all editor locations have been confirmed & updated, it'll be removed.

robertkowalski:
+1

would you create a follow up ticket for the cleanup or should i create one?

benkeen:
I'll create one right now. Thanks @robertkowalski!

robertkowalski:
(just plus +1 if the build is green) :D

benkeen:
Woot! :D

PR #228, opened 2015-01-14
Fix for failing nightwatch test

benkeen:


benkeen:
Merging since Travis confirmed it's now working.

PR #229, opened 2015-01-16
Removes the unused Ace crimson theme

benkeen:
@robertkowalski, you were quite right - I thought there was more regression here, but all other Ace editor references just relied on the default, which I'd changed to the new theme.


robertkowalski:
+1

merge it after the build got green

benkeen:
Merged as 60d29b7

PR #230, opened 2015-01-16
Improvements to doc collapsed state

benkeen:
- This applies to all lists (All Docs / Views)
- Collapsing the list of documents now just shows a single row
for each doc
- The collapsed/expanded state is now stored in memory so doesn't
get lost on subsequent page loads

Closes COUCHDB-2548

### DIFF
--- a/app/addons/documents/resources.js
+++ b/app/addons/documents/resources.js
@@ -424,5 +424,18 @@ function(app, FauxtonAPI, Documents, PagingCollection) {
     }, menuLinks);
   };
 
+  Documents.ListExpandedState = {
+    set: function (isExpanded) {
+      app.utils.localStorageSet('docListCollapsedState', isExpanded);
+    },
+
+    // returns a boolean indicating whether the doc list is expanded or not. Defaults to true if the user
+    // hasn't selected anything
+    get: function () {
+      var docListCollapsedState = app.utils.localStorageGet('docListCollapsedState');
+      return !_.isUndefined(docListCollapsedState) ? docListCollapsedState : true;
+    }
+  };
+
   return Documents;
 });
--- a/app/addons/documents/templates/all_docs_item.html
+++ b/app/addons/documents/templates/all_docs_item.html
@@ -33,8 +33,10 @@ the License.
       </div>
     <% } %>
   </header>
+  <% if (doc.get('expanded')) { %>
   <div class="doc-data">
     <pre class="prettyprint"><%- doc.prettyJSON() %></pre>
   </div>
+  <% } %>
 </div>
 <div class="clearfix"></div>
--- a/app/addons/documents/templates/all_docs_item.html
+++ b/app/addons/documents/templates/all_docs_item.html
@@ -34,9 +34,9 @@ the License.
     <% } %>
   </header>
   <% if (doc.get('expanded')) { %>
-  <div class="doc-data">
-    <pre class="prettyprint"><%- doc.prettyJSON() %></pre>
-  </div>
+    <div class="doc-data">
+      <pre class="prettyprint"><%- doc.prettyJSON() %></pre>
+    </div>
   <% } %>
 </div>
 <div class="clearfix"></div>
--- a/app/addons/documents/tests/nightwatch/expandCollapseRows.js
+++ b/app/addons/documents/tests/nightwatch/expandCollapseRows.js
@@ -19,11 +19,11 @@ module.exports = {
 
     client
       .loginToGUI()
-      .createDocument("doc 1", newDatabaseName)
-      .createDocument("doc 2", newDatabaseName)
-      .createDocument("doc 3", newDatabaseName)
-      .createDocument("doc 4", newDatabaseName)
-      .createDocument("doc 5", newDatabaseName)
+      .createDocument('doc 1', newDatabaseName)
+      .createDocument('doc 2', newDatabaseName)
+      .createDocument('doc 3', newDatabaseName)
+      .createDocument('doc 4', newDatabaseName)
+      .createDocument('doc 5', newDatabaseName)
       .url(baseUrl + '/#/database/' + newDatabaseName + '/_all_docs')
       .waitForElementVisible('#collapse', waitTime, false)
 
@@ -38,11 +38,9 @@ module.exports = {
       .waitForElementNotPresent('.doc-data', waitTime, false)
 
       // lastly, expand a single one and confirm there's only one that's been expanded
-
-      // this is done because we're unable to click() on hidden input, or the :before styled checkbox
-      .execute('$(".js-row-select").first().trigger("click");')
-      .pause(500)
+      .click('.label-checkbox-doclist')
       .click('#collapse')
+      .waitForElementPresent('.doc-data', waitTime, false)
       .source(function (result) {
         var numExpandedDocs = result.value.match(/doc-data/g).length;
         this.verify.ok(numExpandedDocs === 1, 'Checking there is 1 expanded row');
--- a/app/addons/documents/tests/nightwatch/expandCollapseRows.js
+++ b/app/addons/documents/tests/nightwatch/expandCollapseRows.js
@@ -1,0 +1,52 @@
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License. You may obtain a copy of
+// the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+
+module.exports = {
+
+  'Expand/Collapse works on documents': function (client) {
+    var waitTime = 10000,
+      newDatabaseName = client.globals.testDatabaseName,
+      baseUrl = client.globals.test_settings.launch_url;
+
+    client
+      .loginToGUI()
+      .createDocument("doc 1", newDatabaseName)
+      .createDocument("doc 2", newDatabaseName)
+      .createDocument("doc 3", newDatabaseName)
+      .createDocument("doc 4", newDatabaseName)
+      .createDocument("doc 5", newDatabaseName)
+      .url(baseUrl + '/#/database/' + newDatabaseName + '/_all_docs')
+      .waitForElementVisible('#collapse', waitTime, false)
+
+      // confirm all rows are expanded by default (more elegant way to do this?)
+      .source(function (result) {
+        var numExpandedDocs = result.value.match(/doc-data/g).length;
+        this.verify.ok(numExpandedDocs === 5, 'Checking there are 5 expanded rows');
+      })
+
+      // now toggle the items and confirm there are NO expanded items
+      .click('#collapse')
+      .waitForElementNotPresent('.doc-data', waitTime, false)
+
+      // lastly, expand a single one and confirm there's only one that's been expanded
+
+      // this is done because we're unable to click() on hidden input, or the :before styled checkbox
+      .execute('$(".js-row-select").first().trigger("click");')
+      .pause(500)
+      .click('#collapse')
+      .source(function (result) {
+        var numExpandedDocs = result.value.match(/doc-data/g).length;
+        this.verify.ok(numExpandedDocs === 1, 'Checking there is 1 expanded row');
+      })
+      .end();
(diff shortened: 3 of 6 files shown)


benkeen:
Hi @seanbarclay, this updates the collapsed state for the doc list as discussed. Screenshot below. It also now remembers your choice which feels right (defaults to expanded for everyone).

Question: think it's worth changing it so that when clicking Expand/Collapse it only affects selected rows? I guess the one outlier is when no rows are selected. Feels like in that scenario is should toggle everything.

screen shot 2015-01-16 at 12 22 14 pm

garrensmith:
+1 nice work @benkeen

benkeen:
Thanks, @garrensmith! 

Note to self: after reviewing with Sean, be sure to remove the extra `doc-item-expanded` - wasn't needed.

seanbarclay:
Hey @benkeen - looks like if you select an item and click expand/collapse- it will expand or collapse all. It should only expand/collapse selected items. It appears to be acting as an all or nothing, which is a bug. Otherwise, it looks good. 

e.g. 
https://www.dropbox.com/s/4wxigv9pnkm12sm/Screenshot%202015-01-19%2009.18.49.png?dl=0
to 
https://www.dropbox.com/s/sk0m7he7js7zkqr/Screenshot%202015-01-19%2009.19.05.png?dl=0

Notice how the doc in-between and below are expanded as well.

benkeen:
Hi @robertkowalski, mind taking a look at this if you have time tomorrow? I've included a nightwatch test to confirm the base functionality: toggling Expand/Collapse expands/collapses the rows, and selecting a single row toggles just that one.

app/addons/documents/views.js
```diff
@@ -505,21 +517,37 @@ function(app, FauxtonAPI, Components, Documents, Databases, Views, QueryOptions)
...
 
     beforeRender: function() {
-      var docs;
-
       this.removeNestedViews();
 
       this.pagination.setCollection(this.collection);
       this.allDocsNumber.setCollection(this.collection);
 
-      docs = this.expandDocs ? this.collection : this.collection.simple();
```
robertkowalski: don't we need the simple collection any more? can we delete it? for what was it for?

app/addons/documents/views.js
```diff
@@ -453,25 +458,32 @@ function(app, FauxtonAPI, Components, Documents, Databases, Views, QueryOptions)
...
 
       this.toggleTrash();
+      this.updateExpandButtonLabel();
     },
 
     serialize: function() {
       return {
         viewList: this.viewList,
-        expandDocs: this.expandDocs,
+        expandDocs: true, // TODO
```
robertkowalski: this seems to be still an open todo

robertkowalski:
@benkeen I will take a look in detail tomorrow but I just found two things that jumped into my eyes when scrolling over the code

app/addons/documents/views.js
benkeen: No it's no longer needed. It used to handle showing a subset of the data (id + rev only). Now nothing is shown - just the ID row.

app/addons/documents/views.js
benkeen: Good catch! I'll remove it.

benkeen:
Thanks, @robertkowalski :)

app/addons/documents/views.js
robertkowalski: i think we can delete it then, i searched in the code for it and it was the only usage i think (you might want to double check)

benkeen: Yup! I did remove it: see line 259 of app/addons/documents/shared-resources.js

app/addons/documents/tests/nightwatch/expandCollapseRows.js
```diff
@@ -0,0 +1,52 @@
...
+
+      // this is done because we're unable to click() on hidden input, or the :before styled checkbox
+      .execute('$(".js-row-select").first().trigger("click");')
+      .pause(500)
+      .click('#collapse')
+      .source(function (result) {
+        var numExpandedDocs = result.value.match(/doc-data/g).length;
+        this.verify.ok(numExpandedDocs === 1, 'Checking there is 1 expanded row');
+      })
+      .end();
```
robertkowalski: you can use the label, as it is made for clicking on it - you may want to add some wait-calls `waitElement*` for some classes to appear that this example also runs on slow machines

```javascript
      // now toggle the items and confirm there are NO expanded items
      .click('#collapse')
      .waitForElementNotPresent('.doc-data', waitTime, false)

      // lastly, expand a single one and confirm there's only one that's been expanded
      .click('.label-checkbox-doclist')
      .click('#collapse')
      .source(function (result) {
        var numExpandedDocs = result.value.match(/doc-data/g).length;
        this.verify.ok(numExpandedDocs === 1, 'Checking there is 1 expanded row');
      })
      .end();
```

app/addons/documents/tests/nightwatch/expandCollapseRows.js
```diff
@@ -0,0 +1,52 @@
...
+      newDatabaseName = client.globals.testDatabaseName,
+      baseUrl = client.globals.test_settings.launch_url;
+
+    client
+      .loginToGUI()
+      .createDocument("doc 1", newDatabaseName)
+      .createDocument("doc 2", newDatabaseName)
+      .createDocument("doc 3", newDatabaseName)
+      .createDocument("doc 4", newDatabaseName)
+      .createDocument("doc 5", newDatabaseName)
```
robertkowalski: we use single quotes

app/addons/documents/templates/all_docs_item.html
```diff
@@ -33,8 +33,10 @@
       </div>
     <% } %>
   </header>
+  <% if (doc.get('expanded')) { %>
```
robertkowalski: intendation, please move the block inside the if to the right

app/addons/documents/resources.js
```diff
@@ -424,5 +424,18 @@ function(app, FauxtonAPI, Documents, PagingCollection) {
...
+  Documents.ListExpandedState = {
+    set: function (isExpanded) {
+      app.utils.localStorageSet('docListCollapsedState', isExpanded);
+    },
+
+    // returns a boolean indicating whether the doc list is expanded or not. Defaults to true if the user
+    // hasn't selected anything
+    get: function () {
+      var docListCollapsedState = app.utils.localStorageGet('docListCollapsedState');
+      return !_.isUndefined(docListCollapsedState) ? docListCollapsedState : true;
```
robertkowalski: clever and thanks for the comment!

app/addons/documents/views.js
```diff
@@ -453,25 +458,31 @@ function(app, FauxtonAPI, Components, Documents, Databases, Views, QueryOptions)
...
     },
 
-    collapse: function (event) {
-      event.preventDefault();
+    toggleExpandCollapse: function (e) {
+      e.preventDefault();
+      var expand = this.$(e.target).find("i").hasClass('icon-plus');
 
-      if (this.expandDocs) {
-        this.expandDocs = false;
```
robertkowalski: this will get changed back again once the headerbar in react is merged - we can't use the dom for storing state there (which is a good thing) and depend on `this.expandDocs`

is there a special reason to switch to the usage of classes? We can decide to use the `this.expandDocs` here again or we will have to change that when one of us is having a huge merge conflict and maybe there are reasons to use the class that would introduce a bug using the old `this.expandDocs`

app/addons/documents/views.js
```diff
@@ -505,21 +516,37 @@ function(app, FauxtonAPI, Components, Documents, Databases, Views, QueryOptions)
...
+      if (this.collection.length) {
+        var firstItem = this.collection.first();
+        isNewPage = _.isUndefined(firstItem.get('expanded'));
+        if (this.numSelectedOnPage() === this.collection.length) {
+          allSelected = true;
+        }
+      }
 
-      docs.each(function(doc) {
+      this.collection.each(function(doc) {
```
robertkowalski: space missing :space_invader:

app/addons/documents/views.js
benkeen: Yeah... I kinda agree. I'll change this.

One thing though: I never liked the simple toggling of a boolean to keep track of something - I've found it can lead to problems when there's disconnect between what's in the var, and the reality of what's actually in the DOM. But I guess this'll be eliminated once we're using react state, which will have a 1-1 relation.

app/addons/documents/tests/nightwatch/expandCollapseRows.js
benkeen: Huh! So click() on a class will just click the first element with the class? That's super handy.

benkeen:
Great feedback, @robertkowalski! Adding fixes now.

app/addons/documents/views.js
```diff
@@ -564,15 +591,79 @@ function(app, FauxtonAPI, Components, Documents, Databases, Views, QueryOptions)
...
+        }
+      } else {
+        this.setExpandButtonLabel(true); // default show "Expand"
+
+        // show "Collapse" if ALL checked rows have state === expanded (true)
+        if (this.numSelectedOnPage() && this.allCheckedRowsHaveState(true)) {
+          this.setExpandButtonLabel(false);
+        }
+      }
+    },
```
robertkowalski: this one is quite complex code - it seems it has a quite high cyclomatic complexity - i would suggest using early returns so the if / else does not get so deeply nested:

```js
    updateExpandButtonLabel: function () {
      this.setExpandButtonLabel(true); // default show "Expand"
      
     // show "Collapse" if ALL checked rows have state === expanded (true)
      if (!this.defaultDocsStateExpanded && this.numSelectedOnPage() && this.allCheckedRowsHaveState(true)) {
        this.setExpandButtonLabel(false);
        return;
      }

      // override the default label and show "Expand" if ALL checked rows have state === collapse (false)
      if (this.numSelectedOnPage() && this.allCheckedRowsHaveState(false)) {
        this.setExpandButtonLabel(true);
      }
      return;

      this.setExpandButtonLabel(false);
    },

```

app/addons/documents/views.js
```diff
@@ -564,15 +591,79 @@ function(app, FauxtonAPI, Components, Documents, Databases, Views, QueryOptions)
...
+
+    setExpandButtonLabel: function(expanded) {
+      if (expanded) {
+        this.$('#collapse').html('<i class="icon-plus"></i> Expand');
```
robertkowalski: we will also have to delete that code for the headerbar - sorry! :(

app/addons/documents/views.js
```diff
@@ -564,15 +591,79 @@ function(app, FauxtonAPI, Components, Documents, Databases, Views, QueryOptions)
...
+
+    // returns the models of selected items in the current page
+    getSelectedRows: function () {
+      if (!this.bulkDeleteDocsCollection) {
+        return [];
+      }
+      var allSelectedIDs = this.bulkDeleteDocsCollection.pluck('_id');
+      return _.filter(this.collection.models, function(item) {
+        return _.contains(allSelectedIDs, item.id);
+      }, this);
```
robertkowalski: I think we could use http://underscorejs.org/#some but maybe I am missing something

app/addons/documents/views.js
benkeen: It is a bit dense, yeah. I'll check over your code (second return should be in the `if`, presumably?)

app/addons/documents/views.js
benkeen: What do you suggest here? So I removed the actual rendering of the #collapse element entirely out of the template and into this function here. I could change it to use a View instead (much nicer)? But I do need it to be separate from the rest of the template. 

Honestly, w.r.t. React, I'm going to want to refactor a LOT of this code, so I wasn't too worried about

robertkowalski: i mentioned you in https://github.com/robertkowalski/couchdb-fauxton/commit/d92f759960bbae0014b1f2b24fb9ec827ee05f3b at several places to show you what is going to change

app/addons/documents/tests/nightwatch/expandCollapseRows.js
robertkowalski: you could also do

```
.click('label[for="checkbox-' + newDocumentName + '"]')
```

to use a more specific selector

benkeen:
This ticket didn't end getting merged due to conflicts with other work going on with the All Docs header, and is now out of date. I gather @garrensmith will be reactifying the All Docs page content - could this be included in that? 

I'll leave this ticket open for now.

benkeen:
This PR was largely covered by Garren's ticket. I'll shortly commit a separate PR to fully collapse the docs.

PR #233, opened 2015-01-19
React building blocks and navbar

garrensmith:
I've decided to break up #216 into two. This is the first one. It is the main react dependancies along with the main navigation bar done in React.js

PR #235, opened 2015-01-19
React index editor

garrensmith:
This is part two of #216. This is the index editor. This required #233 to land first.

### DIFF
--- a/app/addons/databases/tests/nightwatch/createsView.js
+++ b/app/addons/databases/tests/nightwatch/createsView.js
@@ -48,7 +48,9 @@ module.exports = {
       .click('#nav-header-test_design_doc .dropdown-toggle.icon.fonticon-plus-circled')
       .waitForElementPresent('#nav-header-test_design_doc', waitTime, false)
       .click('#nav-header-test_design_doc a[href="#/database/'+newDatabaseName+'/new_view/test_design_doc"]')
-      .verify.valueContains('#index-name','newView')
+      .waitForElementPresent('#db-views-tabs-nav', waitTime, false)
+      .click('#db-views-tabs-nav')
+      .verify.valueContains('#index-name','new-view')
       .clearValue('#index-name')
       .setValue('#index-name','odd_ids')
       .execute('\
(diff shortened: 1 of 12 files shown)


robertkowalski:
hi garren can you rebase?

garrensmith:
@robertkowalski @benkeen @michellephung this has been rebased. I would appreciate it if you could take another look.

robertkowalski:
i found a small issue when using the editor: https://cloudup.com/cpHQFd-zrAz

robertkowalski:
hi garren i found some small things, the branch looks already really really great!

benkeen:
The code looks totally wicked.

Minor stuff:
- the Edit Index tab should have cursor:pointer
- comparing the Edit Index tab with other tabs (say Changes > Filter) it’s not quite the same. That one has a mouseover state (red) and is grey by default unless clicked. Not sure what @seanbarclay wants.
- I don't suppose "Edit Index" could read "Edit View" instead? It’s listed under Views in the sidebar, and the delete alert() says “do you want to delete this View?”. I've always found the duplicate terms confusing - now I realize index === view.
- there should be more space after the end of the following labels + the “?” icon: “Save to Design Document", “Map function”

Probably unrelated:
- after logging in and going to the Database page I got a blank, disabled page nav << 1 >> at the bottom right, even though there’s only a single page. Shouldn’t that be omitted if there’s only one page?

michellephung:
everything looks okay Garren,
the only thing i was going to say is the space between the circle question marks, which ben already mentioned. 

@benkeen the disable page nav on one page is happening on the live site too. I'll make a ticket for it. I think I saw it being disabled before.

garrensmith:
thanks @benkeen and @michellephung. I've made some of those fixes.

robertkowalski:
Seems we lost the compact view button:

bildschirmfoto 2015-01-27 um 16 36 13

robertkowalski:
currently just the reduce editor has a button for beautifying the code

robertkowalski:
discussed the topic with garren, there will be follow up tickets as the both are currently broken on master, too

+1

garrensmith:
Merged :rocket:

PR #238, opened 2015-01-20
Cleanup: delete unused less & more specific selectors

robertkowalski:
- delete not used less
 - make css more specific

### DIFF
--- a/assets/less/fauxton.less
+++ b/assets/less/fauxton.less
@@ -164,7 +164,7 @@ table.databases {
 }
 
 // customize the twitter bootstrap alert
-.alert {
+.global-notification.alert {
   padding-left: 20px;
   padding-top: 20px;
   min-height: @collapsedNavWidth;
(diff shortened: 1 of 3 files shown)


assets/less/fauxton.less
```diff
@@ -185,15 +185,32 @@ table.databases {
   }
 }
 
-.notification-icon {
-  padding-right: 8px;
+.global-notification.alert-success {
```
garrensmith: With less you could put all the alert info in the `global-notifications` and less will flatten it for you. Like this:

    .global-notification {
      alert-status {}
      alert-info
    }

garrensmith:
This looks good. I've added a comment on less style. But otherwise I'm +1

garrensmith:
+1 nice clean up.

robertkowalski:
a7a391ebf2f264359b1843985dfe1df382bf898c

PR #239, opened 2015-01-20
transform jsx on release

robertkowalski:
transform the jsx in case `grunt dev` wasn't run before,
e.g. when building couchdb using `make


garrensmith:
+1. Could we also do a `grunt release` in travis so we know that doesn't fail.

robertkowalski:
merged as 118aea1e753d3431ba5f8f609df7a206535d7448

PR #242, opened 2015-01-21
Create Header-Control-Bar in React

robertkowalski:
adds a bar with controls for interaction with documents


seanbarclay:
Hey @robertkowalski - nice job. I'm super excited to see this plugged in. There's a few minor tweaks needed here:
1) The padding between the select icon and copy should be 8px. It currently has none. 
2) Make sure the padding on the select button group and the other select options have a padding that equals 20px on left and right. It's currently 17px (5px + 12px). Check out the screen shot of what spacing should look like when tweaked: https://www.dropbox.com/s/8j83rnqyjfrctqt/Screenshot%202015-01-21%2023.06.51.png?dl=0
3) Font-size for icons should be 20px
4) cancel button does not require a vertical rule line on the right
5) add a quick (250 ms) fade in/out transition for menus.  You could also have the .alternative-header slide in from left to right and fade in. If you do the latter, add a background-color to the select button.

Thanks!

robertkowalski:
@seanbarclay fading both header leads to space issues on small screens, so i decided to just fade the alternative header for now. i created a video that shows the issue: https://cloudup.com/c9kWmUm7_Km - the other header is pushed down as both elements are fading but are taking space.

we could throw a ton of code on the problem but it would make the css and code quite complex.

what do you think?

app/addons/documents/header/header.actions.js
```diff
@@ -0,0 +1,62 @@
...
+],
+function (app, FauxtonAPI, ActionTypes) {
+
+  return {
+    toggleCollapseDocuments: function () {
+      FauxtonAPI.dispatch({
+        type: ActionTypes.COLLAPSE_DOCUMENTS
+      });
+
+      FauxtonAPI.Events.trigger('headerbar:collapse');
```
garrensmith: Can you make all of the events constants.

garrensmith:
@robertkowalski this is coming along nicely. The code is looking excellent

app/addons/documents/header/header.actions.js
robertkowalski: I asked Ben regarding that because I sometimes see that they are added to the constants file and sometimes not:

"Yeah. So here's my rationale: events that are relevant script-wide (stuff in /addons/fauxton) for example should be there. Stuff that's addon-specific shouldn't."

seanbarclay:
@robertkowalski really nice work touching this up. 

The only thing I'm noticing is that if you have all docs deselected, then you click on one or more docs in the list, nothing happens if you hit deselect. 

Same thing goes if you have all docs in the list selected, then you deselect one and hit select all. It doesn't re-select the one unselected doc. 

Lastly, "Select All" and "Deselect All" indicator colors should only be turned on if All or None of the docs are selected. For instance, if all docs were selected, "Select All" could be indicated as active (red). However, if you deselect >=1 docs in the list, then it should no longer be a true statement and therefore "Select All" should not be active, it should be back to normal and ready to toggle back on if clicked. 

Ping me if you need further help there.

app/addons/documents/header/header.actions.js
garrensmith: ok cool that makes sense.

garrensmith:
@robertkowalski @seanbarclay do we need both Select All and Deselect All buttons displayed at all times? Wouldn't it make sense that initially the user only sees the Select All button. Then if they click on that then the Select All button disappears and the Deselect All button appears. Same for collapse and expand.

app/addons/documents/header/header.react.jsx
```diff
@@ -0,0 +1,239 @@
...
+          onClick={this.props.toggleCallback}
+          className={containerClasses}
+          >
+          <i className={iconClasses}></i><span>{this.props.text}</span>
+        </button>
+      );
+    }
+  });
+
+  var AlternateHeaderControlBarController = React.createClass({
```
garrensmith: I think we should rename the `AlternativeHeader` to something like `BulkDoumentHeader`. Since when that header is displayed we editing the documents in bulk. Its a little more descriptive than just `AlternativeHeader`

app/addons/documents/header/header.react.jsx
```diff
@@ -0,0 +1,245 @@
...
+
+    deleteSelected: function () {
+      Actions.deleteSelected();
+    }
+  });
+
+  var HeaderBarController = React.createClass({
+    getStoreState: function () {
+      return {
+        isToggled: headerBarStore.getStatus(),
```
garrensmith: Pretty minor thing. But `getStatus` doesn't mean much. Maybe make it `getToggleStatus`. That explains it more.

garrensmith:
This is looking good. Nice tests. +1

robertkowalski:
merged as f5f7f4b93a2ac9f63b21d9a0a5359285e6f2da38 and 25dab477415d417019cf863db393470b11e5d350

PR #243, opened 2015-01-21
Full page doc editor fixes

benkeen:
It was noticed that with larger documents we needed the horizontal
scrollbar to appear, since some users don't have right-left
scrolling mice/trackpads and relying exclusively on the keyboard
wasn't sufficient. This adds the horizontal scrollbar only when
needed, and at the very bottom of the screen, regardless of the
height of the editor, to make it look ok.

### DIFF
--- a/app/addons/documents/templates/code_editor.html
+++ b/app/addons/documents/templates/code_editor.html
@@ -65,7 +65,7 @@ the License.
   </div>
 </div>
 
-<div class="scrollable">
+<div class="code-region">
   <div class="bgEditorGutter"></div>
   <div id="editor-container" class="doc-code"><%- JSON.stringify(doc.attributes, null, "  ") %></div>
 </div>
--- a/app/addons/documents/views-doceditor.js
+++ b/app/addons/documents/views-doceditor.js
@@ -207,7 +207,7 @@ function (app, FauxtonAPI, Components, Documents, Databases, prettify) {
       'click button.upload': 'upload',
       'click button.string-edit': 'stringEditing',
       'click a.js-back': 'onClickGoBack',
-      'click .scrollable': 'focusOnLastLine'
+      'click .code-region': 'focusOnLastLine'
     },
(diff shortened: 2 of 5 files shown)


benkeen:
Hi @robertkowalski, this is from our discussion today. Turns out you were quite right about the `maxLines` option - it just determined the maximum visible lines in the doc, not the number of rows that could be inputted. So this refactor includes it.

benkeen:
N.B. I also snuck in a (*terrible*) fix for https://issues.apache.org/jira/browse/COUCHDB-2555?filter=12329487 The editor needs to know when the primary nav is shrunk/expanded to resize itself. I'm all ears on better solutions while we're porting stuff over to React.

assets/js/libs/ace/ace.js
```diff
@@ -12127,7 +12127,7 @@ top: 0;\
 .ace_scrollbar-h {\
 position: absolute;\
 overflow-x: auto;\
-overflow-y: hidden;\
+overflow-y: auto;\
```
robertkowalski: we should never patch external dependencies directly. for example it leads to external dependencies that are not upgradable to the next version as we would lose our custom changes we made directly in the libraries.

robertkowalski:
garren also created a fix for 2555

assets/js/libs/ace/ace.js
benkeen: Absolutely. My mistake, I didn't realize I was making this change here. I'll override this in the CSS. Well spotted!

benkeen:
Thanks @robertkowalski. All fixed:
- Rebased with master to include 2555 fix. Custom fix in this branch has been removed.
- leaky resize() fixed
- override of ace.js CSS line removed

robertkowalski:
looks nice! tests are also green - +1

benkeen:
Great, thanks @robertkowalski :)

PR #244, opened 2015-01-22
Fix for compaction lookahead tray not redirecting

benkeen:
I noticed the lookahead tray opens but doesn't redirect properly
on the Compaction page. This is the fix.


garrensmith:
+1 works for me.

benkeen:
Merged as 6236556ea36d13eef5274f89803163759afaabff

PR #245, opened 2015-01-22
Fire event when navbar size changes

garrensmith:


garrensmith:
This also fixes https://issues.apache.org/jira/browse/COUCHDB-2555

robertkowalski:
I could not find out what it fixes exactly - both the commit message and the linked ticket does not describe what behaviour exactly is currently wrong (e.g. how can i see the bug) and which behaviour is expected

robertkowalski:
+1

benkeen:
+1

PR #247, opened 2015-01-23
Fix for missing Edit Map Function

benkeen:
The recent full page doc editor changes caused the Map Function
section to disappear from the Edit Index page. This fixes it
by only enabling the auto-adjust height code relevant for the
full page doc editor for that particular instance, via the
addition of an isFullPageEditor boolean.


benkeen:
Before: 

screen shot 2015-01-23 at 12 26 56 pm

After: 

screen shot 2015-01-23 at 12 26 38 pm

michellephung:
+1 works for me

benkeen:
Thanks, @michellephung. Nightwatch tests pass locally so I'm going to merge now.

PR #249, opened 2015-01-26
changes: always remove filter tab

robertkowalski:
remove filtertab when switching from changes to "All Documents"


michellephung:
+1, cool this fixes the problem clicking from changes -> permissions as well. :)

robertkowalski:
merged as cddf2ef042eacb13e1ff4e8457ad9488e84dff67 :)

PR #251, opened 2015-01-28
header-control-bar: fix select-all-state

robertkowalski:
When I select all documents
And there are less documents than I have set to display per page
Then I want the select-all-button to be selected/disabled

As I checked for the setting of the page compared to the selected
documents it can happen that if we have not enough documents to
show on the page that the status of the select-all was wrong.

Added test to avoid regressions in the future and to prove my fix

### DIFF
--- a/app/addons/documents/header/header.stores.js
+++ b/app/addons/documents/header/header.stores.js
@@ -25,8 +25,6 @@ function (FauxtonAPI, ActionTypes) {
 
     reset: function () {
       this._collapsedDocuments = false;
-      this._selectedAllDocuments = false;
-
       this._selectedDocumentsCount = 0;
       this._documentsOnPageCount = FauxtonAPI.constants.MISC.DEFAULT_PAGE_SIZE;
     },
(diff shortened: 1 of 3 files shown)


michellephung:
+1

To test: 
On master, if you have database with 8 documents, and show 5 documents at a time, view the second page (doc 6-8),  and hit 'select all', notice that the 'select all' and 'unselect all' are both grayed out.

When on pr/251, this doesn't happen. 

On another note: the highlight select states are a bit confusing. 
I realize that a highlight state in the UI means that that current button is being shown, but a greyed out button makes me think that a button is disabled, not that that it is a selectable button.  The only thing I can think of that won't involve more colors or lot of ui changes would be to change the color on hover. Any thoughts?

robertkowalski:
@michellephung let's talk together with sean next standup regardign the colors!

robertkowalski:
merged as 18bd88f2fcea4d3aa83a8b55b62a6d9f2763feee

PR #253, opened 2015-01-30
all-dbs, all-docs: more responsiveness

robertkowalski:
don't display search-label if screensize is too small.

before that the searchlabel jumped out of the headerbar

### DIFF
--- a/assets/less/fauxton.less
+++ b/assets/less/fauxton.less
@@ -771,3 +771,44 @@ footer.pagination-footer {
   z-index: 6;
 }
 
+// left navigationbar is opened
+@media (max-width: 780px) {
+  .closeMenu {
+    .one-pane {
+      .searchbox-wrapper {
+        display: none;
+      }
+    }
+  }
+}
+
+@media (max-width: 1050px) {
+  .closeMenu {
+    .with-sidebar {
+      .searchbox-wrapper {
+        display: none;
+      }
+    }
+  }
+}
+
+// left navigationbar is closed
+@media (max-width: 925px) {
+  body:not(.closeMenu) {
+    .one-pane {
+      .searchbox-wrapper {
+        display: none;
+      }
+    }
+  }
+}
+
+@media (max-width: 1225px) {
+  body:not(.closeMenu) {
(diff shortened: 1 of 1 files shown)


robertkowalski:
bildschirmfoto 2015-01-30 um 13 45 25

michellephung:
Hey Robert,
I'm still getting some header-css-brokenness at smaller widths, and the search bar pops back into view at a certain point.

Some sites automatically collapse menu bars when they reach a certain width. 

Suggestions: 
1. We could collapse the left most nav when the screen is below a certain width. 
2. Also maybe when the width gets really small, collapse the api url/ add new database/ search buttons down to icon-only (like in buttons the code editor).

screen shot 2015-01-30 at 11 10 05 am
screen shot 2015-01-30 at 11 10 13 am

2a.(For search) we could have a pop out tray to search like the API URL / add DB buttons. 
3. This might also be a good time to make the API URL pop out smaller/ responsive.

robertkowalski:
Those are great ideas and suggestions!

Sadly I don't have the time to make Fauxton fully responsive and can just fix the most major annoying bugs.

For a full responsive Fauxton I think we would need:

 - throw out the current not maintained Bootstrap 2 which is mainly just used for button styling at the moment
 - make use of an actively maintained bootstrap and use it's helper to create responsive grids

I documented this together with your ideas in Jira: https://issues.apache.org/jira/browse/COUCHDB-2565

Can you describe the issue you have regarding the database-searchlabel or create an video?

michellephung:
just at smaller widths, I'm still getting the search for document tray pop out. 
screen shot 2015-02-03 at 2 14 20 pm

robertkowalski:
@michellephung i was not able to reproduce it... which browser and at which screensize does it happen?

robertkowalski:
the issue happened if the size got smaller than 825 px

michellephung:
+1

robertkowalski:
faster, travis, faster!

robertkowalski:
merged as cfd9ec85913879069d0294779a1eb6b700026b2a

PR #261, opened 2015-02-08
add a warning for deletion of a  systemdatabase on a cluster.

Poltergeist:
Add a warning if the current database is an internal cluster database
and _stats is available on that note. _stats is only available if
it is a cluster internal node.
The warning is only shown for databases with the name:
  - _replicator
  - _users
  - dbs
  - nodes


garrensmith:
@Poltergeist @robertkowalski this looks great. Could you add a few tests for this.

Poltergeist:
@garrensmith I'll add tests asap in the next days.

Poltergeist:
Added test cases for the parse function of the new model.

app/addons/documents/views.js
```diff
@@ -145,9 +145,16 @@ function (app, FauxtonAPI, Components, Documents,
     template: "addons/documents/templates/delete_database_modal",
     initialize: function (options) {
       this.database = options.database;
+      this.isSystemDatabase = options.isSystemDatabase;
       FauxtonAPI.Events.on('database:delete', this.showDeleteDatabase, this);
     },
 
+    serialize: function () {
+      return {
+        isSystemDatabase: this.isSystemDatabase
```
garrensmith: This serialize function also needs to return database. You need to make the return this:

    return {
      isSystemDatabase: this.isSystemDatabase,
      database: this.database
    };

garrensmith:
@Poltergeist there is a slight bug. When I try delete a database the database name doesn't appear in the modal. See screenshot. I've added a comment on how to fix it. Other than that it is looking good. Once the bug is fixed we can merge

Poltergeist:
@garrensmith added the test and also merged the master branch into this branch.

garrensmith:
Thanks for this @Poltergeist this is great. I've merged it in.

kxepal:
cassim database is missed in the list.

Poltergeist:
I'll add the cassim database. I just didn't know about the cassim database.

kxepal:
No worries. Nobody known (: Thought it might be also good idea to make all system databases with leading `_` to make sure.

Poltergeist:
I added the cassim database in another pull request. Test for the _ prefixed names could be a good idea.

kxepal:
@Poltergeist [COUCHDB-2619](https://issues.apache.org/jira/browse/COUCHDB-2619) fixed so all system databases are starts with underscore no matter on which iface they are. Could you please submit new PR? I'd tried by my own, but I don't understand what is going on in `IsSystemDatabaseModel`, why there is need to request `_stats` (which I strongly believe causes [COUCHDB-2607](https://issues.apache.org/jira/browse/COUCHDB-2607) ) when you could just GET app.host and look if uuid is in the response and how to turn it into simple property of Database model without ruining everything. Thanks (:

robertkowalski:
@Poltergeist @kxepal will take care of that now!

PR #262, opened 2015-02-10
Fix for selected text in Ace editor

benkeen:
This updates the background color and the location of the selected
text bg color.

Closes COUCHDB-2573

### DIFF
--- a/assets/js/libs/ace/theme-idle_fingers.js
+++ b/assets/js/libs/ace/theme-idle_fingers.js
@@ -18,9 +18,7 @@ color: #FFFFFF\
 color: #ffffff\
 }\
 .ace-idle-fingers .ace_marker-layer .ace_selection {\
-background: #000000;\
-opacity: 0.4;\
-left: 0px !important;\
+background: #00281f;\
 }\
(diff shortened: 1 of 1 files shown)


robertkowalski:
Hi Ben,

we should never made any changes in external dependencies. If we have to we have to submit a PR to them and then pulling it down after they merged. I also noticed that you already patched internal libraries.

I assigned https://issues.apache.org/jira/browse/COUCHDB-2560 to you, please revert the previous change. Probably that also fixes this issue we currently have. 

If not, you can create an CSS selector in "our" less which is stronger than the one from ace. I remember that you already did that some time ago in `codeeditor.less`

Patching external libraries makes updating them impossible as nobody can tell what happens when we run an update. Please stop doing so - it makes updates unpredictable for the person that updates the external library.

benkeen:
Apologies @robertkowalski, you're quite right of course - I keep forgetting this particular file is part of that lib. I'll do that ticket today. Thanks.

benkeen:
Closing. I'll create a separate ticket that undoes any changes to that file and moves these changes + the previous to our own CSS>

PR #232, opened 2015-01-18
Fix COUCHDB-2543 by avoiding overlapping callbacks for two editors

sebastianrothbucher:
Effectively, I suffix the editor ID when registering for the callbacks. And when doing that (need to clean the "edited" state when hiding the modal) I can also handle the unnerving bug that one can close the string editor without warning and loose all changes.

PR #263, opened 2015-02-10
Ensure save changes conf appears

benkeen:
Extra eyes on this ticket are appreciated! It's a very small change
but there's a bit of explanation needed.

Whenever you make a change to content in an Ace editor then
navigate away, you should see a confirmation alert saying "Are you
sure you don't want to save these changes?". Right now this
appears sometimes, but not always. This ticket makes a small
change to the main event delegation function in main.js that runs
on all `<a>` tags. Before, as long as the `<a>` didn't contain a
`data-bypass` attribute, it did a `FauxtonAPI.navigate` call, which
meant that all beforeUnload functions got ran prior to redirecting: 
and that was where the confirmation alert code got ran.

This has now been changed to only NOT run when the `data-bypass`
attribute is set to `true`, which I think is what was originally
intended. This small change will ensure any pages containing the
Ace editor will have their beforeUnload functions ran prior to
redirecting from any link.

Closes COUCHDB-2574

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -27,9 +27,9 @@ function(app, FauxtonAPI, LoadAddons) {
 
 
   // All navigation that is relative should be passed through the navigate
-  // method, to be processed by the router. If the link has a `data-bypass`
+  // method, to be processed by the router. If the link has a `data-bypass=true`
   // attribute, bypass the delegation completely.
-  $(document).on("click", "a:not([data-bypass])", function(evt) {
+  $(document).on("click", "a:not([data-bypass=true])", function(evt) {
 
     // Get the absolute anchor href.
     var href = { prop: $(this).prop("href"), attr: $(this).attr("href") };


garrensmith:
I'm not sure about this. This is where we check the [custom beforeUnloads](https://github.com/apache/couchdb-fauxton/blob/master/app/core/router.js#L26-L47). Then we also listen to the [window.beforeUnload](https://github.com/apache/couchdb-fauxton/blob/master/app/addons/fauxton/components.js#L821-L825). So I'm not convinced your fix would make much difference.

benkeen:
Yes, but the beforeUnload code you noted only gets executed if a .navigate() fires; if an `<a>` just ends up firing its `href="#whatever"` the before unload methods don't get called. 

Here's some steps to test this ticket:
- open up the full page editor on any doc
- make a small change
- click something in the primary nav (e.g. Active Tasks). 

In the current build, it just redirects there. With this ticket, it prompts you to save.

robertkowalski:
I am not sure why this would fix the issue - the cancel button in the editor does not use a `data-bypass` and changing the selector does not change the way event delegation in jQuery works right now. Can you explain how this would remove the race condition you describe and how the race condition looks like? 

After further investigation I think this was fixed in https://github.com/apache/couchdb-fauxton/pull/232 - I can't reproduce the issue on the fresh master.

benkeen:
Hi @robertkowalski thanks for looking at this. That's correct: the issue with the back was already fixed. 

Did you try the steps that I outlined above? This should fix the problem for clicking on any other link in the page that contains a `data-bypass` attribute. 

Maybe let's talk in person, I'm not sure how else to describe this.

benkeen:
Closing. #267 fixes the observed issue.

PR #265, opened 2015-02-11
Changes page Filters tab moved to React

benkeen:
This updates the Filter tab and filter on the Changes page
to use React. It also ports over Components.FilterView and
Components.FilterViewItem found in fauxton/components.js


benkeen:
Hey @robertkowalski, @garrensmith - my first foray into React. Woot!

I largely copied the structure etc. from the code that was already there, but I'm sure there's plenty to criticize. Mind taking a glance if you get any time today? Thanks!

robertkowalski:
looks promising!

 - we will need tests in order to merge it, one test could be that adding a filter adds it to the react wrapper component

 - I think you can delete the `Components.FilterView`

I'll add some other comments next to the code

robertkowalski:
When running the code I get:

```
Linting app/addons/documents/changes/changes-header.react.js ...ERROR
[L27:C8] W033: Missing semicolon.
```

robertkowalski:
stopping here for now - @garrensmith will talk to you later

app/addons/fauxton/components.react.jsx
```diff
@@ -150,12 +151,133 @@ function(FauxtonAPI, React, Stores, Actions) {
...
+
+    getInitialState: function () {
+      return this.getStoreState();
+    },
+
+    submitForm: function (e) {
+      e.preventDefault();
+      e.stopPropagation();
+
+      var filter = this.refs.addItem.getDOMNode().value.trim();
```
benkeen: Check this.

app/addons/fauxton/components.react.jsx
```diff
@@ -150,12 +151,133 @@ function(FauxtonAPI, React, Stores, Actions) {
...
+
+    componentWillUnmount: function() {
+      this.filterViewStore.off('change', this.onChange);
+    },
+
+    render: function () {
+      var tooltip = '';
+      if (this.props.tooltip) {
+        tooltip = '<i ref="tooltip" className="icon icon-question-sign" data-toggle="tooltip" data-original-title=' +
+          this.props.tooltipText + '></i>';
```
benkeen: move to helper

app/addons/fauxton/components.react.jsx
```diff
@@ -150,12 +151,133 @@ function(FauxtonAPI, React, Stores, Actions) {
...
+      }
+
+      var filters = _.map(this.state.filters, function (filter) {
+        return <FilterViewItem namespace={this.props.namespace} key={filter} label={filter} />;
+      }, this);
+
```
benkeen: also move to helper

robertkowalski:
@benkeen this looks a lot better, we will need tests to merge this, examples:

- `ChangesHeader `: the element shows/removes the subcomponent conditionally if the state changes / actions are fired

- test what happens on `onToggle={this.toggleFilterSection}`

- are filters really removed after clicking on the remove button

benkeen:
Thanks, @robertkowalski! Just working on the tests now & ran into some trouble. I'll commit what I have, but I'll have questions tomorrow. Basically I'm unable to confirm that after entering text into the filter it gets added to the DOM. No matter how I try it, I can't find the elements in the updated DOM (possibly my assert() runs prior to the DOM being updated?). Committing now.

benkeen:
Alrighty! Tests added. I think this is good to go.

app/addons/documents/changes/changes-header.react.jsx
```diff
@@ -0,0 +1,236 @@
...
+      e.stopPropagation();
+
+
+      var filter = this.refs.addItem.getDOMNode().value.trim();
+      if (!filter) {
+        return;
+      }
+
+      Actions.addFilterViewItem(filter);
+      this.refs.addItem.getDOMNode().value = '';
```
robertkowalski: You are still directly accessing the DOM from an event handler?

Shouldn't we use an action connected to a store which then updates the element like this:

```
<input type="text" ref="addItem" placeholder={this.props.addItemPlaceholder} value="{this.state.addItemValue}" />
```

robertkowalski: How does that relate to Flux? Should we always directly access the DOM instead of using the roundtrip using actions and stores?

garrensmith: @robertkowalski is correct. This could be simplified to how we assigned a value from onChange

app/addons/documents/changes/changes-header.react.jsx
```diff
@@ -0,0 +1,236 @@
...
+
+    onChange: function () {
+      this.setState(this.getStoreState());
+    },
+
+    // props belong to the parent, so these aren't stored in the related store
+    getDefaultProps: function () {
+      return {
+        addItemPlaceholder: 'Type a filter',
+        helpText: 'e.g. debug or <1.4.1> or any regex',
```
garrensmith: Could you add a better helpText. Why would anyone filter the changes feed with <1.4.1>?

app/addons/documents/changes/changes-header.react.jsx
garrensmith: Here are two examples of how we have worked with inputs using React:
[Index Editor](https://github.com/apache/couchdb-fauxton/blob/master/app/addons/documents/index-editor/components.react.jsx#L64)
[Cors](https://github.com/apache/couchdb-fauxton/blob/master/app/addons/cors/components.react.jsx#L76)

app/addons/documents/changes/changes-header.react.jsx
```diff
@@ -0,0 +1,236 @@
...
+        filters: changesFilterStore.getFilters()
+      };
+    },
+
+    onChange: function () {
+      this.setState(this.getStoreState());
+    },
+
+    // props belong to the parent, so these aren't stored in the related store
+    getDefaultProps: function () {
```
garrensmith: I don't think we need these as props since they are never changed by the parent.

app/addons/documents/changes/stores.js
```diff
@@ -0,0 +1,73 @@
...
+], function (FauxtonAPI, ActionTypes) {
+
+
+  var FilterViewStore = FauxtonAPI.Store.extend({
+    initialize: function () {
+      this.reset();
+    },
+
+    reset: function () {
+      this.filters = [];
```
garrensmith: All internal variables for stores must be prefixed with an `_`. e.g `this._filters.

app/addons/documents/tests/changesHeader.componentsSpec.react.jsx
```diff
@@ -0,0 +1,138 @@
...
+      changesFilterEl = TestUtils.renderIntoDocument(<ChangesHeader.ChangesFilter />, container);
+    });
+
+    afterEach(function () {
+      Stores.changesFilterStore.reset();
+      React.unmountComponentAtNode(container);
+    });
+
+    it('should add filter markup', function () {
+      var $el = $(changesFilterEl.getDOMNode()),
```
garrensmith: This is cool. I like how you have broken the selecting of the DOM here. Makes the test a lot more readable.

app/addons/fauxton/components.react.jsx
```diff
@@ -20,6 +20,7 @@ define([
 function(FauxtonAPI, React, Stores, Actions) {
   var navBarStore = Stores.navBarStore;
 
+
```
garrensmith: This is unrelated to the changes filter. I'm guessing this was by accident.

app/addons/fauxton/stores.js
```diff
@@ -185,6 +185,7 @@ function(app, FauxtonAPI, ActionTypes) {
   Stores.navBarStore = new Stores.NavBarStore();
   Stores.navBarStore.dispatchToken = FauxtonAPI.dispatcher.register(Stores.navBarStore.dispatch);
 
+
```
garrensmith: This is unrelated to the changes filter. I'm guessing this was by accident.

app/addons/documents/changes/changes-header.react.jsx
benkeen: True. I just thought it would be good to design it in a generic way so that those could be easily changed down the road. But I guess I can remove them.

app/addons/documents/changes/changes-header.react.jsx
benkeen: Thanks, guys. I gave this one some thought and I'm kind of on the fence. I'm more than happy making to the change, but it seems to add a lot of complexity just to make it more idiomatic.

If I understand it properly, I'd need to do the following:
- new onChange handler to DOM element
- handler which passing work to actions.js
- new event added in actiontypes.js
- state of input field value stored in the store
- a handler in the store to update the value as 

vs. what is it right now:
- one line within the react view setting the input value to ''.

I'm all for keeping the code as React-y as possible, but for me, this crosses the line. Ultimately, whole need to add in an onchange handler feels like a workaround to ensure state gets modified and that we can reset the value.

[Perfectly happy being overruled, here. Just my 2 cents :)]

benkeen:
Alrighty! *This* time it's totally good to go. ;) All suggested changes made.

app/addons/documents/changes/components.react.jsx
```diff
@@ -0,0 +1,238 @@
...
+
+    getInitialState: function () {
+      return this.getStoreState();
+    },
+
+    submitForm: function (e) {
+      e.preventDefault();
+      e.stopPropagation();
+
+      var filter = this.refs.addItem.getDOMNode().value.trim();
```
garrensmith: This needs to change to a [controlled controller](http://facebook.github.io/react/docs/forms.html)

garrensmith:
@benkeen making good progress. As I mentioned in the previous comment, the way the filter is added needs to be a controlled component.

benkeen:
Thanks @garrensmith. I think the input field was all set up properly - I just wasn't grabbing the current value from the store here. Updated.

benkeen:
Hey @garrensmith, I confess I was getting a bit confused about some things so I thought I'd just experiment a bit and organize things in what I regard as a logical manner. We can just revert it if I'm barking up the wrong tree.

Structure (pretty much the same):

```xml
<ChangesHeader />
	<ChangesHeaderTab />
	<ChangesFilter />
		<Filter /> (list of N components)
		<AddFilterForm />
			<AddFilterFormContent />
				<FilterTooltip />
```

Testing: these would be the key components, like before:
`<ChangesHeaderTab />` - test the hide/show of the tab
`<ChangesFilter />` - test the add/remove filters, tooltip content

Comments:

- It seemed weird that the components were all nice and discrete but shared the same store, so I split it into three: one for the header to track its hide/show state, one for the changes filter to track the list of visible filters, and one for the AddFilterForm to track the state of the text field. [Maybe a bit overengineered now?] I like the new arrangement because looking at a component you can see by checking its store what data is related to it. Before, I was losing track of what piece of state is associated with which component.
- I added a fauxton/mixins.js file to cut down on boilerplate code - namely the this.store.on('change', callback) assignment and destruction.
- I moved as much code relating to a particular component to the component itself, and stopped passing stuff down the chain. e.g. `<Filter />` now handles its own onRemove event and just fires an action. I'm not sure this is in keeping with the "dumb component" idea, but I find this way there's less dependencies between components and code is where I expect to find it.
- I'm still hazy on the need to separate `AddFilterForm` and the form content (`AddFilterFormContent`). They seem so functionally entwined, it feels a bit unnecessary - I'll have to get you to explain that one to me again.

benkeen:
Thanks for all your help, @garrensmith! All updated.
- actions + store tests added
- form moved back to a single component
- dropped third unnecessary store just to track the blank state of the AddFilterForm

app/addons/documents/tests/changes.componentsSpec.react.jsx
```diff
@@ -0,0 +1,215 @@
...
+      assert.ok(Stores.changesHeaderStore.isTabVisible() === true);
+    });
+
+    it('reset() changes tab visiblity to hidden', function() {
+      Stores.changesHeaderStore.toggleTabVisibility();
+      Stores.changesHeaderStore.reset();
+      assert.ok(Stores.changesHeaderStore.isTabVisible() === false);
+    });
+
+//    reset: function () {
```
garrensmith: You can probably just remove these

app/addons/documents/tests/changes.componentsSpec.react.jsx
```diff
@@ -0,0 +1,215 @@
...
+      });
+
+      // similar as previous, except it confirms that the action gets fired, not the custom toggle func
+      it('calls toggleTabVisibility action on selecting a tab', function () {
+        TestUtils.Simulate.click($(tab.getDOMNode()).find('a')[0]);
+        assert.ok(spy.calledOnce);
+      });
+    });
+
+    it('toggleTabVisibility() changes state in store', function() {
```
garrensmith: Can you move the store tests into their own stores test file. It doesn't make sense to have them in the components test file.

app/addons/documents/changes/components.react.jsx
```diff
@@ -0,0 +1,244 @@
...
+
+
+  var Filter = React.createClass({
+    propTypes: {
+      label: React.PropTypes.string.isRequired
+    },
+
+    remove: function (e) {
+      e.preventDefault();
+      Actions.removeFilter(this.props.label);
```
garrensmith: The `<Filter />` component should be "dumb" and not call an action directly but call a callback passed to it from its controller component.

app/addons/documents/changes/components.react.jsx
```diff
@@ -0,0 +1,244 @@
...
+      };
+    },
+
+    submitForm: function (e) {
+      e.preventDefault();
+      e.stopPropagation();
+      if (_.isEmpty(this.state.filter)) {
+        return;
+      }
+      Actions.addFilter(this.state.filter);
```
garrensmith: This should also be a dumb component that just calls a callback on a submit form and not an action directly.

app/addons/documents/changes/components.react.jsx
```diff
@@ -0,0 +1,244 @@
...
+    },
+
+    render: function () {
+      var filters = this.getFilters();
+
+      return (
+        <div className="tab-content">
+          <div className="tab-pane active" ref="filterTab">
+            <div className="changes-header js-filter">
+              <AddFilterForm tooltip={this.props.tooltip} filter={this.state.filter} onSubmit={this.submitForm} />
```
garrensmith: You don't have a `submitForm` method in this component.

garrensmith:
@benkeen this is looking better. Two components just need to be more self contained. I also picked up some bugs while testing. 

* Once I add a filter it doesn't clean the input field
* If I remove a fliter it doesn't regenerate the changes list without that filter
* Everytime I click on add I get this error `Warning: flattenChildren(...): Encountered two children with the same key, `.$hello`. Child keys must be unique; when two children share a key, only the first child will be used.`

benkeen:
All fixed except for the last React error you spotted. I'll fix that up now. The code shouldn't allow adding the same item multiple times as it does right now; I spoke to Sean and I'm going to add an flash effect on the input field to indicate there's a problem. This is very much a fringe case, so that form of error should be sufficient.

benkeen:
k, here we go. Had to do a hack to get the error highlight blink to work, but it was between two evils: using jQuery or plain CSS, so I vouched for the latter.

benkeen:
Commits squashed. The failing unit test was unrelated to this ticket. Let's see what Travis says this time.

garrensmith:
+1. This is great work @benkeen. Merge it.

benkeen:
Merged as 392dc52

PR #267, opened 2015-02-12
Fix missing confirmation when leaving editor

robertkowalski:
To not render a property in React we need to pass `null` to it. If
we are passing `false` or `"false"` it gets rendered into the DOM.

This fixes a bug where you edited a document and then got no
confirmation dialog if you are really sure you want to leave the
page without saving if you clicked on the navigation bar on the
left.

Having `data-bypass="false"` will mean to the selector
`:not(["data-bypass"])` that there is a `data-bypass` - containing
a String! (which then matches the selector)

What we want is _no_ `data-bypass` if we don't want to bypass.

Looking at the old DOM a month ago shows how we used `data-bypass`:

bildschirmfoto 2015-02-12 um 19 12 19

Closes COUCHDB-2574

superseeds #263

### DIFF
--- a/app/addons/documents/tests/nightwatch/navigateAfterEditingDocShouldShowAConfirmationBox.js
+++ b/app/addons/documents/tests/nightwatch/navigateAfterEditingDocShouldShowAConfirmationBox.js
@@ -1,0 +1,37 @@
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License. You may obtain a copy of
+// the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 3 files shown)


benkeen:
Awesome. +1 

I still think https://github.com/apache/couchdb-fauxton/pull/263 should be added, because the currently logic is incorrect: it just checks for the *existence* of the data-bypass attribute and not its value (true/false). As such, it will never run the event delegated code (FauxtonAPI.navigate()) for anything with that attribute regardless of value. 

Your fix fixes it for React components, but not for any old code. 

Just my 2 cents.

robertkowalski:
@benkeen I like the idea of #263 as it follows the principles of the least surprise.

What do you think of something like this for changing the routers api: match on all, but not on `data-bypass="false"`. This would also keep older components like the global notifications working.

robertkowalski:
merged as b030b218a5e9e1fd891bc9750c593f356dd5cb7f and c6e6a438f9bd29434215c93f0552db8192fecf75

PR #269, opened 2015-02-13
Fix for polling interval label not updating

benkeen:
Due to a bug in IE11, the onInput event doesn't fire on the range
slider on incremental changes. This adds both onInput and onChange
events to ensure it works for all browser environments.

Closes COUCHDB-2579


robertkowalski:
+1

robertkowalski:
works like a charm :)

benkeen:
Thanks Robert! Merged as ff1922a022762ab8b416d164118469846d0203fd

PR #270, opened 2015-02-13
Fix for DB page not updating after new db added

benkeen:
Note: there's no new test for this because a test already
existed to confirm exactly this, except it would have failed on IE
if it ran on that browser.

Closes COUCHDB-2580


garrensmith:
+1

benkeen:
Thanks Garren. Merged as 133eda4a2aa4b835bd1f66dc7215ddbce12e1f18

PR #272, opened 2015-02-16
View name resets

garrensmith:
When creating a new view if a user changed the reduce editor settings the view name would reset
back to original
This fixes COUCHDB-2582


robertkowalski:
+1 - thank you!

PR #274, opened 2015-02-17
Fix for encoding unicode char

benkeen:
Contains a small fix to ensure a unicode char in the db typeahead
component is URL encoded. If the server needed to recognize the
endkey parameter it would cause problems on IE.

This doesn't currently cause any problems on Fauxton, note, so 
there's nothing to test there.

Closes COUCHDB-2587

### DIFF
--- a/app/addons/fauxton/components.js
+++ b/app/addons/fauxton/components.js
@@ -615,19 +615,24 @@ function(app, FauxtonAPI, ace, spin, ZeroClipboard) {
       }
       _.bindAll(this);
     },
-    source: function(query, process) {
+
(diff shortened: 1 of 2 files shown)


garrensmith:
@benkeen can you add a test. All fixes must include a test.

benkeen:
Mocha test added. I had to refactor the component a bit to expose a convenient `getURL()` method through which I could examine the URL.

robertkowalski:
+1

benkeen:
Merged as 88148d6 [squashed, hence the travis re-run]

PR #288, opened 2015-02-24
Two Bug fixes

garrensmith:
* Fix Previous '<' works correctly from view
* Only remove all doc numbers if it exists


robertkowalski:
Hi Garren,

the code breaks the next button, the failing test for it is:

```
Running:  Select all works after changing the page 
```

robertkowalski:
+1

PR #290, opened 2015-02-24
Nightwatch auth moved to separate folder

benkeen:
This has been done to allow overriding, simply by specifying a
different location for the folder in the settings.json file.


benkeen:
The failing zeroclipboard failing test is unrelated.

robertkowalski:
Ok, how do we proceed? Maybe we should try to make the test more stable to make live easier for everyone?

benkeen:
Yes, obviously the test needs to be fixed. We can either merge this now or wait for that test to be fixed - either is fine. I've had zero time to look at it this week, but I can look next.

garrensmith:
+1. I'm ok with this being merged in. I think the test failure is unrelated and this is useful.

benkeen:
Thanks Garren. Merged as f9b0ee0.

PR #293, opened 2015-02-26
Constants: move to Api

robertkowalski:
move constants to the creation of FauxtonAPI to make them
available sooner as they are a property of FauxtonAPI

Fixes the broken testsuite in a browser and App boot problems


benkeen:
+1

michellephung:
+1

PR #294, opened 2015-02-27
delete unused test skeletons

robertkowalski:
they are not used and backbone based view testing is painful
anyway, so i don't think someone will add some in the future.

if it happens, we can revert this commit.


michellephung:
+2 for the pain

robertkowalski:
merged as 69d098bf560ebebba4500361f1d090899dfbaa5b

PR #295, opened 2015-02-27
Extract common react components

robertkowalski:
- extract components: header-toggle-button-component
 - extract components: styled-select component


garrensmith:
+1 this is awesome. We could also add the general components to something like `Fauxton.components.StyledSelect` so we can use them anywhere.

robertkowalski:
I have some mixed feelings regarding that, let's talk tomorrow.

PR #298, opened 2015-03-02
fix broken api-url link

robertkowalski:
also removes unused parameter from function, the only usage of
`indexedDocs.urlRef` is in `routes-index-editor.js`

### DIFF
--- a/app/addons/documents/tests/resourcesSpec.js
+++ b/app/addons/documents/tests/resourcesSpec.js
@@ -29,7 +29,11 @@ define([
         database: {id: 'databaseId', safeID: function () { return this.id; }},
         design: '_design/myDoc'
       });
+    });
 
+    it('does not remove an id attribute', function () {
+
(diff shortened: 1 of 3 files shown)


app/addons/documents/tests/resourcesSpec.js
```diff
@@ -29,7 +29,11 @@ define([
         database: {id: 'databaseId', safeID: function () { return this.id; }},
         design: '_design/myDoc'
       });
+    });
+
+    it('does not remove an id attribute', function () {
```
kxepal: accidental newline?

michellephung:
+1

app/addons/documents/tests/resourcesSpec.js
robertkowalski: no, we often make a newline before the final assertion

robertkowalski:
merged as bbebeb7e0538588934ff82ac5bb8814c3b50c4b3

PR #299, opened 2015-03-02
Detect system databases on _

robertkowalski:
All system databases are now prefixed by `_`, so we can remove
a lot of detection logic.

Additionally colored the warning in red.

closes COUCHDB-2607

### DIFF
--- a/app/addons/databases/resources.js
+++ b/app/addons/databases/resources.js
@@ -82,45 +82,6 @@ function(app, FauxtonAPI, Documents) {
     }
   });
 
-  Databases.IsSystemDatabaseModel = FauxtonAPI.Model.extend({
-
-    initialize: function (options) {
-      this.name = options.name;
-    },
-
-    url: function () {
-      return app.host + '/_stats';
-    },
-
-    sync: function (method, model, options) {
-      options.url = this.url();
-      return $.ajax(options);
-    },
-
-    parse: function (data) {
-      var isOnFrontendNode,
-        isSystemDatabase = false,
-        systemDatabases = [
-          '_replicator',
-          '_users',
-          'nodes',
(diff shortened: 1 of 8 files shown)


kxepal:
LGFM. Few questions:

1. Shouldn't Database mode have a property `isSystemDatabase` which encapsulates logic about how this determines?
2. Shouldn't Fauxton instead of warn about deletion system database completely prevent this action? Synthetic restriction, but will definitely save users from shooting them self in the foot.

What do you think?

robertkowalski:
good points!

1. fixed, you are right
2. hmmm... I think we should allow it as it, I often delete my system dbs, e.g. _replicator

kxepal:
@robertkowalski yea, killing _replicator and _users is safe, but I'm very sceptical that _dbs, _nodes and _metadata would be too. Anyway, just thinking loudly, worth to check this moment.

robertkowalski:
@kxepal i'll ask some folks regarding that second point

robertkowalski:
@kxepal got feedback, they say the users should be able to delete them, as sometimes replications start failing so badly that there's no way to delete it by rev or even PID and just deleting the _replicator db helps in those cases.

kxepal:
@robertkowalski hm...that's definitely a bad use case, some sort of problem workaround. Worth to look on this problem closely. But that's beyond of this PR (:

Thanks!

michellephung:
+1 after you change app/addons/databases/resources.js: 53 to `return (/^_/).test(this.id);`

btw, pretty neat @robertkowalski ! 
:D

robertkowalski:
merged as 4b3b73f44349230ea8ae3a877f4c2e2a87258d9d :)

PR #301, opened 2015-03-03
React docs pagination

garrensmith:
This is the pagination converted to Flux and React


garrensmith:
This fixes COUCHDB-2193

michellephung:
holy macaroni, 989 additions and 835 deletions!

+1

benkeen:
Nice.

The ">" icon sometimes doesn't get highlighted when it should. I've noticed this on Chrome + Safari. Reproducing it consistently has been tough, but I find these steps often work:

- starting on the databases page, select a DB with lots of records (>1 page)
- on the db page [sometimes I've found it's not enabled here, right off the bat] click on the edit icon for the first record
- on the full page editor page, click the "cancel" link
- back on the documents page I find that here often isn't enabled. Note: on Safari I find it works more often, but when you click ">" it takes you to a blank page.

Doing a refresh on the pages where it isn't enabled doesn't always fix it.

garrensmith:
@benkeen thanks for reviewing. After a bit of digging around I found that actually isn't a bug. If you set the per page to 20 docs per page. Then edit a doc and then click cancel. It returns you to the _all_docs page but sets the limit = 20 docs. So that means that you cannot paginate any further.

The issue here, which I cannot fix on this branch, is that the cancel url shouldn't have the limit of 20. It should be the same as pressing '<'. 

But I did manage to find another bug while testing this so thanks. I've pushed a small fix.

app/addons/documents/pagination/pagination.react.jsx
```diff
@@ -0,0 +1,198 @@
...
+      previousClicked: function (event) {
+        event.preventDefault();
+        event.stopPropagation();
+        if (!this.state.canShowPrevious) { return; }
+        Actions.paginatePrevious();
+      },
+
+      render: function () {
+        var canShowPreviousClassName;
+        var canShowNextClassName;
```
robertkowalski: would default them to an empty string, see https://issues.apache.org/jira/browse/COUCHDB-2577

app/addons/documents/pagination/tests/paginationStoreSpec.js
```diff
@@ -0,0 +1,237 @@
...
+        store.paginatePrevious();
+
+        assert.equal(store.getPageEnd(), 20);
+      });
+
+    });
+
+    describe('totalDocsViewed', function () {
+      beforeEach(function () {
+        store.setPerPage(20);
```
robertkowalski: you can add `store.reset();` here and remove it from the testcases

app/addons/documents/shared-routes.js
```diff
@@ -162,15 +126,9 @@ define([
...
-          defaultPageSize,
-          isLazyInit,
-          pageSize,
+          pageSize = PaginationStores.indexPaginationStore.getPerPage(),
           collection;
 
-      isLazyInit = _.isUndefined(this.documentsView) || _.isUndefined(this.documentsView.allDocsNumber);
-      defaultPageSize = isLazyInit ? FauxtonAPI.constants.MISC.DEFAULT_PAGE_SIZE : this.documentsView.perPage();
-      docParams.limit = pageSize = this.getDocPerPageLimit(urlParams, defaultPageSize);
-
```
robertkowalski: :heart_eyes:

app/addons/documents/views.js
```diff
@@ -512,22 +492,6 @@ function (app, FauxtonAPI, Components, Documents,
...
-    setParams: function (docParams, urlParams) {
-      this.docParams = docParams;
-      this.params = urlParams;
-      this.perPageDefault = this.docParams.limit;
-
-      if (this.params.limit) {
-        this.pagination.docLimit = this.params.limit;
-      }
-    },
-
```
robertkowalski: :heart_eyes:

robertkowalski:
fixes: https://issues.apache.org/jira/browse/COUCHDB-2572 :)

robertkowalski:
+1 great!

garrensmith:
Merged.

PR #303, opened 2015-03-03
Changes page content moved to React

benkeen:
One comment: I couldn't decide whether to group the entire page into *one* React element that includes everything - Header + Results. Very tempting, but I decided not to because it would mean duplicating the markup in the core template & would mean the Changes page always rewrote the entire page rather than just load content into the `#dashboard-upper-content` and `#dashboard-lower-content` like other pages.


app/addons/documents/routes-documents.js
```diff
@@ -219,16 +217,6 @@ function(app, FauxtonAPI, BaseRoute, Documents, Changes, Index, DocEditor, Datab
...
-    addFilter: function (filter) {
-      this.changesView.filters.push(filter);
-      this.changesView.render();
-    },
-
-    removeFilter: function (filter) {
-      this.changesView.filters.splice(this.changesView.filters.indexOf(filter), 1);
-      this.changesView.render();
-    },
-
```
robertkowalski: :heart_eyes:

robertkowalski:
One minor thing, you can use `$('.element').text()` or `$element.html()` in all cases where you call `$('.element')[0].innerHTML`

app/addons/documents/tests/changes.componentsSpec.react.jsx
```diff
@@ -179,7 +188,185 @@ define([
...
+    var maxChanges = 10;
+
+    beforeEach(function () {
+
+      // to keep the test speedy, override the default value (1000)
+      Stores.changesStore.setMaxChanges(maxChanges);
+
+      // better underscorey way to do this?
+      var changes = [];
+      for (var i=0; i<maxChanges + 10; i++) {
```
robertkowalski:
```js
_(10).times(function (i) { 
  changes.push(new Backbone.Model({ id: 'doc_' + i, seq: 1, changes: { code: 'here' } }));
});
```

app/addons/documents/changes/stores.js
```diff
@@ -15,51 +15,58 @@ define([
...
+              match = true;
+            }
+          });
+          return match;
+        });
+      }, this._changes, this);
+
+      this._showingSubset = matches.length > this._maxChangesListed;
+      return matches.slice(0, this._maxChangesListed);
     },
```
robertkowalski: you can already limit the max-changes before the reduceby slicing them - that should increase performance a lot

app/addons/documents/tests/changes.storesSpec.js
```diff
@@ -53,21 +49,50 @@ define([
...
+      assert.equal(Stores.changesStore.getDatabaseName(), '');
+    });
+
+    it("getChanges() should return a subset if there are a lot of changes", function () {
+
+      // to keep the test speedy, override the default max value
+      var maxChanges = 10;
+      Stores.changesStore.setMaxChanges(maxChanges);
+
+      // better underscorey way to do this?
```
robertkowalski: times

robertkowalski:
Commit message is missing that it also limit the number of changes to 1000 which closes COUCHDB-2216

app/addons/documents/views-changes.js
```diff
@@ -18,108 +18,51 @@ define([
...
-      } else {
-        $button.text("View JSON")
-          .addClass("btn-primary")
-          .removeClass("btn-secondary");
-      }
-
-      $container.slideToggle();
+    afterRender: function () {
+      ChangesActions.setChanges({
+        changes: this.model.changes.models,
```
garrensmith: Out of interest, any specific reason you pass the model array instead of the collection? I would think it makes more sense to pass the collection.

garrensmith:
@benkeen this is excellent. I also like the idea of grouping this into one React component. If its not too much extra work could you convert it to do that.

app/addons/documents/changes/components.react.jsx
```diff
@@ -11,34 +11,39 @@
...
 
 
-  // the top-level component for the Changes Filter section. Handles hiding/showing
-  var ChangesHeader = React.createClass({
+  // the top-level component for the Changes Filter section. Handles hiding/showing of the filters form
+  var ChangesHeaderController = React.createClass({
     getInitialState: function () {
       return {
-        showTabContent: Stores.changesHeaderStore.isTabVisible()
+        showTabContent: changesStore.isTabVisible()
```
garrensmith: This and line 37 could be made into a `getStoreState()` and then for `getInitialState` can just return `getStoreState` and for onchange you can do `this.setState(this.getStoreState());`

app/addons/documents/changes/components.react.jsx
```diff
@@ -268,18 +271,169 @@ define([
   });
 
 
+  var ChangesController = React.createClass({
+    getInitialState: function () {
+      return this.getState();
+    },
+
+    getState: function () {
```
garrensmith: could you rename to `getStoreState`

app/addons/documents/changes/components.react.jsx
```diff
@@ -268,18 +271,169 @@ define([
...
+
+    toggleJSON: function (e) {
+      e.preventDefault();
+      this.setState({ codeVisible: !this.state.codeVisible });
+    },
+
+    getChangesCode: function () {
+      var json = '';
+      if (this.state.codeVisible) {
+        var code = {
```
garrensmith: You could neaten this a bit by moving it into a function.

app/addons/documents/tests/changes.componentsSpec.react.jsx
```diff
@@ -179,7 +188,185 @@ define([
...
+
+      // confirm only one item shows up and the ID maps to what we'd expect
+      assert.equal(1, $(changesEl.getDOMNode()).find('.change-box').length);
+      assert.equal('doc_3', $(changesEl.getDOMNode()).find('.js-doc-id')[0].innerHTML);
+    });
+
+
+    // confirms that if there are multiple filters, ALL are applied to return the subset of results that match
+    // all filters
+    it('multiple filters should all be applied to results', function () {
```
garrensmith: For future reference this test and the test before it you could actually test at the store level since the views just render what ever changes they are given. The way you have done it here is more like an integration test than a unit test.

app/addons/documents/tests/views-changesSpec.js
```diff
@@ -18,70 +18,30 @@ define([
       ViewSandbox = testUtils.ViewSandbox,
       viewSandbox;
 
-  describe('Documents Changes', function () {
+  describe('ChangesReactWrapper', function () {
```
garrensmith: Is this file still useful since all the tests are now in React?

app/addons/documents/tests/changes.componentsSpec.react.jsx
benkeen: Niiiice :D

app/addons/documents/changes/stores.js
```diff
@@ -15,51 +15,58 @@ define([
...
+      this._databaseName = options.databaseName;
+      this._changes = _.map(options.changes, function (change) {
+        return {
+          id: change.get('id'),
+          seq: change.get('seq'),
+          deleted: change.get('deleted') ? change.get('deleted') : false,
+          changes: change.get('changes'),
+          doc: change.get('doc') // only populated with ?include_docs=true
+        };
+      });
```
robertkowalski: You can do `return change.toJSON()` instead, much shorter and no need to manually maintain it in the future

robertkowalski: oh wait, even simpler:

```js
this._changes = options.changes.toJSON();
```

benkeen: True! The thing is, the models have a whole bunch of stuff in them - I just wanted to specify exactly what's needed for the element, so although it's manual you can see at a glance what the changes array contains. Also, it forces the `deleted` property to have a default `false` value, which is needed for the filter searches later. 

This okay?

app/addons/documents/views-changes.js
benkeen: This relates to the recent conversation on Slack. I wanted to remove as much Backbone as possible before sending it to the store. Just felt a bit cleaner passing an array rather than a collection. But I realize this is kind of moot since they're models anyway. I can pass the collection.

app/addons/documents/changes/stores.js
benkeen: True!

app/addons/documents/tests/changes.componentsSpec.react.jsx
benkeen: Yeah that makes sense, thanks.

benkeen:
Thanks for the feedback, guys. Very helpful! I've updated the code as per your feedback.

@garrensmith, I tried combining the elements into a single ChangesPageController element - turned out that part was very straightforward, even with a generic `TabbedPage` component that allowed you passing the header + body react components. But the single React element caused lots of problems with the different pages being handled by the route object. Some of them do view.remove() on the header but not the main body content when navigating around, which proved very difficult when the whole page was a single React component. After a lot of experimenting I decided it may be a little too soon for that refactor.

garrensmith:
@benkeen I've spotted some performance issues while testing. Even though you set a display limit of 1000, the changes request still fetches ALL changes. So I have a database with 37000 changes. Those changes are still kept in memory. 

To fix this change the `buildChanges` function in `/app/addons/databases/resources.js` to something like this:

```javascript
buildChanges: function (params) {
      if (!params.limit) {
        params.limit = 100;
      }

      this.changes = new Databases.Changes({
        database: this,
        params: params
      });

      return this.changes;
    }
```

I've set a limit of 100. I think thats actually a better size. I think once this PR has landed We need to add the functionality for a user to look at older changes and also see any new changes while viewing the changes page.

benkeen:
Good call. Updated.

garrensmith:
@benkeen great I'm +1 then.

benkeen:
Rebased with master; a new test in master caused the last build to fail. Patched.

benkeen:
Merged as b3e5b44

PR #310, opened 2015-03-06
Add missed license headers

kxepal:


robertkowalski:
+1 thank you!

kxepal:
@robertkowalski I also noticed that some .html files contains `<!--...-->`, while others `<%/*...*/%>` - any difference?

robertkowalski:
@kxepal we are migrating away from `<!--` because it gets rendered in development mode and clutters the DOM which makes it hard to inspect.

Using `/*` makes the license headers also disappear in development mode and not just after a production build.

kxepal:
@robertkowalski hm...so in html block used `<%...%>` while on pages html - `<!--...-->` right? otherwise such strip isn't good for the releases we ships.

How to distinguish where is a block and where is a page? and what about *.less files?

robertkowalski:
i am not sure if i understand your question... fauxton's "html"-files are lodash templates where you can write javascript inside those guys: `<%   %>`

the templates are then converted to js code and the the js in those <% %> is evaluated before the templates are converted to full js code and get embedded in our AMD modules - that process strips comments from the js (like /* foo */ or // foo )

i don't see why this could break a release build

in general we are slowly moving from  in all of our html files to <% /* */ %> because it make sinspecting the DOM easier and has no negative effect on the production-build

in less files we use usually // but /* is also valid
// gets stripped on a build, /* not 
(for less files)

kxepal:
It's not about breaking release in term or...it wouldn't work at all. It's about breaking license requirements (have a ASF header in each file) for release artefacts. But I feel that would be another part of the story - just asked because of curiosity.

PR #312, opened 2015-03-06
fix update-seq

robertkowalski:
sometimes the update sequence is a string so we have to handle it

### DIFF
--- a/app/addons/databases/resources.js
+++ b/app/addons/databases/resources.js
@@ -137,6 +137,8 @@ function(app, FauxtonAPI, Documents) {
       var updateSeq = this.get("update_seq");
       if (full || (typeof(updateSeq) === 'number')) {
         return updateSeq;
+      } else if (typeof(updateSeq) === 'string') {
+        return updateSeq.split('-')[0];
       } else if (updateSeq) {
         return updateSeq[0];
       } else {


app/addons/databases/resources.js
```diff
@@ -137,6 +137,8 @@ function(app, FauxtonAPI, Documents) {
       var updateSeq = this.get("update_seq");
       if (full || (typeof(updateSeq) === 'number')) {
         return updateSeq;
+      } else if (typeof(updateSeq) === 'string') {
+        return updateSeq.split('-')[0];
```
kxepal: Above you return a number as like as below. Here you return a string. Wouldn't it cause any problems somewhere ahead?

garrensmith: I think you can use `_.isNumber` and `_.isString` to be clearer.

app/addons/databases/resources.js
```diff
@@ -137,6 +137,8 @@ function(app, FauxtonAPI, Documents) {
...
       } else if (updateSeq) {
         return updateSeq[0];
       } else {
```
garrensmith: this final else isn't needed. Just return 0 at the end of the function.

garrensmith:
This needs a test.

app/addons/databases/resources.js
kxepal: Sure I can, but should I if function will guarantee to return a single type? (;

garrensmith: I don't think it matters the type as this is just rendered. So ultimately its converted to a string. We could always convert to a string in this method to be safer.

kxepal: Ok.

PR #313, opened 2015-03-09
componentize: add confirmbutton component

robertkowalski:


robertkowalski:
we will use that one in mango and other screens

app/addons/components/react-components.react.jsx
```diff
@@ -188,7 +188,20 @@ function (app, FauxtonAPI, React, Components, beautifyHelper) {
     }
   });
 
+  var ConfirmButton = React.createClass({
```
garrensmith: Maybe we should rather call it a `SubmitButton`.

robertkowalski: it is called confirmbutton because it 

```
btn-success save
```

and

```
fonticon-ok-circled
```

garrensmith:
Thanks for the explanation. +1

robertkowalski:
merged as 1de65dc28ce85aa1e3a2f4f102df6fd70a66235e

PR #318, opened 2015-03-15
Call JSX build in a way that works on *nix and Windows

sebastianrothbucher:


robertkowalski:
+1 - feel free to merge after travis is green :)

Gruntfile.js
```diff
@@ -407,7 +407,7 @@ module.exports = function(grunt) {
     shell: {
         'build-jsx': {
             command: [
-              './node_modules/react-tools/bin/jsx -x jsx app/addons/ app/addons/',
+              'node ./node_modules/react-tools/bin/jsx -x jsx app/addons/ app/addons/',
                 'rm -rf <%= src_path %>/js/app/views/.module-cache/'
```
garrensmith: will `rm -rf` work on a windows machine?

garrensmith: Actually that whole line can be removed since it actually doesn't work. The `.module-cache` isn't saved to that directory.

sebastianrothbucher:
With mingw32 you indeed have *nix commands. Anyhow I'll remove and retest 2night.
Thx 4 the Feedback!

kxepal:
How about to move such platform depended commands to the platform independent script? Since you require to have install nodejs in anyway, but require mingw32 to build Fauxton sounds as overkill.

garrensmith:
@kxepal the `rm -rf` command is not needed. If we remove that then Fauxton will run on Windows with no special requirements.

kxepal:
@garrensmith that's simplifies everything (:

sebastianrothbucher:
OK, I removed the rm and re-tested (delete a .react.js and redo grunt dev) on both mingw32 and the good old cmd.exe => both work. 
I'll wait for travis' verdict and then merge. At the latest 2morrow during Hamburg Couch meetup ;-)
Thanks!   Sebastian

garrensmith:
+1

PR #320, opened 2015-03-17
tests: fix description for test

robertkowalski:
the test tests that the right api-url is returned (no relative
url for the app, but an absolute for the url)

### DIFF
--- a/app/addons/documents/tests/resourcesSpec.js
+++ b/app/addons/documents/tests/resourcesSpec.js
@@ -31,9 +31,9 @@ define([
       });
     });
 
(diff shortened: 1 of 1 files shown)


garrensmith:
That looks better, this test will fail if you run it on its own. You need to include `'addons/documents/base'` in the require list.

robertkowalski:
i see, should be fixed now

garrensmith:
+1

robertkowalski:
merged as f199df6d9b0c9f61675160e6a8c47abe40a66833 and 2ab01436a1f47773a96fe6c9a1b80fb003c85fe7

PR #323, opened 2015-03-18
Fix for deleting database with '/' in name

benkeen:
Previously you'd get an error when trying to delete a database
that contained a slash in the name.

### DIFF
--- a/app/addons/documents/views.js
+++ b/app/addons/documents/views.js
@@ -196,7 +196,7 @@ function (app, FauxtonAPI, Components, Documents,
         clear: true
       });
 
-      this.database.url = FauxtonAPI.urls('databaseBaseURL', 'server', this.database.id, '');
+      this.database.url = FauxtonAPI.urls('databaseBaseURL', 'server', this.database.safeID(), '');
 
       this.database.destroy().then(function () {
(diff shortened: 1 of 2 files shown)


app/addons/documents/views.js
```diff
@@ -196,7 +196,7 @@ function (app, FauxtonAPI, Components, Documents,
         clear: true
       });
 
-      this.database.url = FauxtonAPI.urls('databaseBaseURL', 'server', this.database.id, '');
+      this.database.url = FauxtonAPI.urls('databaseBaseURL', 'server', encodeURIComponent(this.database.id), '');
```
robertkowalski:
```js
this.database.saveID()
```

what do you think?

robertkowalski: https://github.com/apache/couchdb-fauxton/blob/b3e5b44e278c6aa750eb21c1b2f1025ede83c9a8/app/addons/databases/resources.js#L77

benkeen: Yeah that's better. Will update in a sec.

michellephung:
+1

benkeen:
Merged. (I squashed, hence the re-running of the test - but it was green)

PR #325, opened 2015-03-22
tooling: don't require globally installed grunt

robertkowalski:
`grunt` is one of our dependencies, using `npm run dev` we can
substitute `grunt dev` which requires a globally installed grunt
which can have a completely different version than we are defining
in our `package.json`

Using `scripts` in the npm package.json file npm takes care that
our locally installed modules which provide a `bin` file are part
of the env of the executing context.

New commands:

```
npm run dev
npm run nightwatch
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,10 @@
     "stylecheck": "jsxcs app && jsxcs test/nightwatch_tests && jsxcs tasks && jsxcs Gruntfile.js",
     "test": "grunt test",
     "couchdebug": "grunt couchdebug",
-    "couchdb": "grunt couchdb"
+    "couchdb": "grunt couchdb",
+    "couchapp": "grunt couchapp_deploy",
+    "dev": "grunt dev",
+    "nightwatch": "grunt nightwatch"
   },
   "repository": {
     "type": "git",


kxepal:
Does it was? I always used `./bin/grunt <cmd>` inside the fauxton directory without having grunt installed system wide...

robertkowalski:
@kxepal yes you can still use `./bin/grunt` - using the npm scripts section is the convenient way to do that plus that everyone new to the project can run `npm run -l` and gets a list of available tasks. Additionally the npm run commands would not change if we decide to use another tool instead of grunt.

It is also quite common that I am looking into the npm scripts section of a new project to see what the possible commands are.

kxepal:
Sounds good. Cool! May I have a request a command there for couchapp deploy then?(:

robertkowalski:
my pleasure :) fixed!

kxepal:
You are the wizard (: Thank you so much!

benkeen:
+1 neat trick!

robertkowalski:
travis is green, rebasing...

garrensmith:
@robertkowalski could you close this.

PR #337, opened 2015-03-26
alldocs: make text configurable

robertkowalski:
the current text says: `no index created yet`. For all-docs it
must be something like `no document created yet`


benkeen:
+1

PR #338, opened 2015-03-26
docs: cleanup old views

robertkowalski:
remove a lot of old cold that we don't need any more after we
refactored the all-docs-list


app/addons/documents/shared-resources.js
```diff
@@ -36,9 +36,9 @@ define([
     initialize: function (_attrs, options) {
       if (this.collection && this.collection.database) {
         this.database = this.collection.database;
-      } else if (options.database) {
-        this.database = options.database;
+        return;
       }
+      this.database = options.database;
```
benkeen: Seems like the logic changed a little here. The old version only set this value if options.database evaluates to true; this always sets it.

app/addons/documents/shared-resources.js
```diff
@@ -27,7 +27,7 @@ define([
       return FauxtonAPI.constants.DOC_URLS.GENERAL;
     },
     url: function (context) {
-      if (context === undefined) {
+      if (!context) {
```
benkeen: Again, slight logic change here (if context is 0 or false this will execute) This okay?

I kinda like _.isUndefined(). Verbose but explicit!

robertkowalski: who calls `Model.url(0);` and why? and if they do I guess they are happy with an url from the server

app/addons/documents/shared-resources.js
robertkowalski: The model can't work in any way without setting a database (e.g. see `url` method)

app/addons/documents/shared-resources.js
robertkowalski: some background info, sorry for the short answer:

the implementation is meant to take care of calls without a parameter: `.url()` - if that is the case it will choose to return the url for `server`.

this is also true for `.url(undefined)` - I am not sure if the user wants the url for the server if passing `undefined`

robertkowalski: i think the most cleanest way would be to remove the lazy context choosing, i grepped the code and just saw two occurences.

that way you have to pass the argument every time explicitly and the error logging in the url registry works in a reliable way

robertkowalski:
hi @benkeen. i removed the code in question, so we don't refactor the old code any more. please also read my comments above.

sebastianrothbucher:
Isn't most of the code going to disappear anyway as react is introduced further? (then it makes sense to just don't refactor the old code any more indeed). BTW: from a fucntional point of view the PR seems to work OK

robertkowalski:
we still use the backbone models for accessing the api endpoints

garrensmith:
+1 thanks.

robertkowalski:
rebased and merged as

79d5f320a4a708cf7c2bb5dad3480f0953e402de
5a10f1ca67347709d5654c4f5f775c63c8866e26

PR #340, opened 2015-03-27
removing old unused event

benkeen:
This event is no longer used & was replaced with NAVBAR_SIZE_CHANGED


robertkowalski:
+1 merge after travis is green

PR #330, opened 2015-03-23
Design doc names truncated in sidebar

benkeen:
Long design doc names displayed in the sidebar of the Docs page would
overlap with the rightmost ("+") icon. This truncates them with an
ellipsis and adds a tooltip (just a title attr) to provide a means to
let user see the full design doc name.

PR #345, opened 2015-03-30
Design doc names truncated in sidebar

benkeen:
Long design doc names displayed in the sidebar of the Docs page would
overlap with the rightmost ("+") icon. This truncates them with an
ellipsis and adds a tooltip (just a title attr) to provide a means to
let user see the full design doc name.


benkeen:
This is a copy of https://github.com/apache/couchdb-fauxton/pull/330 which wouldn't kick off a travis build for some reason. See that ticket for screenshots & +1

benkeen:
Merged as b671e75

PR #348, opened 2015-03-31
Remove Secure from Cookie in dev proxy

garrensmith:
Remove Secure from the Cookie so that this will work with a CouchDB
instance that is using https and a Secure cookie.


robertkowalski:
+1

PR #349, opened 2015-03-31
Add support for extensions on doc list

benkeen:
This PR allows extensions to add buttons to the list of
documents in the React DocList. This functionality used to be
there but was removed with the recent React code refactor.

Also, a title attribute was added to the doc Edit icon.


benkeen:
@garrensmith can I get you to glance at this when you get a moment? Basically it's identical to the old code, except now it assumes the extension is a React component instead of a Backbone View.

garrensmith:
+1 nice @benkeen

benkeen:
Merged as f183176

PR #350, opened 2015-03-31
Gruntfile fix

benkeen:
A recent commit assumes the template always has a certain structure;
this just adds a check to only set i18n for elements that exist.


michellephung:
+1

benkeen:
This is now moot.

PR #351, opened 2015-04-01
build: don't hardcode environments

robertkowalski:


ksnavely:
Do you need to add `couchapp` to [this](https://github.com/apache/couchdb-fauxton/pull/351/files#diff-35b4a816e0441e6a375cd925af50752cR102)?

garrensmith:
@ksnavely I don't actually think so.

garrensmith:
merged

robertkowalski:
@ksnavely no, as these are just default values - `settings.template` is set in https://github.com/apache/couchdb-fauxton/blob/d7641f16e7a12faff896136826b43f7affb0a393/tasks/helper.js#L21

PR #353, opened 2015-04-01
sidebar: fix collapsible for special chars

robertkowalski:
fix collapsible for design docs with special chars in their name,
like `_design////design//foo` or `_design/bar.foo.bar`

internally sizzle works with `querySelectorAll` which needs
escaping for special chars in css selectors.

based on a patch from Michelle Phung <michellephung@gmail.com>

### DIFF
--- a/app/addons/documents/shared-views.js
+++ b/app/addons/documents/shared-views.js
@@ -17,7 +17,8 @@ define([
        "addons/fauxton/components",
 
        "addons/documents/resources",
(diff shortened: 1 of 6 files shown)


michellephung:
+1

robertkowalski:
merged as fb8387ea6119d4ed7d409d6935098bb0eaacc899 and 8260f0854e90ca3594b166d3cfb08e80a1a76ac3

PR #354, opened 2015-04-01
Allow customization of View loading spinners

benkeen:
By default, the core Fauxton code inserts a spinner for each view
while it's loading. This can be removed via the `disableLoader` param
but not customized. This PR lets you define an optional
`loaderStyles` property on your view to override the styles. The
use-case here is to provide a light-coloured loader on the dark-
background Doc Editor page.

I investigated using the new React spinner, but ran into problems
targeting views that weren't loaded yet in the page load lifecycle.
This seemed like a very elegant solution that tapped into what we 
already had.


michellephung:
+1 cool

benkeen:
Merged as e229a45

PR #360, opened 2015-04-04
Cluster configuration fixes for 2.0

robertkowalski:
In CouchDB 2.0 we don't have `_config` on the clustered interface. As user/admin management also uses _config, also hide it on a cluster.

We still support it on the backdoor ports.

closes COUCHDB-2601 COUCHDB-2599 COUCHDB-2390
COUCHDB-2390


app/addons/config/routes.js
```diff
@@ -76,11 +81,114 @@ function (app, FauxtonAPI, Config, Views, CORS) {
     },
 
     establish: function () {
-      return [this.configs.fetch()];
+      var deferred = FauxtonAPI.Deferred(),
+          checked = FauxtonAPI.isRunningOnBackdoorPort();
+
+      FauxtonAPI.when(checked).then(function (res) {
```
garrensmith: You don't need `FauxtonAPI.when` here. You only doing one fetch

app/addons/config/routes.js
```diff
@@ -76,11 +81,114 @@ function (app, FauxtonAPI, Config, Views, CORS) {
...
+
+    routes: {
+      'createAdmin': 'createAdmin'
+    },
+
+    establish: function () {
+      var deferred = FauxtonAPI.Deferred(),
+          checked = FauxtonAPI.isRunningOnBackdoorPort();
+
+      FauxtonAPI.when(checked).then(function (res) {
```
garrensmith: Same here, `FauxtonAPI.when` isn't needed.

app/addons/config/templates/infoscreen.html
```diff
@@ -0,0 +1,28 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+License for the specific language governing permissions and limitations under
+the License.
+*/ %>
+
+<div class="config-cluster-warning">
+  Configuration is just available on the so called Backdoor-Ports for each single node of
```
garrensmith: Would it be possible to list the available backdoor port urls?

robertkowalski:
todo: remove the caching

just "cache" the promise

garrensmith:
@robertkowalski this looks good once you remove the caching.

app/core/utils.js
```diff
@@ -124,7 +124,41 @@ function ($, _) {
...
+      }
+
+      data = window.sessionStorage.getItem(key);
+      try {
+        return JSON.parse(data);
+      } catch (e) {
+        return data;
+      }
+
+      return data;
```
sebastianrothbucher: minor - but: unreachable statement (can just be left out?)

app/addons/auth/test/baseSpec.js
```diff
@@ -35,20 +48,45 @@ define([
...
+      var stub = sinon.stub(FauxtonAPI.session, 'isAdminParty').returns(true);
+      var deferred = FauxtonAPI.Deferred();
+      sinon.stub(FauxtonAPI, 'isRunningOnBackdoorPort').returns(deferred);
+      Base.initialize();
+      deferred.resolve({runsOnBackportPort: false});
+
+      FauxtonAPI.session.trigger('change');
+
+      assert.ok(spy.calledOnce);
+      var args = spy.getCall(0).args[0];
```
sebastianrothbucher: add-then-remove is hard 2 test, quite possibly this is the best way. The only alternative I can think of is overwriting add and remove, keeping track of the "list" and checking auth is not in it

sebastianrothbucher:
though I didn't run it I like it - and it's a big step 2wards a usable 2.0

robertkowalski:
this got stale as user management was converted to react

PR #362, opened 2015-04-07
Part 2/2 for Mango:

robertkowalski:
Adds search functionality, makes the feature visible and enables
deletion of mango-created-indexes.

closes COUCHDB-2627


app/addons/documents/resources.js
```diff
@@ -254,6 +339,24 @@ function (app, FauxtonAPI, Documents, PagingCollection) {
...
+    url: function () {
+      return app.host + '/' + this.databaseId + '/_index/_bulk_delete';
+    },
+
+    createPayload: function (documents) {
+      var documentList = documents.map(function (doc) {
+        return '_design/' + doc._id;
+      });
+
+      console.log(documentList);
```
robertkowalski: oh debug code

app/addons/documents/index-results/actions.js
```diff
@@ -71,7 +71,49 @@ function (app, FauxtonAPI, ActionTypes, Stores, HeaderStores, HeaderActions, Doc
...
+            type: 'error'
+          });
+
+          this.resultsListReset();
+        }.bind(this));
+    },
+
     reloadResultsList: function () {
+      if (indexResultsStore.getTypeOfIndex() === 'mango') {
+        return this.runMangoFindQuery({
```
garrensmith: Maybe split this into its own method called `reloadMangoResultsList`

app/addons/documents/resources.js
garrensmith: debug code

garrensmith:
This is looking good. Tests are failing though

app/addons/documents/index-results/actions.js
robertkowalski: i wish i could!

problem is that it is called internally from our bulkDeleteAction: https://github.com/robertkowalski/couchdb-fauxton/blob/mango-editor/app/addons/documents/index-results/actions.js#L168

robertkowalski:
@garrensmith added pagination etc...

michellephung:
+1

garrensmith:
@robertkowalski this looks great. One thing I've noticed is if I select __All Mango Indexes__ then the header for databases says Indexes and the cog disappears.

app/addons/documents/index-results/stores.js
```diff
@@ -96,17 +132,74 @@ function (FauxtonAPI, ActionTypes, HeaderActionTypes, Documents) {
...
+        url: this.isEditable(doc) ? doc.url('app') : null,
+        isDeletable: this.isDeletable(doc),
+        isEditable: this.isEditable(doc)
+      };
+
+    },
+
+    getResults: function () {
+      return this._collection
+        .filter(function (doc) {
```
garrensmith: If you break this into a separate function it would make it a lot clearer. Something like `getMangoDocsFromCollection`

app/addons/documents/index-results/stores.js
```diff
@@ -96,17 +132,74 @@ function (FauxtonAPI, ActionTypes, HeaderActionTypes, Documents) {
...
+          return doc.get('language') !== 'query';
+        })
+        .map(function (doc, i) {
+          if (doc.get('def') || !Object.keys(doc.attributes).length) {
```
garrensmith: Is it possible to move this onto a function on the doc, something like `getMangoDoc`?

app/addons/documents/mango/mango.actions.js
```diff
@@ -41,18 +55,54 @@ function (app, FauxtonAPI, Documents, ActionTypes, Stores, IndexResultsActions)
...
+          type: ActionTypes.MANGO_NEW_QUERY_FIND_CODE_FROM_FIELDS,
+          options: {
+            fields: queryCode.index.fields
+          }
+        });
+
+        window.setTimeout(function () {
+          FauxtonAPI.navigate(url);
+          FauxtonAPI.addNotification({
+            msg:  'Feel free to search now.',
```
garrensmith: Maybe we should say something like `Index is ready for querying` or `Index is ready` or something like that. I'm not sure `Feel free to search now.` is the correct tone.

app/addons/documents/mango/mango.components.react.jsx
```diff
@@ -121,35 +277,39 @@ function (app, FauxtonAPI, React, Stores, Actions, ReactComponents) {
...
+          editorTitle={app.i18n.en_US['mango-title-editor']}
+          additionalIndexesText={app.i18n.en_US['mango-additional-indexes-heading']} />,
         el
       );
     },
-    removeHelpScreen: function (el) {
+    removeQueryEditor: function (el) {
       React.unmountComponentAtNode(el);
     },
     renderMangoIndexEditor: function (el) {
```
garrensmith: Can you update all of these render functions to the new supported `setComponent` in the RouteObjects.

app/addons/documents/index-results/stores.js
robertkowalski: Sadly it is not that easy and i would keep it like this (after some additional refactoring)

We have many different collections in this store (mangoIndexResults, AllDocs/Docs, MangoIndexes).

I just took a look how I could move that method to the corresponding collections. Many menthods called in the code that I would have to move to the collection is implemented in this store and depends on state in this store, e.g. `this.isCollapsed`.

Don't say it is impossible but it would be a major refactor of this store which I would like to avoid in this PR for the mango feature.

Maybe we should decide on moving away from collections completely and just use stores for our state/data-layer instead of mixing both collections and stores.

garrensmith: Ok thats fine. Thanks for trying.. I agree with moving away from collections.

garrensmith:
+1 nice work.

app/addons/documents/index-results/stores.js
```diff
@@ -96,17 +137,68 @@ function (FauxtonAPI, ActionTypes, HeaderActionTypes, Documents) {
...
+        header = MangoHelper.getIndexName(doc);
+
         return {
-          content: this.getDocContent(doc),
-          id: this.getDocId(doc),
-          keylabel: doc.isFromView() ? 'key' : 'id',
+          content: this.getMangoDocContent(doc),
+          header: header,
+          id: doc.getId(),
+          keylabel: '',
```
michellephung: can we take this out?

michellephung: (the `keylabel:''` line)

michellephung: actually i see it might be helpful to be consistent so it matches the objects

garrensmith:
@robertkowalski this is looking good. Two issue's on the Run a Mango Query. If I edit the query before the list of indexes I have loads then the query is reset. Also if I edit the query but don't click run. Instead I press the '<' then I get a modal asking me am I sure I should leave the page.

robertkowalski:
> Also if I edit the query but don't click run. Instead I press the '<' then I get a modal asking me am I sure I should leave the page.

that is a feature all our editors share - if you change the query / document / map function and don't save/run it it will ask you if you are sure to leave...

benkeen:
- there should be a space after the "Mango Query" label and the "?" icon
- the default mango query uses 2-spaces, but Ace's default behaviour is tabs (e.g. click ENTER - it'll automatically indent using a tab). Bit of an annoyance.

garrensmith:
+1 nice work.

benkeen:
One curious thing, when I add I create a new Mango index, the page show two items - the first of which doesn't have an option to select it. See screenshot: 

create-mango-index

[Also, should be a space before the "?" on the Index label :) ]

benkeen:
Repeating the same index name is okay?

double-index

robertkowalski:
> Repeating the same index name in "Available Indexes" is okay?

yes, you created the second one, the first one is the special index "all_docs" which is additionally undeletable

benkeen:
Ah okay, thanks @robertkowalski - I wasn't clear on what that meant. Other than the missing radio (not clear on that either), looks good. +1

michellephung:
+1 on the code and being able to delete documents from the query!! :D :+1: 

although it's working,  it's still a bit unclear how it's supposed to work. Maybe we can make some documentation of a specific example for people to understand better.

robertkowalski:
merged as 425fc9b

PR #364, opened 2015-04-07
Tweaks to nightwatch tests

benkeen:
- Increases the timeout to 20 seconds because very rarely it can timeout on the two tests.
- **maxWaitTime** global created & referenced by all tests
- couple of minor changes, like click() now clickWhenVisible()


michellephung:
+1

PR #368, opened 2015-04-09
Changes auto-update option added

benkeen:
This PR adds a "Auto-update changes list" option on the Changes
page, within the Filters tab. When enabled, it will automatically
update the page with whatever changes occur on the database.


app/addons/documents/changes/components.react.jsx
```diff
@@ -134,15 +135,30 @@ define([
...
     render: function () {
       return (
         <div className="tab-content">
           <div className="tab-pane active" ref="filterTab">
             <div className="changes-header js-filter">
+              <div className="changes-polling">
+                <input
+                  type="checkbox"
+                  id="changes-toggle-polling"
+                  checked={this.state.pollingEnabled ? 'checked="checked"': null}
```
michellephung: i think this can be:  `checked={this.state.pollingEnabled}`  

https://facebook.github.io/jest/docs/tutorial-react.html

benkeen: Ooh nice, thanks!

app/addons/documents/changes/actions.js
```diff
@@ -38,11 +45,59 @@ function (app, FauxtonAPI, ActionTypes) {
...
+        params.since = lastSeqNum;
+        params.feed = 'longpoll';
+        params.timeout = pollingTimeout;
+      } else {
+        params.limit = 100;
+      }
+
+      var query = $.param(params);
+      var db = app.utils.safeURLName(changesStore.getDatabaseName());
+      currentRequest = $.get('/' + db + '/_changes?' + query);
```
michellephung: +1 I like this

michellephung:
+1

app/addons/documents/changes/actions.js
```diff
@@ -38,11 +45,59 @@ function (app, FauxtonAPI, ActionTypes) {
...
+        type: ActionTypes.INIT_CHANGES,
         options: options
       });
+      this.getLatestChanges();
+    },
+
+    getLatestChanges: function () {
+      var params = {};
+
+      var lastSeqNum = changesStore.getLastSeqNum();
```
garrensmith: Instead of default value for getLastSeqNum of null rather make it return `now`. Changes supports the `now` parameter. Then you don't need to do the if statement and instead just set the `since`, `feed` and `timeout`.

garrensmith:
@benkeen this is very cool. Could you add a small notification for each new change that is added. Maybe something like it changes colour slightly just so a user notices that a new set of changes have come through.

benkeen:
@garrensmith that's a good idea, will do.

app/addons/documents/changes/actions.js
benkeen: I just tried this, but for the initial request it can't be a `longpoll` or I don't ever get the first batch of data returned. But I like using 'now' as a default - much cleaner - so I refactored the code a bit to only set it to longpoll once `currentRequest` has been set. Bit cleaner. [Will commit in a moment]

garrensmith:
@benkeen this looks great. I'm getting a strange behaviour if I view changes. Then make a few changes to the db and then click on the filter and check auto update. It then does a lot of updating.

If I view the changes page, then check auto update and then make a few changes to the db I don't get as many updates.

benkeen:
Ah, drat. Good catch. What's happening is that the first time you check the auto update feature it'll show all changes that have occurred since the page was originally loaded, rather than from that moment on which is what you'd expect. I'll fix it.

benkeen:
Squashed & merged.

PR #370, opened 2015-04-09
Databases view in react

sebastianrothbucher:
just wanted to let you guys know I started working on it - and it's far from complete. So do not merge just yet... (I stole a lot from michellephung's active tasks btw - thanks!)
Pagination is still missing, so is search for a DB and all tests (including the one for the tray that we can share). 
When I'm too slow pls. feel free move ahead (pls. let me know ;-) )


michellephung:
nice! 
this is looking pretty good so far, keep going!

sebastianrothbucher:
thanks! - still lacks some tests, then I guess we're good to go...

sebastianrothbucher:
Did the test also - would be great if you could have a look - thanks!

app/addons/databases/components.react.jsx
```diff
@@ -0,0 +1,271 @@
...
+          <JumpToDatabaseWidget />
+        </div>
+      );
+    }
+  });
+
+  var AddDatabaseWidget = React.createClass({
+
+    onTrayToggle : function () {
+      var that = this;
```
robertkowalski: we use bind and try to avoid this = that

app/addons/databases/components.react.jsx
```diff
@@ -0,0 +1,271 @@
...
+    },
+
+    onKeyUpInInput : function (e) {
+      if (e.which === 13) {
+        this.onAddDatabase();
+      }
+    },
+
+    onAddDatabase : function () {
+      var databaseName = $(this.refs.newDbName.getDOMNode()).val();
```
robertkowalski: that is how we would do it in jquery, take a look at https://github.com/apache/couchdb-fauxton/blob/master/app/addons/cors/components.react.jsx#L69 for how to solve that in react components without jquery

app/addons/databases/components.react.jsx
```diff
@@ -0,0 +1,271 @@
...
+      Actions.createNewDatabase(databaseName, this.refs.newDbTray.hide, FauxtonAPI.addNotification);
```
robertkowalski: no need to inject `FauxtonAPI.addNotification`

also no need to inject the callback `this.refs.newDbTray.hide` - you can update a store which notifies this component after a successful update

app/addons/databases/components.react.jsx
```diff
@@ -0,0 +1,271 @@
...
+  var JumpToDatabaseWidget = React.createClass({
+
+    componentDidMount : function () {
+      var that = this;
+      $(this.refs.searchDbName.getDOMNode()).typeahead({
+        source : databasesStore.getDatabaseNames(),
+        updater: function (item) {
+          that.jumpToDb(item);
+        }
+      });
```
robertkowalski: that = this vs bind

app/addons/databases/components.react.jsx
```diff
@@ -0,0 +1,271 @@
...
+      // turn of browser autocomplete
+      $(this.refs.searchDbName.getDOMNode()).attr("autocomplete", "off");
```
robertkowalski: why not just add `autocomplete=off` to the markup?

app/addons/databases/components.react.jsx
```diff
@@ -0,0 +1,271 @@
...
+    },
+
+    jumpToDb : function (dbname) {
```
robertkowalski: this method should be an action and should not use jquery

app/addons/databases/components.react.jsx
```diff
@@ -0,0 +1,271 @@
...
+        </div>
+      );
+    }
+  });
+
+  var DatabasePagination = React.createClass({
+
+    render : function () {
+      var page = this.props.page;
+      var total = this.props.total || databasesStore.getDatabaseNames().length;
```
robertkowalski: `databasesStore.getDatabaseNames().length` - please use this.state instead

garrensmith:
@sebastianrothbucher this is exciting. Thanks for working on it. @robertkowalski has given some nice recommendations. Keep going.

app/addons/databases/stores.js
```diff
@@ -0,0 +1,99 @@
...
+      ).error(function (xhr) {
+          var responseText = JSON.parse(xhr.responseText).reason;
+          notifyCallback({
+            msg: 'Create database failed: ' + responseText,
+            type: 'error',
+            clear: true
+          });
+        }
+      );
+    },
```
robertkowalski: you can't know but this async operation belongs to an action. all async stuff happens in actions for fauxton

app/addons/fauxton/components.react.jsx
```diff
@@ -84,9 +84,131 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
...
+        <ul className="pagination">
+          <li className={(page === 1 ? "disabled" : null)}><a href={prefix + Math.max(page - 1, 1) + suffix}>&laquo;</a></li>
+          {rangeItems}
+          <li className={(page < totalPages ? null : "disabled")}><a href={prefix + Math.min(page + 1, totalPages) + suffix}>&raquo;</a></li>
+        </ul>
+      );
+    }
+  });
+
+
```
robertkowalski: not sure while this file uses 
```
{
  property : 'attribute'
}
```

while we/you use 

```
{
  prop: 'foo'
}
```
elsewhere

app/addons/fauxton/components.react.jsx
```diff
@@ -84,9 +84,131 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
...
+        <div ref="myself" style={styleSpec} className={classSpec}>{this.props.children}</div>
+      );
+    }
+  });
+
+  var Pagination = React.createClass({
+
+    getInitialState: function () {
+      return {};
+    },
```
robertkowalski: i think we can delete it

app/addons/fauxton/components.react.jsx
```diff
@@ -84,9 +84,131 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
...
+      if (this.state.show) {
+        this.hide(done);
+      } else {
+        this.show(done);
+      }
+    },
+
+    componentDidMount : function () {
+      var that = this;
+      $('body').on('click.Tray-' + this.props.trayid + "-" + this._rootNodeID, function (e) {
```
robertkowalski: I think we don't need a trayid - you can take `this.props.className`

I thin we don't nee a _rootNodeID - there will just one tray at all for creating databases

app/addons/fauxton/components.react.jsx
```diff
@@ -84,9 +84,131 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
...
+    },
+
+    hide : function (done) {
+      var that = this;
+      $(this.refs.myself.getDOMNode()).velocity('reverse', FauxtonAPI.constants.MISC.TRAY_TOGGLE_SPEED, function () {
+        that.setState({show: false});
+        if (done) {
+          done(false);
+        }
+      });
```
robertkowalski: we use bind

robertkowalski:
looks very promising! i left a few comments

michellephung:
Hey @sebastianrothbucher,

We are trying not to use jQuery in the React code, because usually there is a good way to handle an event without it. I don't think this is written anywhere, since we just started building with React code. Sorry about that, if that was unclear.

However! for the tray/tooltip, it's okay to use jQuery, since there's not another way to capture a `$('body').onclick` event or init the tooltip. 

Also, @garrensmith just added https://github.com/apache/couchdb-fauxton/commit/02a5ee3f0067432d3abf4fd683397ed6d66137ab, so we can render/remove the react components from the routes file.  I also changed my Active Tasks PR to reflect this change, and I'll wait for Garren's feedback before I merge it in, but something to keep in mind while your making changes.

app/addons/databases/components.react.jsx
sebastianrothbucher: I'm with you with FauxtonAPI.addNotification
However, I'd keep the hide callback: using another action dispatch just to hide the tray is a lot of overhead - and it gets more brittle: 1.) right now you have one tray, but for how long? 2.) When passing in, you already KNOW which react node you're dealing with, you don't have to remember for an event coming in and 3.) you might not want to change behavior - which right now is: hide the tray in the middle of things (when all trivial checks were performed and going back is possible).
So, I'd leave the callback here (plus: testing is also quite simple)

app/addons/databases/components.react.jsx
sebastianrothbucher: you can't - React will drop the attribute. And we're doing JQuery here anyway. As long as we don't tynamically update the input field I think we're very OK and won't break things. We just have to remember not to bind anything to the input dynamically via react. Otherwise there would be "data-autocomplete='off'" - but still: we're using JQuery anyway and data-... does at least have a similar chance of getting in the way...

app/addons/fauxton/components.react.jsx
sebastianrothbucher: actually: no - you'll break the logic, removing these three lines will crash the UI (or you just drop the possibility to set some things via state also and not just via props)

app/addons/fauxton/components.react.jsx
sebastianrothbucher: For Databases I'm with you, but when you want to make the tray reusable, I think it's not the worst idea to make it very sure we're distinct...

sebastianrothbucher:
thanks for all your feedback - I incorporated it (also the react routes) to the best of my know...

app/addons/fauxton/components.react.jsx
```diff
@@ -84,9 +84,129 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
...
+    toggle: function (done) {
+      if (this.state.show) {
+        this.hide(done);
+      } else {
+        this.show(done);
+      }
+    },
+
+    componentDidMount: function () {
+      $('body').on('click.Tray-' + this.props.trayid + "-" + this._rootNodeID, function (e) {
```
robertkowalski: be careful regarding `this._rootNodeID`, it will break as soon as we update to 0.13:

from: https://facebook.github.io/react/blog/2015/03/10/react-v0.13.html

```
Access to most internal properties has been completely removed, including this._pendingState and this._rootNodeID.
```

robertkowalski:
one small comment - this is coming along very nicely! 

also :heart: the testcoverage - this is awesome work!

app/addons/fauxton/components.react.jsx
sebastianrothbucher: I found a better solution for that one - thanks!
BTW: while testing I realized that TestUtils.renderIntoDocument drops our container. So we can get more stable results by just using React.renderComponent

app/addons/fauxton/components.react.jsx
```diff
@@ -84,9 +84,129 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
...
+    getDefaultProps: function () {
+      return {
+        perPage: FauxtonAPI.constants.MISC.DEFAULT_PAGE_SIZE,
+        page: 1,
+        total: 0
+      };
+    },
+
+    render: function () {
+      function getVisiblePages (page, totalPages) {
```
michellephung: you can pull this function out of the render, and add it to the Pagination object

app/addons/fauxton/components.react.jsx
michellephung: can you explain what you mean @robertkowalski ?

sebastianrothbucher:
thanks! - actually we lost the spinner when getting rid of views.js (was tied to establish), I have to put it back in 2morrow...

michellephung:
theres a component for that ! :)

loadinglines:  https://github.com/apache/couchdb-fauxton/blob/master/app/addons/documents/index-results/index-results.components.react.jsx#L81,

although if you prefer the look of the spinner, i don't mind

sebastianrothbucher:
@michellephung the LoadLines are too cool to not use them, so I did that - I still have to take care of your other comment though

app/addons/fauxton/components.react.jsx
robertkowalski: it was related to the previous formatting of the code, take a look at the whitespace around the `:`

sebastianrothbucher:
OK, did the last comment also - you think this is OK now?

app/addons/databases/components.react.jsx
```diff
@@ -0,0 +1,303 @@
...
+      );
+    }
+  });
+
+  var DatabaseTable = React.createClass({
+
+    createRows: function () {
+      return _.map(this.props.body, function (item, iteration) {
+        return (
+          <DatabaseRow row={item} />
```
michellephung: <DatabaseRow row={item} key={iteration} />

app/addons/fauxton/components.react.jsx
```diff
@@ -83,10 +83,132 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
...
+        );
+      }
+      return (
+        <ul className="pagination">
+          <li className={(page === 1 ? "disabled" : null)}><a href={prefix + Math.max(page - 1, 1) + suffix}>&laquo;</a></li>
+          {rangeItems}
+          <li className={(page < totalPages ? null : "disabled")}><a href={prefix + Math.min(page + 1, totalPages) + suffix}>&raquo;</a></li>
+        </ul>
+      );
+    }
```
michellephung:
```
<li key={i} className={(page === i ? "active" : null)} >
   <a href={prefix + i + suffix} >{i}</a>
</li>
```

to get rid of console error

app/addons/fauxton/components.react.jsx
```diff
@@ -83,10 +83,132 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
...
+      var perPage = this.props.perPage;
+      var prefix = this.props.urlPrefix || "";
+      var suffix = this.props.urlSuffix || "";
+      var totalPages = total === 0 ? 1 : Math.ceil(total / perPage);
+      var visiblePages = this.getVisiblePages(page, totalPages);
+      var rangeItems = [];
+      for (var i = visiblePages.from; i < visiblePages.to; i++) {
+        rangeItems.push(
+          <li className={(page === i ? "active" : null)}><a href={prefix + i + suffix}>{i}</a></li>
+        );
```
michellephung:
```
<li key={i} className={(page === i ? "active" : null)} >
   <a href={prefix + i + suffix} >{i}</a>
</li>
```

to get rid of console error

michellephung: you might be able to pull this for-loop out of render and into it's own function as well. I haven't tested this, but if you can pull it out, it'd be nice :)

michellephung:
if you're feeling up to it, some whitespace nits:

in the render functions, 
1.) there are some html code that have nested tags and create long lines of html, put each tag on a newline (the nested tags that are short don't need new lines, but for the longer ones, its easier and faster to review and debug)
2.) and for very long tags with lots of attributes, put each attribute on a new line, for example:
https://github.com/apache/couchdb-fauxton/blob/master/app/addons/documents/index-editor/components.react.jsx#L96

other than adding the key prop, 
tests look good :)  code logic looks good :)

FWIW, I am plus one after adding the key attribute to those 2 files, and white space changes are a "nice to have".

app/addons/databases/actions.js
```diff
@@ -0,0 +1,105 @@
...
+      FauxtonAPI.dispatch({
+        type: ActionTypes.DATABASES_STARTLOADING,
+        options: {}
+      });
+    },
+
+    setLoadComplete: function () {
+      FauxtonAPI.dispatch({
+        type: ActionTypes.DATABASES_LOADCOMPLETE,
+        options: {}
```
garrensmith: Since the object is empty do we need this property?

app/addons/databases/actions.js
```diff
@@ -0,0 +1,105 @@
...
+      });
+    },
+
+    createNewDatabase: function (databaseName, nameAccCallback) {
+      databaseName = (databaseName === null ? null : databaseName.trim());
```
garrensmith: could you change this from `databaseName === null` to `_.isNull(databaseName)`. I think an if statement might be nicer here.

app/addons/databases/actions.js
```diff
@@ -0,0 +1,105 @@
...
+      if (databaseName === null || databaseName.length === 0) {
+        FauxtonAPI.addNotification({
+          msg: 'Please enter a valid database name',
+          type: 'error',
+          clear: true
+        });
+        return;
+      }
+      // name accepted, make sure prompt can be removed
+      nameAccCallback();
```
garrensmith: This doesn't really work, callbacks violate the flux one direction flow.
I think what is better is to have a property on the store that maintains the state of the prompt. And rather change that which then causes the React Component to hide the prompt.

app/addons/databases/actions.js
```diff
@@ -0,0 +1,105 @@
...
+
+      var db = Stores.databasesStore.obtainNewDatabaseModel(databaseName);
+      FauxtonAPI.addNotification({ msg: 'Creating database.' });
+      db.save().done(function () {
+          FauxtonAPI.addNotification({
+            msg: 'Database created successfully',
+            type: 'success',
+            clear: true
+          });
+          var route = '#/database/' + app.utils.safeURLName(databaseName) + '/_all_docs?limit=' + Resources.DocLimit;
```
garrensmith: This needs to be updated to use the `Fauxton.urls` method. @michellephung could you help with this.

app/addons/databases/actions.js
```diff
@@ -0,0 +1,105 @@
...
+            type: 'error',
+            clear: true
+          });
+        }
+      );
+    },
+
+    jumpToDatabase: function (databaseName) {
+      databaseName = databaseName === null ? null : databaseName.trim();
+      if (databaseName === null || databaseName.length === 0) {
```
garrensmith: You could change this to:

```javascript
if (_.isNull(databaseName) || _.isEmpty(databaseName) {...
```

Makes it a little more readable.

app/addons/databases/actions.js
```diff
@@ -0,0 +1,105 @@
...
+        return;
+      }
+      if (Stores.databasesStore.getDatabaseNames().indexOf(databaseName) >= 0) {
```
garrensmith: Could you make this a function on the databasesStore. Something like `Stores.databasesStore.doesDatabaseExist(databaseName)`

app/addons/databases/components.react.jsx
```diff
@@ -0,0 +1,303 @@
...
+    componentWillUnmount: function () {
+      databasesStore.off('change', this.onChange, this);
+    },
+
+    onChange: function () {
+      this.setState(this.getStoreState());
+    },
+
+    render: function () {
+      var collection = this.state.collection || {};
```
garrensmith: Instead of having these initialisations here. Could you move them to the store. And the store can either return the collection or an empty array if the collection hasn't been fetched yet.

app/addons/databases/routes.js
```diff
@@ -46,43 +41,37 @@ function (app, FauxtonAPI, Databases, Views, Components) {
...
+      return this.databases.slice(start, end);
+    },
 
-      this.databasesView.setPage(dbPage);
+    allDatabases: function () {
+      var params = app.getParams();
+      this.page = params.page ? parseInt(params.page, 10) : 1;
+      this.perPage = FauxtonAPI.constants.MISC.DEFAULT_PAGE_SIZE;
+
+      Actions.setStartLoading();
```
garrensmith: Could you move all of this fetching code into an action.

app/addons/databases/routes.js
```diff
@@ -46,43 +41,37 @@ function (app, FauxtonAPI, Databases, Views, Components) {
...
-      this.databasesView = this.setView("#dashboard-content", new Views.List({
-        collection: this.databases,
-        perPage: perPage,
-        page: dbPage
-      }));
-
-      this.rightHeader = this.setView("#right-header", new Views.RightAllDBsHeader({
-        collection: this.databases,
-      }));
+    paginated: function () {
```
garrensmith: I think this paginated could go into the collection.

app/addons/fauxton/components.react.jsx
```diff
@@ -83,10 +83,132 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
...
+    render: function () {
+      var page = this.state.page || this.props.page;
+      var total = this.state.total || this.props.total;
+      var perPage = this.props.perPage;
+      var prefix = this.props.urlPrefix || "";
+      var suffix = this.props.urlSuffix || "";
+      var totalPages = total === 0 ? 1 : Math.ceil(total / perPage);
+      var visiblePages = this.getVisiblePages(page, totalPages);
+      var rangeItems = [];
+      for (var i = visiblePages.from; i < visiblePages.to; i++) {
```
garrensmith: Could you change this for loop to use (_.range)[http://underscorejs.org/#range]. Something like:
```javascript
return _.range(visiblePages.from, visiblePages.to).map(function (i) {
   return <li className={(page === i ? "active" : null)}><a href={prefix + i + suffix}>{i}</a></li> 
});
```

I also think this would be better in a function something like `createDBItemsForPage()`

garrensmith:
@sebastianrothbucher wow this is amazing work. Thank you. I've left a fair amount of comments, just somethings to get it more consistent with how we are using React. You are almost there. Please let me know if I can help you with any of it or if my comments don't make sense.

garrensmith:
@sebastianrothbucher I also found one bug. If I go to page 2 and then click new database, it then takes me back to page 1. I think you might need to add `data-bypass=true` for the new database link

app/addons/databases/actions.js
sebastianrothbucher: It depends on whether we just always want to have options - if we do it case-by-case we can just kick them out.

app/addons/databases/actions.js
michellephung: ```var route = FauxtonAPI.urls('allDocs', 'app', app.utils.safeURLName(databaseName), '?limit=' + Resources.DocLimit);```

a bit of context:
A couple of months ago we changed how to retrieve generalized urls in Fauxton.
https://github.com/apache/couchdb-fauxton/pull/258

there started to be several ways one could build a url string:
-models.url(type);
-string building
-sequences of if-elseif statements

so we centralized it to keep things uniform across the board.

If you need to create a urls() function, you'd register/specify it in its respective base.js file.
For Example, for documents urls: https://github.com/apache/couchdb-fauxton/blob/master/app/addons/documents/base.js

benkeen:
Hey @sebastianrothbucher - one more issue. To reproduce:
- Create a new database - which will redirect you to the new database's All Docs page.
- Click the "<" icon to return to the Databases page.
- Click the "Add Database" again. The page just reloads and you don't see the Add Database tray. The URL hash changes to `#` - looks like the JS isn't firing.

To second everyone else here: really awesome work! If you want a hand with any of this, let me know.

app/addons/databases/components.react.jsx
```diff
@@ -0,0 +1,303 @@
...
+
+    componentDidMount: function () {
+      $(this.refs.myself.getDOMNode()).tooltip();
+    },
+
+    render: function () {
+      var row = this.props.row;
+      return (
+        <i className="js-db-graveyard icon icon-exclamation-sign" ref="myself" title={"This database has just " + row.status.numDocs() + " docs and " + row.status.numDeletedDocs() + " deleted docs"}></i>
+      );
```
benkeen: We generally try to keep row length to <= 120 chars.

sebastianrothbucher:
this is great feedback - thanks! Give me a few days... ;-) (and I'll come back to you when stuck)

sebastianrothbucher:
did most of it - but the bug persists. data-bypass is not it
Leaving out href does it - strange. Need to investigate further
The failing travis is somewhere else btw - I guess a wait was added (otherwise I'll just do it)

sebastianrothbucher:
maybe I did some subtle mistakes yesterday. Now we get it done with data-bypass (thx to @garrensmith for the hint!). Feedback is included now...

app/addons/databases/components.react.jsx
```diff
@@ -0,0 +1,321 @@
...
+    componentDidMount: function () {
+      databasesStore.on('change', this.onChange, this);
+      $(this.refs.searchDbName.getDOMNode()).typeahead({
+        source: this.state.databaseNames,
+        updater: function (item) {
+          this.jumpToDb(item);
+        }.bind(this)
+      });
+      // turn of browser autocomplete (direct markup is not passed on)
+      $(this.refs.searchDbName.getDOMNode()).attr("autocomplete", "off");
```
robertkowalski: no need for the hack, as promised i searched: it is `autoComplete` (note the camelCase): https://facebook.github.io/react/docs/tags-and-attributes.html#html-attributes

app/addons/databases/actions.js
```diff
@@ -0,0 +1,123 @@
...
+              backboneCollection: databases,
+              page: page
+            }
+          });
+        }.bind(this));
+      }.bind(this));
+    },
+
+    setPage: function (_page) {
+      FauxtonAPI.dispatch({
```
benkeen: Small thing. We don't use underscore prefixes for local vars. Mind just calling it `page`?

robertkowalski:
hey @sebastianrothbucher i think after you fixed the last two small issues and rebased the branch we can merge it

sebastianrothbucher:
K, great

robertkowalski:
@sebastianrothbucher yes, we all feel sorry for making this a long process for you.

benkeen:
+1 on Robert's last remark. :)

sebastianrothbucher:
it's merged - thanks for all your great feedback and for cheering on!

benkeen:
Great work @sebastianrothbucher :D Great to see this merged.

PR #373, opened 2015-04-09
Docs now fully collapse to single line

benkeen:
This reduces selected documents to a single line containing
only the ID, rather than show a subset of the doc with an ID
and rev.

Closes COUCHDB-2548


michellephung:
+1

benkeen:
Merged as f95b7a07f0eb4f13c59e1812a13517fe8ee4c6b7

PR #374, opened 2015-04-09
Prevent overlap of text and arrows in styled-select

benkeen:
Super minor thing. If an option in a styled select box
(.styled-select) is quite long, it will overlap the down arrow
overlay which looks kind of bad. Unfortunately browsers are
limited in that you can't put an ellipsis, so this just adds
padding instead to cut off the text, which still looks a lot
better.

### DIFF
--- a/app/addons/components/assets/less/styled-select.less
+++ b/app/addons/components/assets/less/styled-select.less
@@ -24,10 +24,10 @@
   appearance: none;
   background-color: #e6e6e6;
   border: 1px solid #b3b3b3;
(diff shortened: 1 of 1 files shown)


benkeen:
Before: 
screen shot 2015-04-09 at 11 16 11 am

After: 
screen shot 2015-04-09 at 11 16 01 am

michellephung:
+1

benkeen:
Merged as 93aa09fe288e910d2b87021f7968559058107990

PR #377, opened 2015-04-09
Beautify tooltip bug fix

benkeen:
Fixes a small bug where the tooltip for the Beautify button on the
Edit Index page sometimes doesn't disappear after clicking it.
Also moved the tooltip to the right to prevent overlapping, reduced
the button size a little (it was larger than the main buttons on
the page) and removed a wayward semicolon.

### DIFF
--- a/app/addons/components/react-components.react.jsx
+++ b/app/addons/components/react-components.react.jsx
@@ -89,7 +89,7 @@ function (app, FauxtonAPI, React, Components, beautifyHelper) {
             target="_blank"
           >
           <i className="icon-question-sign"></i>
-          </a>;
+          </a>
         </label>
       );
     },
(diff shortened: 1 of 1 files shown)


robertkowalski:
code looks good but i can't get the button displayed: https://cloudup.com/cy8qL_nN-EN

am i doing something wrong?

benkeen:
Hey @robertkowalski. Yeah, I noticed that too, but on reflection I think it's probably working as it should. It only shows the button if on page load the code is a single line, not when you edit it on the fly. The reason I think that's probably the best behaviour is that when editing the JS content, they may reduce it to a single line (e.g. deleting all lines but the first when re-writing the function). Having the button appear at that point would be a bit weird.

robertkowalski:
not in this scope, but i think presenting the button always might be a good idea and helps also when creating a completely new view where i copy&paste stuff

back to the issue: +1 :)

benkeen:
That might not be a bad idea. My only concern is when there's nothing to beautify and nothing happens when the user clicks it. Wonder if we could detect that scenario and disable the button... but yeah, another ticket for that. :)

benkeen:
Merged as 7d0d2dc

PR #378, opened 2015-04-11
Avoid overlap of footer with scrollbar

sebastianrothbucher:
have some space right of footer for display of scroll - pareto solution


robertkowalski:
that's not quite right, here is what it looks like on a mac:

bildschirmfoto 2015-04-13 um 11 08 48

garrensmith:

PR #382, opened 2015-04-15
Doc Workflow improvements

benkeen:
This PR is part of Doc Workflow to improve the UX of navigating around the UI with respect to managing the documents.
### What's changed
1. Clicking back/cancel from a full page doc editor no longer limits the results to 20 on the doc list page.
2. Clicking back to a list of documents from the full page doc editor now returns you to the last PAGE you were on, and the SCROLL offset of wherever you were on the page as well so it's way easier to pick up where you left off.
3. Introduces fauxton/memory.js as a way to store information across multiple pages.
### Notes
- still missing better tests. Waiting on Robert's feedback regarding the change to pagingcollection before I really "commit" to this ticket (hah!)
- there’s an existing bug where when you click from one page to the next it doesn’t update the scrollTop to 0 on the scrollable. That’s unrelated to this ticket.
### Big flaw
- The `cloudant.pagingcollection.js` file change. Why is that a separate external lib? I ran into serious problem trying to persuade it to re-initialize on a non-first page and couldn’t get around it with the way the code was written, hence the modification to the file. There is a solution around it by circumventing it in the store, but that felt very hacky & not addressing the root cause.


app/addons/documents/routes-documents.js
```diff
@@ -181,11 +194,14 @@ function (app, FauxtonAPI, BaseRoute, Documents, Changes, ChangesActions, DocEdi
       },
 
       cleanup: function () {
+
+        // to ensure garbage collection on the React components
+        this.removeComponent('#dashboard-lower-content');
```
garrensmith: This shouldn't be needed. the `FauxtonAPI.RouteObject.prototype.cleanup.apply(this);` should do that for you

benkeen: I added that because none of the React components were unmounting when leaving the page.

app/addons/fauxton/components.js
```diff
@@ -163,6 +164,11 @@ function (app, FauxtonAPI, ace, spin, ZeroClipboard) {
     initialize: function (options) {
       this.crumbs = options.crumbs;
       this.toggleDisabled = options.toggleDisabled || false;
+      this.clickBack = options.clickBack || null;
+    },
+
+    clickBack: function (e) {
+      this.clickBack && this.clickBack(e);
```
robertkowalski: this.clickBack in line 167 and line 170 don't make sense

benkeen:
Hi @robertkowalski - you mentioned you're pretty familiar with that pagination lib. I explored it some more, but don't see a loophole to get it to properly set `this.paging.hasPrevious` to be true when you're starting on a page other than 1. Any suggestions?

I also investigated updating the lib itself, but none of the unit tests ran for me [possibly outdated dependencies?].

robertkowalski:
Maybe it is possible to use the new react router with just parts of Fauxton so we can incrementally switch over. We then could add it to the doclists at first. 

Let's chat at the summit about it!

benkeen:
** We spoke about this. This PR is going to be put on hold for the moment. I'll leave it open. **

garrensmith:

PR #385, opened 2015-04-15
ESC key closes any open notification

benkeen:
This PR just lets users use their keyboard to close notifications.
We discussed doing the same with mouse clicks and mouse overs, but
they were a bit more contentious so I figure we can add them later
if we want.


michellephung:
+1

benkeen:
Merged as 5920016

PR #390, opened 2015-04-17
Updating Auth module to use React

benkeen:
One thing to discuss: I updated `setComponent()` to allow calling code to pass props to the React component via a 3rd param. Seemed a reasonable approach, and will allow us to pass props rather than having to do an action and store that info in a store.


app/addons/auth/actions.js
```diff
@@ -0,0 +1,116 @@
...
+
+    login: function (username, password, session, urlBack) {
+      var promise = session.login(username, password);
+
+      promise.then(function () {
+        FauxtonAPI.addNotification({ msg: FauxtonAPI.session.messages.loggedIn });
+        if (urlBack) {
+          return FauxtonAPI.navigate(urlBack);
+        }
+        FauxtonAPI.navigate('/');
```
robertkowalski: cool, i love early returns!

app/addons/auth/actions.js
```diff
@@ -0,0 +1,116 @@
...
+        type: ActionTypes.AUTH_UPDATE_CHANGE_PWD_CONFIRM_FIELD,
+        value: value
+      });
+    },
+
+    createAdmin: function (username, password, session, loginAfter) {
+      var promise = session.createAdmin(username, password, loginAfter);
+
+      promise.then(function () {
+        FauxtonAPI.addNotification({ msg: FauxtonAPI.session.messages.adminCreated });
```
robertkowalski: I noticed that `session` is injected and also used from our `FauxtonAPI`

I was not aware that session is available on the `FauxtonAPI`-object - I think we can just use `FauxtonAPI.session` everywhere? Are there reasons against using it?

app/addons/auth/components.react.jsx
```diff
@@ -0,0 +1,299 @@
...
+      createAdminSidebarStore.on('change', this.onChange, this);
+    },
+
+    componentWillUnmount: function () {
+      createAdminSidebarStore.off('change', this.onChange);
+    },
+
+    selectPage: function (e) {
+      var newPage = e.target.href.split('#')[1];
+      AuthActions.selectPage(newPage);
```
robertkowalski: that's a responsibility of our router, can you use the main router for that?

robertkowalski: ah i understand now...

app/addons/auth/actions.js
benkeen: You know, that's a fair point. I agree - I'll remove the explicit pass-by-props.

robertkowalski:
awesome work ben!

feel free to steal the test for the setComponent/routeobject change from my mango branch (which stole the implementation for the props from your branch) --> https://github.com/robertkowalski/couchdb-fauxton/blob/mango-editor/app/core/tests/routeObjectSpec.js#L184-190

+1 after adding a test for the routobject change

benkeen:
Thanks @robertkowalski! I'll add that now.

benkeen:
Merged as 410d2c7

PR #392, opened 2015-04-22
Fix for DB typeahead link

benkeen:
This adds a test to confirm the db name typeahead field in the
header of the Database page works as expected, and that the URL
it redirects to doesn't include an invalid 'undefined' string at
the end of it.

Closes COUCHDB-2671


garrensmith:
+1

benkeen:
Thanks Garren. Merged as e9f4330

PR #393, opened 2015-04-22
Verify install page converted to React

benkeen:
This converts the Verify Install page to React.

@robertkowalski, the one thing I had trouble with was writing a decent test for `actions.startVerification()`. I couldn't override the model that does all the async tests to the server. So instead, this relies on the  other store, React + existing nightwatch tests confirming the functionality.


app/addons/verifyinstall/components.react.jsx
```diff
@@ -0,0 +1,145 @@
...
+      var result = '';
+      if (!this.props.testResults[test].complete) {
+        result = '';
+      } else if (this.props.testResults[test].success) {
+        result = <span>&#10003;</span>;
+      } else {
+        result = <span>&#x2717;</span>;
+      }
+      return result;
+    },
```
robertkowalski: you can use early returns here instead of nesting ifs/elses

app/addons/verifyinstall/stores.js
```diff
@@ -0,0 +1,101 @@
...
+    reset: function () {
+      this._isVerifying = false;
+
+      // reset all the tests
+      this._tests = {};
+      var keys = ['CREATE_DATABASE', 'CREATE_DOCUMENT', 'UPDATE_DOCUMENT', 'DELETE_DOCUMENT', 'CREATE_VIEW', 'REPLICATION'];
+      _.each(keys, function (key) {
+        this._tests[Constants.TESTS[key]] = { complete: false };
+      }, this);
+    },
```
robertkowalski: something like 

```
this._tests = Object.keys(Constants.TESTS).reduce(function (acc, el) {
  acc[Constants.TESTS[key]] = { complete: false };
  return acc;
}, {});
```

should do the trick so you don't have to hardcode all the keys again here

app/addons/verifyinstall/stores.js
```diff
@@ -0,0 +1,101 @@
...
+
+    checkIsVerifying: function () {
+      return this._isVerifying;
+    },
+
+    updateTestStatus: function (test, success) {
+
+      // shouldn't ever occur since we're using constants for the test names
+      if (!_.has(this._tests, test)) {
+        console.error("Invalid test name passed to updateTestStatus()");
```
robertkowalski: in that case i would just do:

```js
throw new Error('Invalid test name passed to updateTestStatus()');
```

app/addons/verifyinstall/tests/componentsSpec.js
```diff
@@ -0,0 +1,136 @@
...
+      container = document.createElement('div');
+      el = ReactTestUtils.renderIntoDocument(React.createElement(Components.VerifyInstallResults, {testResults: testResults}), container);
+
+      _.each(tests, function (test) {
+        assert.equal($(el.getDOMNode()).find('#' + test.id).html(), '');
+      });
+    });
+
+    it('confirm each result field shows success after successful test', function () {
+      _.each(tests, function (test) {
```
robertkowalski: Sidenote: I don't get the overuse of `_.each` as we have native `.forEach` in JS for years:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

app/addons/verifyinstall/stores.js
benkeen: Good point.

app/addons/verifyinstall/tests/actionsSpec.js
```diff
@@ -0,0 +1,33 @@
...
+  describe('Verify Install Actions', function () {
+
+    it('resets the store when action called', function () {
+      var spy = sinon.spy(Stores.verifyInstallStore, 'reset');
+      FauxtonAPI.dispatch({ type: ActionTypes.VERIFY_INSTALL_RESET });
+      assert.ok(spy.calledOnce);
+    });
+
+  });
+
```
sebastianrothbucher: Testing the actual run would be lots of mocking - not sure whether it would be worth it

app/addons/verifyinstall/tests/componentsSpec.js
```diff
@@ -0,0 +1,136 @@
...
+    });
+
+    afterEach(function () {
+      React.unmountComponentAtNode(container);
+    });
+
+    it('confirm all result fields blank before tests ran', function () {
+      container = document.createElement('div');
+      el = ReactTestUtils.renderIntoDocument(React.createElement(Components.VerifyInstallResults, {testResults: testResults}), container);
+
```
sebastianrothbucher: Just a general remark: ReactTestUtils.renderIntoDocument ignores the container param - see https://github.com/facebook/react/blob/master/src/test/ReactTestUtils.js?source=c#L46

app/addons/verifyinstall/tests/componentsSpec.js
```diff
@@ -0,0 +1,136 @@
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
```
sebastianrothbucher: did not get it at first - but: is this file not a compilation result? and the next file is the actual one (i.e. this one shhould also not be checked in?)

app/addons/verifyinstall/tests/changesStoreSpec.js
```diff
@@ -0,0 +1,51 @@
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
```
sebastianrothbucher: Why "changesStore"?

app/addons/verifyinstall/tests/componentsSpec.js
sebastianrothbucher: normally, the "something.react.js" pattern might avoid it

sebastianrothbucher:
Looks great to me, +1 with the filename thing resolved (or made me get it)

app/addons/verifyinstall/tests/actionsSpec.js
benkeen: Exactly. I went down that route and ended up spending a lot of time fussing around with mocking - so decided against it & just included the end-to-end Nightwatch test.

app/addons/verifyinstall/tests/componentsSpec.js
benkeen: Oops, you're quite right. I'll remove this file.

benkeen:
Thanks @sebastianrothbucher! Good calls on the above.

app/addons/verifyinstall/stores.js
```diff
@@ -0,0 +1,99 @@
...
+    checkIsVerifying: function () {
+      return this._isVerifying;
+    },
+
+    updateTestStatus: function (test, success) {
+
+      // shouldn't ever occur since we're using constants for the test names
+      if (!_.has(this._tests, test)) {
+        throw new Error('Invalid test name passed to updateTestStatus()');
+      }
```
michellephung: nice!

michellephung:
+1 
note: i like how this was organized

benkeen:
Thanks @michellephung! Squashed and merged as 90c07fa

PR #394, opened 2015-04-24
Fix for right panel not being scrollable

benkeen:
Certain two-panel pages like the Create Admin page had overflow
hidden so content would get cut off an unreachable for smaller
browser widths. This patches it to only apply the overflow for pages
that need it (e.g. the doc list).


garrensmith:
+1

benkeen:
Merged as 49267a4

PR #396, opened 2015-05-04
Open up require's config.js file to be extended

benkeen:
This PR separates the content of the /app/config.js file into
a different file so scripts that extend Fauxton can choose to
load it first, then add/override whatever additional settings
it needs.

One nice thing is that it removes a hack for the mocha code 
which had to work around the fact that the config data 
was innaccessible in require.config();

Note: this PR doesn't include a test because there's nothing really
that can be tested. If it didn't work, everything would break 
immediately.

Closes COUCHDB-2678


app/config_source.js
```diff
@@ -0,0 +1,68 @@
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
```
garrensmith: We don't have `_` in file names.

benkeen: Ah sorry, I was making it compatible with load_addons.js. Will update in a sec.

garrensmith:
+1

benkeen:
Merged as fad972a

PR #400, opened 2015-05-07
Detect if we run on a backdoor port

robertkowalski:
This enables us to find out in Fauxton if we run on a backdoor
port.

COUCHDB-2390

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -126,7 +126,30 @@ function (app, $, _, Backbone, Bootstrap, Helpers, Utils, FauxtonAPI, Couchdb) {
           type: 'REMOVE_NAVBAR_LINK',
           link: link
       });
+    },
+
+    isRunningOnBackdoorPort: function () {
+
+      if (this._backdoorDeferred) {
+        return this._backdoorDeferred;
+      }
+
+      this._backdoorDeferred = FauxtonAPI.Deferred();
+
+      $.ajax({
+        type: 'GET',
+        url: app.host + '/_config'
+      })
+      .then(function () {
+        this._backdoorDeferred.resolve({runsOnBackportPort: true});
+      }.bind(this))
+      .fail(function (res) {
+        this._backdoorDeferred.resolve({runsOnBackportPort: false});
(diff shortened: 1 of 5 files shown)


app/app.js
```diff
@@ -126,7 +126,30 @@ function (app, $, _, Backbone, Bootstrap, Helpers, Utils, FauxtonAPI, Couchdb) {
           type: 'REMOVE_NAVBAR_LINK',
           link: link
       });
+    },
+
+    isRunningOnBackdoorPort: function () {
```
kxepal: I think the name is too broad. If...when.../_config will be available on cluster interface it will be easy to break all the logic behind that name. 

I would suggest to use some specific like "isConfigAvailable" to control features availability per resource basis or use some cluster-only resource like `_membership` or `_cluster_setup` e.g. those that only makes a sense for cluster interface. 

Alternatively, you may figure if you're on cluster or not from `GET /` request - cluster has no uuid in response, but that's may be also temporary solution.

robertkowalski: for now it is the feature detection if we run on a backdoor port or not

we'll have to change fauxton anyway if we get `_config` back - if ...

robertkowalski: `_cluster_setup ` is a good idea

robertkowalski: thanks for the hint, we are using `_cluster_setup` now

kxepal: Good! (:

michellephung:
how to test?

robertkowalski:
you can:

a) run the testsuite
b) on couchdb 2 master open fauxton on 15984, logged in and logged out, validate the return values of the promise. then do the same for port 15986 (backdoor ports), validate the return values of the promise

one way to run fauxton on different ports is to change the port in https://github.com/apache/couchdb-fauxton/blob/master/tasks/helper.js#L17

garrensmith:
+1 looks good

robertkowalski:
merged as 9ab1292d4e926ecba7c15291a04feb0342b40b21

PR #408, opened 2015-05-12
Improvement to NW tests

benkeen:
Just adds a couple of extra checks to ensure the element is both present + visible. The previous code didn't check for visibility.


michellephung:
+1

PR #415, opened 2015-05-14
JumpToDB component now updates DB list after initial render

benkeen:
Small change. Before, the component would only render the list of DB in the typeahead field if it was passed the data on initial mount. This now sets it on subsequent re-renders.


michellephung:
+1

benkeen:
Thanks @michellephung!

PR #416, opened 2015-05-14
JumpToDB component now updates DB list after initial render

benkeen:


benkeen:
This went green and was +1'd here: https://github.com/apache/couchdb-fauxton/pull/415

This just puts it in a named branch.

benkeen:
Merged as e49fbea

PR #417, opened 2015-05-14
Fix for switchDatabase.js test

benkeen:
This is a patch for a buggy test.

It slows down the test to wait for the body of the databases
table to have been loaded before redirecting. That ensures that
the databases have actually been loaded, so the jumptodb option
has a data source loaded. Formerly it would *probably* work.


michellephung:
+1

benkeen:
Merged as ceb51a73896d22350ebcd4d969f0eaa48ab2e6ef

PR #418, opened 2015-05-14
Improving robustness of bulkDelete.js

benkeen:
This test has been known to fail. This slows it down a bit to ensure
elements are present and visible before they are clicked. I was able
to see this test fail locally; this has fixed it for me.


michellephung:
+1

benkeen:
Merged as 13b760b4187a235f260907696405f22268b4ed2e

PR #421, opened 2015-05-19
Permissions react

garrensmith:
Permissions for a database done in React


benkeen:
Code looks great, but I'll look it over a little longer. One minor UX niggle: after adding an item in any of the four sections on the Permissions page, it should probably remove the text you just entered and re-focus on the field.

michellephung:
+1

benkeen:
Besides previous comment, +1

app/addons/permissions/components.react.jsx
```diff
@@ -0,0 +1,221 @@
...
+        type: 'roles',
+        section: this.props.section,
+        value: this.state.newRole
+      });
+
+      this.setState({newRole: ''});
+    },
+
+    getItems: function (items, type) {
+      return _.map(items, function (item, i) {
```
robertkowalski: you can use native map, i think we can remove underscore/lodash in the future as those features are all now implemented into browsers.

http://kangax.github.io/compat-table/es5/#Array.prototype.map

michellephung:
if your going to change the UI, I'd reword the sub-line under "Roles" to:
 `All users with any of the following role(s) will have admins access.`,
or
`Users with the following role(s) will have admins access.`

it's unclear if you need all of the roles to have admin access, or just one of the roles.

michellephung:
i think it's `any of the following`

benkeen:
Sorry to be a pain, one more minor thing: it allows you to add a blank entry. I think this is an old bug. The UI looks a little wonky due to the empty user name. 

screen shot 2015-05-19 at 9 24 49 am

garrensmith:
Issued fixed and merged.

PR #426, opened 2015-05-20
Fix View Results with same id not showing

garrensmith:
If a view emits the same doc multiple times only the first doc was
shown. This fixes that.


robertkowalski:
two super minor nits, merge it after adressing them! :)

garrensmith:
Issued fixed. And merged.

PR #430, opened 2015-05-20
Index editor fixes

benkeen:
This makes two small fixes to the index-editor component.
- the database name now shows up in the page
- the database name now links to the database and doesn't throw a
JS error in the page


robertkowalski:
i added a fix for mango

benkeen:
Cool thanks. I'll squash - Travis seems to have choked on that last one.

michellephung:
+1

PR #435, opened 2015-05-25
Convert Design Doc Metadata page to React

garrensmith:
This converts the Design Doc Metadata page to use React.js instead of
Backbone.js


benkeen:
Very nice! Code + functionality looks great. +1

PR #438, opened 2015-05-26
Prevent duplicate mocha tests running

benkeen:
When running mocha for environments where Fauxton has been
extended, e.g. the Cloudant dashboard, tests unique to that other
environment get run twice because the original tests are included
once from the original folder and once from where they've been
copied - in the Fauxton /app folder. Since we know that the files
would need to be copied over to the /app folder for inclusion
in Fauxton, this PR removes the extra tests from the original
source.

### DIFF
--- a/app/load_addons.js.underscore
+++ b/app/load_addons.js.underscore
@@ -16,7 +16,7 @@
  * THIS IS A GENERATED FILE. DO NOT EDIT.
  */
 define([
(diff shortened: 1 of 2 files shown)


garrensmith:
+1

benkeen:
Thanks. Merged as 0e34b78

PR #440, opened 2015-05-26
Databases component fix

benkeen:
This fixes a small problem with the mocha test for the databases
component. Because it didn't reset the store it would sometimes
interfere with other mocha tests running. Only seen with the dash.

### DIFF
--- a/tasks/fauxton.js
+++ b/tasks/fauxton.js
@@ -116,6 +116,12 @@ module.exports = function (grunt) {
     var fileSrc = grunt.option('file') || data.files.src;
     var testFiles =  grunt.file.expand(fileSrc);
 
+    // filter out any tests that aren't found in the /app/ folder. For scripts that are extending Fauxton, we still
+    // know that all addons will have been copied into /app. This prevent tests being ran twice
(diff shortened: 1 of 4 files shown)


tasks/fauxton.js
```diff
@@ -116,6 +116,12 @@ module.exports = function (grunt) {
     var fileSrc = grunt.option('file') || data.files.src;
     var testFiles =  grunt.file.expand(fileSrc);
 
+    // filter out any tests that aren't found in the /app/ folder. For scripts that are extending Fauxton, we still
+    // know that all addons will have been copied into /app. This prevent tests being ran twice
+    testFiles = _.filter(testFiles, function (filePath) {
+      return /\/app\//.test(filePath);
+    });
+
```
robertkowalski: we also have unit tests in `core`

how about blcklisting the dist folder?

benkeen: The duplicates were actually relative links in `dashboard/`, not dist; checking for `app/` (i.e. a folder in Fauxton) is the only appropriate way, I think.

robertkowalski:
+1

benkeen:
Merged as 1739a8e

PR #441, opened 2015-05-27
cors: don't throw on undefined origins

robertkowalski:


garrensmith:
This is fine. Otherwise show the Loadlines until everthing is fetched.

robertkowalski:
merged as d4e6997e9c228e5a90b33a55af873cc6b60896d2

PR #443, opened 2015-05-27
Grunt test task reordered

benkeen:
Small change to only run the JSX step after the clean and
dependencies steps. This ensures that any external addons being
included also get any .jsx tests compiled and ran. Up to now
any of those tests weren't being ran, when ran via this command.


robertkowalski:
merge after its green

benkeen:
Merged as cffe43a

PR #444, opened 2015-05-27
Clears the databases store after a test

benkeen:
This test can cause problems because the store isn't cleared out.
Explicitly resets the store.


michellephung:
+1

PR #448, opened 2015-05-28
Explicitly create DB in NW tests

benkeen:
This shouldn't be required, but we noticed that sometimes the
NW beforeEach() doesn't appear to create the database before the
test starts running.


benkeen:
Got the go-ahead from @robertkowalski to merge this when green. Merged.

PR #450, opened 2015-05-29
Loading lines now animate on FF

benkeen:
The new React loading lines weren't animating on FF. This just updates the CSS a bit so it works on non-webkit browsers.


michellephung:
+1 nice! thanks ben!

PR #451, opened 2015-06-01
Mango: reset Header Controller

robertkowalski:
Reset the Header-Controller (select-buttons) after creating a new
index and clicking on the link in the notification.


michellephung:
+1

PR #452, opened 2015-06-01
IndexResults: don't throw on deleting a second doc

robertkowalski:
Fix an issue where all result lists were throwing in case we
deleted one document and then tried to select another document
and delete it.


michellephung:
+!

PR #456, opened 2015-06-05
Workaround for React oddity

benkeen:
React does some odd stuff with the DOM like inserting `<noscript>`
tags to keep track of rendering info. In this instance it was
optionally inserting the noscript into a `<tr>`, which threw an error 
because it was invalid markup.

This is a workaround to detect the scenario until the issue is
resolved.


benkeen:
Nahh!

PR #458, opened 2015-06-08
Zen Mode option for small Ace Editor fields

benkeen:
This adds a zen-mode option to small Ace Editor fields to allow a lot more real estate for entering code. See screenshots:

Icon:
screen shot 2015-06-08 at 12 40 04 pm

Dark zen-mode (default):
screen shot 2015-06-08 at 12 40 26 pm

Light zen-mode: 
screen shot 2015-06-08 at 12 40 35 pm

The choice of theme (dark/light) is remembered in local storage.

### DIFF
--- a/app/addons/components/assets/less/code-editor.less
+++ b/app/addons/components/assets/less/code-editor.less
@@ -1,0 +1,81 @@
+.zen-editor-icon {
+  float: right;
+  font-size: 13px;
+  margin-top: 3px;
+}
+
+.full-page-editor-modal-wrapper {
+  position: fixed;
+  padding: 110px;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  margin: 0;
+  width: 100%;
+  z-index: 1025;
+
+  &.zen-theme-dark {
+    background-color: #323232;
+  }
+  &.zen-theme-light {
+    background-color: #ffffff;
+  }
+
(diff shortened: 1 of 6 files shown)


michellephung:
i think this is really neat! :grinning:

kxepal:
Why code highlight is different in dark/light modes?

benkeen:
@kxepal the "idle fingers" theme we use throughout Fauxton just doesn't work with light background (see example below). I started tweaking it, but then figured it was kind of pointless: we'll end up with two different themes anyway. So instead I just picked a theme already designed for a white background.

screen shot 2015-06-08 at 1 13 31 pm

michellephung:
+1

benkeen:
Merged as b1e6332

michellephung:
hey ben, was just thinking,

I think people kind of like being able to code on one side, and see the results on the other. Is there a way to make a medium editor, that takes up half the screen (a middle panel-sized editor) with the save button still there for fast previewing? Or maybe a re-sizeable editor?

benkeen:
Sorry @michellephung, I missed your comment. Yup, that's very do-able too. The little editors are a little odd in that their height defaults to the number of lines in the code, but only updates the size when you blur off the panel. Even changing that to the way it used to work - having it automatically resize as you enter text - would be an improvement.

PR #459, opened 2015-06-10
Refactoring <CodeEditor />

benkeen:
This PR refactors the `<CodeEditor />` component to make it
more generic. The existing React Ace editor instances now 
use this via wrapper components (`<CodeEditorPanel />` and 
`<ZenModeOverlay />`). The goal is to make this the one any 
only place where the ace editor is actually instantiated. It'll 
also be needed for the next step: refactoring the full page 
doc editor.

Other change:
- the small editors on the Mango + add/edit index pages now
dynamically resize up to 10 lines according to the content
that's been entered.


app/addons/components/react-components.react.jsx
```diff
@@ -220,56 +332,49 @@ function (app, FauxtonAPI, React, Components, ace, beautifyHelper) {
...
+    // 1. rename to hasErrors()
     hadValidCode: function () {
       var errors = this.getAnnotations();
-      // By default CouchDB view functions don't pass lint
       return _.every(errors, function (error) {
         return this.isIgnorableError(error.raw);
       }, this);
     },
+    // 2. This doesn't make sense. Could just call hasChanged() directly. Replace with clearChanges()
+    editSaved: function () {
```
garrensmith: This could be replaced with hasChanged(). I cannot remember why I kept this in.

garrensmith:
+1 this looks cool.

michellephung:
+1

PR #460, opened 2015-06-11
Gruntfile watch task performance improvements

benkeen:
As Fauxton's grown the watch tasks which execute during development
have been taking increasingly longer to run. The old code re-ran
various tasks for all file types in the application: JSX generation, copying 
remote files over to Fauxton, jshint and stylechecks. This is really slow. 
When working on a wrapper (like the Dash) it's become especially bad: 
often when you save a file and refresh the page, the various watcher 
tasks haven't finished executing so you get errors in the browser. Worse, 
often I've found the tasks can run simultaneously and interfere with 
eachother, with the result that you need to restart the server. Basically 
it's making development on Fauxton a real pain.

This updated version changes the behavior for the watch tasks for 
JS and JSX files. Now:
- JSX files found outside of the Fauxton folder are compiled
directly into the Fauxton folder.
- JSHint and stylecheck are ran on the changed files only.
- JSX module cache disabled. Since it's only ran during the initial
bootup and only takes a fraction of second, disabling it ensures
that on restarting the server everything is completely regenerated
afresh.

Unchanged:
- booting up Fauxton (`grunt dev`) behaves just as before and does
a clean slate pulling in the latest content. 
- the Less/CSS watcher tasks are unchanged. They aren't too slow 
and had some dependencies (@import) which would be sticky to 
get around.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -100,7 +100,7 @@ module.exports = function (grunt) {
 
   var templateSettings = function () {
     var defaultSettings = {
-     "development": {
+      "development": {
         "src": "assets/index.underscore",
         "dest": "dist/debug/index.html",
         "variables": {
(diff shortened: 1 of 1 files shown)


michellephung:
this is cool. I keep getting this error though: 
`Warning: Command failed: /bin/sh: jsxcs: command not found`,
+1 otherwise

benkeen:
That's really weird... it's just using the same dependencies as before. The only difference is that it's now running them directly via a shell command rather than via an npm command, but that shouldn't cause problems. Let's debug this tomorrow if you have time.

michellephung:
+1, all fixed.

benkeen:
Thanks, @michellephung! I'll ask @robertkowalski to take a glance at this when he gets back before merging, as well.

benkeen:
Welcome back, @robertkowalski. Now you're back, I can start bugging you to review stuff again. You're welcome!

Mind taking a glance at this? It runs *far* faster now. This ticket modifies the watcher task to only recompile changed js/jsx files instead of everything each time.

The one oddity you may notice is that when editing a JSX file, it appears to compile the file multiple times, but it actually doesn't: it recompiles all JSX files in the folder of the changes files (e.g. the tests). This is because the command-line utility for compiling the JSX only allows targeting a folder, not a single file: so any other JSX files (tests, for example) are compiled as well. Despite that, it still runs far faster.

michellephung:
@robertkowalski bump

robertkowalski:
wow this is exciting!

+1

robertkowalski:
good work ben, really nice stuff!

benkeen:
Thanks! Merged as 4799333

PR #462, opened 2015-06-16
Cloudant Query page link fix

benkeen:
Small link fix for the database name on the Cloudant Query page.
It was failing to link properly to the All Docs page.

There's no test because it works fine for test environments: the
link fails when Fauxton or a wrapper is ran under a separate
location, like whatever.html#/database/mydb/_all_docs


michellephung:
+1

benkeen:
Merged as d072d9f

PR #464, opened 2015-06-17
Url fix

garrensmith:
THis is a fix for [COUCHDB-2717](https://issues.apache.org/jira/browse/COUCHDB-2717). It also sets the back button for all views and indexes to _all_docs instead of trying to keep a history.

### DIFF
--- a/app/addons/documents/base.js
+++ b/app/addons/documents/base.js
@@ -75,7 +75,7 @@ function (app, FauxtonAPI, Documents) {
     },
 
     app: function (database, doc) {
-      return '/database/' + database + '/' + doc;
+      return '/database/' + database + '/' + encodeURI(doc);
     },
 
     apiurl: function (database, doc) {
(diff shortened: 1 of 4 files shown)


benkeen:
Hey @garrensmith, I think the biggest usability problem with this fix is when the user is inside a View, clicks into a doc, then `<` back. Having them return to the _all_docs page is kinda crumby - and I vaguely remember that was the how it used to function a long time ago, which people complained about.

Since this is kind of a fringe case, perhaps an interim solution would be to add in a crude check in the helper method for the `\` char, then redirect back to the _all_docs for that only. Obviously pretty crap, but it'd buy us some time until we get the React Router up and running.

garrensmith:
@benkeen can you take another look. I've changed the behaviour. It now uses the history when a user is viewing a doc. But uses `_all_docs` when returning from a view. I think this is a pretty nice solution.

benkeen:
Nice. Yeah, this works as I've expect. +1

garrensmith:
merged

PR #466, opened 2015-06-18
Database link fix on Add/Edit View page

benkeen:
This is the same issue as d072d9ffad7fc57c3cbb548c879950cd164cf1c6 just on a 
different page.

There's no test because it works fine for test environments: the
link fails when Fauxton or a wrapper is ran under a separate
location, like whatever.html#/database/mydb/_all_docs


michellephung:
+1

benkeen:
Merged as 9a445a0

PR #468, opened 2015-06-19
JS validation added for Add Database

benkeen:
Right now we rely on the server to validate the database name
that the user enters and return the error message if it's invalid.
That works in most cases, except when the db name contains a # or
? char. This PR just adds a simple one-liner regexp as suggested
in the CouchDB docs to validate it on the client first.

### DIFF
--- a/app/addons/databases/actions.js
+++ b/app/addons/databases/actions.js
@@ -65,9 +65,9 @@ function (app, FauxtonAPI, Stores, ActionTypes, Resources) {
     },
 
     createNewDatabase: function (databaseName) {
-      if (_.isNull(databaseName) || databaseName.trim().length === 0) {
+      if (_.isNull(databaseName) || databaseName.trim().length === 0 || !this.isValidDatabaseName(databaseName)) {
         FauxtonAPI.addNotification({
-          msg: 'Please enter a valid database name',
+          msg: 'Please enter a valid database name. The database must start with a letter and can only contain lowercase letters (a-z), digits (0-9) and the following characters _, $, (, ), +, -, and /.',
           type: 'error',
           clear: true
         });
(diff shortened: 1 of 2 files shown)


app/addons/databases/actions.js
```diff
@@ -65,9 +65,9 @@ function (app, FauxtonAPI, Stores, ActionTypes, Resources) {
     },
 
     createNewDatabase: function (databaseName) {
-      if (_.isNull(databaseName) || databaseName.trim().length === 0) {
+      if (_.isNull(databaseName) || databaseName.trim().length === 0 || !this.isValidDatabaseName(databaseName)) {
         FauxtonAPI.addNotification({
-          msg: 'Please enter a valid database name',
+          msg: 'Please enter a valid database name. Only lowercase letters (a-z), digits (0-9), and any of the characters _, $, (, ), +, -, and / are allowed. Moreover, the database name must begin with a letter.',
```
garrensmith: What about -

`Please enter a valid database name. The database must start with a letter and can only contain lowercase letters (a-z), digits (0-9) and the following characters _, $, (, ), +, -, and /.`

app/addons/databases/tests/actionsSpec.js
```diff
@@ -246,6 +246,40 @@ define([
...
+          "one%",
+          "one*",
+          "one@",
+          "one#",
+          "one^",
+          "one~",
+          "one two",
+          " one"
+        ].forEach(function (item) {
+          assert.equal(Actions.isValidDatabaseName(item), false, "item: " + item);
```
garrensmith: Nice tests.

garrensmith:
+1

app/addons/databases/actions.js
benkeen: Yeah that's much better, thanks!

PR #469, opened 2015-06-22
Move sidebar click handler to section header

benkeen:
The toggle event was placed one node too high on the Index
section in the sidebar, which resulted in the section being
expanded/contracted when a subelement is clicked, rather than
just the parent.


michellephung:
+1

benkeen:
Thanks Michelle. Merged as 52753f1

PR #473, opened 2015-06-23
Includes the second ace theme for bundling

benkeen:
This just updates the gruntfile to include the new Ace theme
used for the Zen editor. It wasn't working in bundled (prod)
environments because the file needed to be explictly included.


michellephung:
+1

PR #474, opened 2015-06-24
config: use the new _nodes endpoints

robertkowalski:
introduces a new screen where the user has to chose a node that
they want to edit.

I need help with the design:

bildschirmfoto 2015-06-24 um 12 02 39

bildschirmfoto 2015-06-24 um 12 02 45


janl:
nice!

Maybe post to dev@ about the design help request?

benkeen:
Brilliant @robertkowalski, been looking forward to this feature for months. Like you said earlier, even getting a functional draft merged in is a great step. UX can be improved later (don't judge me, UX people!).

Andreayurei:
How many nodes do we need to consider here? for example, could you start with one node and scale up to a thousand? If you had a thousand, would you have to page through it? Also, could we put CORS to configure page instead of making it separate?

Andreayurei:
@benkeen @seanbarclay says no the the sprinkle on UX design approach :-)

janl:
I totally understand that “make it pretty” is not a good approach, and I don’t think that is what’s being asked here.

It was quite hard to get the technical underpinnings right and to show them off @robertkowalski made the little demo of what it could look like. I don't think a design-first approach would have worked here, but I might be wrong.

From a design perspective, feel free to start from scratch! :)

As for the questions:

> How many nodes do we need to consider here? for example, could you start with one node and scale up to a thousand?

In theory, this could mean 100s and 1000s of nodes, but in practice, there are limits. The scenario for this is to get people to set up small-ish clusters, say 3-10, maybe 20 nodes and let them do this manually and guide them along.

If users get into the 20+ range of nodes, they will (very likely) use automated ways to set up their cluster nodes, and they could/would use the same system to roll out config changes across a cluster.

I don’t expect anyone updating a config value on 1000 nodes through this.

Now this poses an interesting question: CouchDB traditionally has not added any features that would break at scale. It’s not that Fauxton couldn’t handle 1000 nodes here, but it wouldn’t be practical for an end-user to use this, so should we have this feature or not?

I come down at the “it will be obvious when this doesn’t scale”-point, so I welcome the feature. And yes, it shouldn’t break when a users has 1000 nodes, but it is totally fine if it is impractical. We might even show a notice that says as much, when you go over, say 10 or 20 or 50 nodes.

> If you had a thousand, would you have to page through it?

I’d say this is up to the designer, I wouldn’t feel comfortable proposing a solution, just the technical constraints.

> Also, could we put CORS to configure page instead of making it separate?

I’m not sure what this refers to.

Andreayurei:
thanks for the feedback. we were just playing with @benkeen regarding UX design :-)

in terms of scaling nodes- that's good insight. should we add the ability to search and filter for nodes? assumption is yes...

in terms of paging- we would suggest infinite scrolling or filtering, unless there was a technical requirement that would prohibit this and require paging. e.g. the database list page requires paging due to technical constraints

on a selected node, on the secondary navigation, CORS is a separate section. we were wondering if it could be consolidated into the main config page- perhaps at the bottom.

janl:
> thanks for the feedback. we were just playing with @benkeen regarding UX design :-)

ah ;)

> in terms of scaling nodes- that's good insight. should we add the ability to search and filter for nodes? assumption is yes...

Can’t hurt, but for an MVP could be considered out of scope, if you want to keep the task small.

> in terms of paging- we would suggest infinite scrolling or filtering

I’m not a fan of the infinite scroll, *unless* done in a way that it is always obvious how navigation works, and using the back button will get me to my position in the list, not the top (like twitter does), but again, you are the ux experts :)

Thanks for explaining the CORS bit, I’ll leave that to @robertkowalski et. al. :)

wohali:
Great work.

Before this lands I'd like to see it tested and functional against 3 nodes not all running on localhost.

robertkowalski:
@wohali do you mean that you want better testing of the endpoint

```
_node/<fqdn>/_config
```

For Fauxton this endpoint is just like any other endpoint, like `/<db>/<doc>` - if you think we should test `_node/<fqdn>/_config` better that is not a task which is part of Fauxton as you can get the same problems using curl.

Short documentation of 

```
/_node/<fqdn>/_config
```

Allows you to reach the config of every node in a cluster from any node in the cluster.

wohali:
I simply meant that we need someone to *manually* run a test case where we validate that the configuration process works when setup against not just 3 localhost nodes - the way someone would actually deploy a 3 node cluster.

PR #475, opened 2015-06-25
Reactifying full page doc editor

benkeen:
This PR converts the full page document editor to React. Most of it
is just for the React conversion + dropping the Backbone code (includes
removing the old Backbone code editor), but there are also a few 
improvements:

- File uploads now work locally
- New docs hide the button row altogether (attachments, delete, clone) until saved.
- the string edit modal has been moved to be an optional subelements of `<CodeEditor />` enabled with a boolean (omited by default)
- various fixes to the string edit modal


garrensmith:
This is looking very good. One issue is I get this error when I view the full page:

`Warning: Failed propType: Required prop `doc` was not specified in `UploadModal`. Check the render method of `DocEditorController`.`

benkeen:
Thanks @garrensmith - I'll take a look.

benkeen:
N.B. this is pretty much done, but put on hold until I get time to return to it. Soon, I hope!

michellephung:
+1 everything working locally for me

app/addons/fauxton/components.react.jsx
```diff
@@ -269,12 +269,63 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
   });
 
 
+  // A super-simple replacement for window.confirm()
+  var ConfirmationModal = React.createClass({
+    propTypes: {
+      visible: React.PropTypes.bool.isRequired,
+      text: React.PropTypes.string.isRequired,
+      onClose: React.PropTypes.func.isRequired,
+      onSubmit: React.PropTypes.func.isRequired
```
robertkowalski: is there a reason to prefer callbacks instead of managing the state with a store?

app/addons/documents/doc-editor/components.react.jsx
```diff
@@ -0,0 +1,540 @@
...
+      var code = JSON.stringify(this.state.doc.attributes, null, '  ');
+      var editorCommands = [{
+        name: 'save',
+        bindKey: { win: 'Ctrl-S', mac: 'Ctrl-S' },
+        exec: this.saveDoc
+      }];
+
+      return (
+        <GeneralComponents.CodeEditor
+          id="docEditor"
```
robertkowalski: i think we are using no camelcase for classes and ids

app/addons/documents/index-editor/components.react.jsx
```diff
@@ -310,20 +308,18 @@ function (app, FauxtonAPI, React, Stores, Actions, Components, ReactComponents)
...
-        if (editorName === 'reduceEditor' && !indexEditorStore.hasCustomReduce()) {
-          return true;
-        }
-        var editor = this.refs[editorName].getEditor();
-        return editor.hadValidCode();
-      }, this);
+    hasErrors: function () {
+      var mapEditorErrors = this.refs.mapEditor.getEditor().hasErrors();
+      var customReduceErrors = (indexEditorStore.hasCustomReduce()) ? this.refs.reduceEditor.getEditor().hasErrors() : false;
+      console.log("errors? ", mapEditorErrors, customReduceErrors);
```
robertkowalski: debug output?

app/addons/documents/doc-editor/components.react.jsx
```diff
@@ -0,0 +1,540 @@
...
+      return {
+        uuid: null
+      };
+    },
+
+    cloneDoc: function () {
+      if (this.props.onSubmit) {
+        this.props.onSubmit();
+      }
+      Actions.cloneDoc(this.refs.newDocId.getDOMNode().value);
```
robertkowalski: you probably want to use a controlled component instead: https://facebook.github.io/react/docs/forms.html

app/addons/documents/doc-editor/components.react.jsx
```diff
@@ -0,0 +1,540 @@
...
+      }
+
+      this.setState({
+        inProgress: true,
+        errorMessage: ''
+      });
+
+      Actions.uploadAttachment({
+        url: this.props.doc.url(),
+        $form: $(this.refs.uploadForm.getDOMNode()),
```
robertkowalski: ???

app/addons/documents/doc-editor/components.react.jsx
```diff
@@ -0,0 +1,540 @@
...
+        </div>
+      );
+    },
+
+    uploadComplete: function () {
+      Actions.initDocEditor({
+        doc: this.state.doc,
+        onLoaded: function () {
+          this.getEditor().setValue(JSON.stringify(this.state.doc.attributes, null, '  '));
+          this.refs.uploadModal.reset();
```
robertkowalski: i think you can handle it better with stores and actions

app/addons/documents/doc-editor/components.react.jsx
```diff
@@ -0,0 +1,540 @@
...
+          editorCommands={editorCommands}
+          notifyUnsavedChanges={true}
+          stringEditModalEnabled={true}
+          change={this.onChangeDoc} />
+      );
+    },
+
+    onChangeDoc: function (doc) {
+      // super ugly, but necessary. This tells the user they can't delete the _id or _rev fields as they type and actually
+      // prevents it from being removed in the editable doc
```
robertkowalski: is there a reason why we keep this?

sometimes i want to create a new document by deleting the id and rev

app/addons/components/tests/stringEditModalSpec.react.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+  var assert = utils.assert;
+  var TestUtils = React.addons.TestUtils;
+
+  describe('String Edit Modal', function () {
+    var container, modalEl, sandbox;
+    var stub = function () { };
+
+    beforeEach(function () {
+      sandbox = sinon.sandbox.create();
```
robertkowalski: sandbox can get deleted

app/addons/components/react-components.react.jsx
```diff
@@ -385,18 +495,145 @@ function (app, FauxtonAPI, React, Components, ace, beautifyHelper) {
...
+          <div ref="ace" className="js-editor" id={this.props.id}></div>
+          <button ref="stringEditIcon" className="btn string-edit" title="Edit string" disabled={!this.state.stringEditIconVisible}
+            style={this.state.stringEditIconStyle} onClick={this.openStringEditModal}>
+            <i className="icon icon-edit"></i>
+          </button>
+          <StringEditModal
+            ref="stringEditModal"
+            visible={this.state.stringEditModalVisible}
+            onSave={this.saveStringEditModal}
+            onClose={this.closeStringEditModal} />
```
robertkowalski: this callback means that the subcomponent directly alters the state of the parent component, i think in react we have to use actions and stores to transfer state to components.

robertkowalski:
wow, massive code amount. the only thing i am unhappy about is the use of callbacks which then alter state of components from actions or other components, e.g. because references to `this.setState` are passed from the caller to the callee

app/addons/fauxton/components.react.jsx
benkeen: In this case, absolutely. We can get a lot of utility out a generic confirmation modal like this. Tying it to a specific store would prevent being used elsewhere.

app/addons/documents/doc-editor/components.react.jsx
benkeen: Oops, good point.

app/addons/documents/doc-editor/components.react.jsx
benkeen: Agreed, I don't much like this.

app/addons/documents/doc-editor/components.react.jsx
benkeen: Legacy! Good catch. The old code used $form + a plugin to handle the request generation. Removed.

app/addons/documents/doc-editor/components.react.jsx
benkeen: Very good, I didn't think of that. I think the parent component (which passes the onSuccess method to the upload component) could listen to a published "done" msg and do whatever it needs there. Much cleaner.

benkeen:
Thanks @robertkowalski! Great feedback, I agree about the callbacks.

app/addons/components/react-components.react.jsx
benkeen: Hmm, I'm on the fence on this one. In many cases (like one or two instances you mentioned above) I agree, but the nice thing about callbacks is that dumb components like StringEditModal don't need to maintain their own state and can be used anyway very, very easily (just pass in props). I think it's simpler, less code, and easier to re-use the component. 

I don't think passing callbacks in of itself is "un-Reacty" or necessarily a bad approach.

app/addons/documents/doc-editor/components.react.jsx
benkeen: Urgh, just been looking at this. Can we chat about it tomorrow? Frankly there's downsides to both approaches and I'm not sure a controlled component is an improvement.

robertkowalski:
i see a lot of merge conflicts coming, maybe it is possible to slice this pr into smaller ones (working from the modals upwards)

robertkowalski:
it also makes it easier to get it merged over time given the current workload for other tasks

robertkowalski:
ping me any time, just read your comment btw

benkeen:
Thanks @robertkowalski! Actually I think this is now good to merge. This ticket has gone of for so long and is now ready, I'd strongly favour just merging. I was waiting on fixing things that would be broken upstream - but they're now ready as well.

Beyond the stuff to talk about that I commented on, I agree with your idea to remove the section that prevents the user creating a new doc by changing the ID. But rather than prolonging this even further, I say we tackle it later. Let's talk!

app/addons/documents/doc-editor/components.react.jsx
benkeen: I agree, be nice to remove. But I played around with it + it'll require a little more effort to check for the values + do a create request for that scenario. Very doable, but I'm relegating it to the Out Of Scope pile for now. :)

benkeen:
Hey Robert, I’ve updated a few things. I started working through the whole scenario of moving the callbacks for the ConfirmationModal into a store, but abandoned it once I saw how much more complicated + klutzy it was going to be. :( 

Let me explain. Say we remove the onClose and onSubmit props and get the confirmation modal to posts its own generic “submit” & “close” events as we discussed. The confirmation modal’s store would then listen to those events and set some corresponding values internally (`_isSubmitted`, `_isClosed`, say). The component that’s using `<ConfirmationModal />` would then have something like an `isSubmitted: store.isConfModalSubmitted()` prop in its own state. Then, in componentWillUpdate (I guess?) It would have detect for the change, then actually reset the “isSubmitted” value in the store immediately afterwards, like so:

```javascript
componentWillUpdate: function (nextProps, nextState) { 
	if (!this.state.isSubmitted && nextState.isSubmitted) {
		ComponentAction.doSomething();
		FauxtonAction.resetConfModalSubmitted(); // this would ultimately reset _isSubmitted in the 
	}
}
```

Is there a better approach, do you think? I guess we could remove the `FauxtonAction.resetConfModalSubmitted()` call by having a “_numSubmitted” counter in the store and use that to determine when in the parent component to actually submit - but this all feels tremendously hacky to me compared to the original approach of just passing in callbacks via props.

app/addons/fauxton/components.react.jsx
benkeen: As per our last conversation, I think this is okay. The parent component just fires an action in the onClose and onSubmit methods and doesn't modify itself directly.

benkeen:
Thanks everyone. :) Merged as 8cd744a

michellephung:
yay for getting this in! :D

PR #480, opened 2015-07-15
Just enable config for a "cluster-of-one"

robertkowalski:
Warn users if they run multi node setups.

Testing instructions for single node setups, given you usually run
a cluster of three nodes:

1. stop couchdb, connect fauxton to a single node, change

```
exports.couch = 'http://localhost:15984/';
```

in `tasks/helper.js`

2. get rid of `dev/lib/` in the couchdb folder, so the cluster does
not know about the other nodes any more, caution: you will lose
your current dbs.

3. boot couchdb with one node:

```
./dev/run --with-admin-party-please -n1
```
4. PROFIT!


garrensmith:
+1 looks good. Nice work.

robertkowalski:
merged!

PR #481, opened 2015-07-15
add stop svg for warehousing

Andreayurei:
@benkeen here's the stop icon


benkeen:
I think this can be closed.

PR #482, opened 2015-07-20
Trigger error on failed session

garrensmith:
If a session fetch fails then fire an event


app/core/couchdbSession.js
```diff
@@ -58,6 +58,8 @@ function (FauxtonAPI) {
           // this will return the user as a value to all function that calls done on this
           // eg. session.fetchUser().done(user) { .. do something with user ..}
           return user;
+        }, function (session, xhr, type, message) {
+          that.trigger('session:error', xhr, type, message);
         });
```
robertkowalski: we use bind

```js
function (session, xhr, type, message) {
  this.trigger('session:error', xhr, type, message);
}.bind(this));
```

robertkowalski:
+1 can you kick off travis again?

i will try to fix the flaky test in the meantime

robertkowalski:
hmm no idea how to fix it: in this travis run the test for replication really failed. the validation test seems to be broken by design as it just sets up a replication in a fire and forget style. this means the validation can happen and the replication is not finished.

robertkowalski:
idea: before we kick of [this test](https://github.com/apache/couchdb-fauxton/blob/master/app/addons/verifyinstall/resources.js#L138-158) in the "verify your couchdb installation" we poll if the replication is finished

garrensmith:
Merged.

PR #485, opened 2015-07-24
Add callback for when Tray closed by clicking off it

benkeen:
The Tray component has custom callbacks for explicit hide/show
methods, but not when the Tray is closed by clicking outside of
it. This is useful when using a Tray within another component
and the visiblity status is stored in a store. We need to track
the auto close event to update the store.


michellephung:
+1

benkeen:
Thanks @michellephung!

benkeen:
Merged as 2d8cabf

PR #488, opened 2015-08-03
Use CSRF protection if available

rnewson:
Use CSRF protection if available

COUCHDB-2762

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -63,6 +63,33 @@ function (app, $, _, Backbone, Bootstrap, Helpers, Utils, FauxtonAPI, Couchdb) {
   // Localize or create a new JavaScript Template object
   var JST = window.JST = window.JST || {};
 
+  var parseCookies = function (cookies) {
+    if (!cookies) {
+      return {};
+    }
+    return _.reduce(cookies.split(';'), function (list, cookie) {
+      var parts = cookie.split('=');
+      list[parts.shift().trim()] = decodeURI(parts.join('='));
+      return list;
+    }, {});
+  };
+
+  $._ajax = $.ajax;
+
+  $.ajax = function (settings) {
+    var cookies = parseCookies(document.cookie);
+    var csrf = cookies['CouchDB-CSRF'] ? cookies['CouchDB-CSRF'] : 'true';
+    var origBeforeSend = settings.beforeSend;
+    var newBeforeSend = function (xhr) {
+      if (origBeforeSend) {
+        origBeforeSend(xhr);
+      }
+      xhr.setRequestHeader('X-CouchDB-CSRF', csrf);
+    };
+    settings.beforeSend = newBeforeSend;
+    return $._ajax(settings);
+  };
+
   // Configure LayoutManager with Backbone Boilerplate defaults
   FauxtonAPI.Layout.configure({
     // Allow LayoutManager to augment Backbone.View.prototype.


rnewson:
Just for demonstration purposes, I expect there's better ways to do this.

app/app.js
```diff
@@ -63,6 +63,29 @@ function (app, $, _, Backbone, Bootstrap, Helpers, Utils, FauxtonAPI, Couchdb) {
   // Localize or create a new JavaScript Template object
   var JST = window.JST = window.JST || {};
 
+  var parseCookies = function (cookies) {
```
garrensmith: You can simplify this function by using [`_.reduce`](http://underscorejs.org/#reduce).

app/app.js
```diff
@@ -63,6 +63,29 @@ function (app, $, _, Backbone, Bootstrap, Helpers, Utils, FauxtonAPI, Couchdb) {
...
+    var list = {}, rc = cookies;
```
garrensmith: Can you put each argument declaration on a new line.

app/app.js
```diff
@@ -63,6 +63,29 @@ function (app, $, _, Backbone, Bootstrap, Helpers, Utils, FauxtonAPI, Couchdb) {
...
+  };
+
+  $._ajax = $.ajax;
+
+  $.ajax = function (settings) {
+    var cookies = parseCookies(document.cookie);
+    var csrf = cookies['CouchDB-CSRF'] ? cookies['CouchDB-CSRF'] : 'true';
+    var origBeforeSend = settings.beforeSend;
+    var newBeforeSend = function (xhr) {
+      if (origBeforeSend) origBeforeSend(xhr);
```
garrensmith: Can you add `{` and `}` for the if statement.

app/app.js
```diff
@@ -63,6 +63,29 @@ function (app, $, _, Backbone, Bootstrap, Helpers, Utils, FauxtonAPI, Couchdb) {
   // Localize or create a new JavaScript Template object
   var JST = window.JST = window.JST || {};
 
+  var parseCookies = function (cookies) {
+    var list = {}, rc = cookies;
+    rc && rc.split(';').forEach(function (cookie) {
```
garrensmith: Can you rather change the `rc` check to an if statement. Its better to make it a bit more explicit.

app/app.js
```diff
@@ -63,6 +63,29 @@ function (app, $, _, Backbone, Bootstrap, Helpers, Utils, FauxtonAPI, Couchdb) {
...
+      var parts = cookie.split('=');
+      list[parts.shift().trim()] = decodeURI(parts.join('='));
+    });
+    return list;
+  };
+
+  $._ajax = $.ajax;
+
+  $.ajax = function (settings) {
```
garrensmith: Instead of overriding `$.ajax` could we rather use a [global setting](https://api.jquery.com/jQuery.ajaxSetup/)?

rnewson: we'll need to derive the right value from the current cookie (or absence of), it's not a constant.

rnewson:
@garren are we good to go now?

garrensmith:
I don't love the fact we overriding ajax. But I cannot think of a better way. So +1

rnewson:
An iteration over this would be for all our currently $.ajax calls to call a function of our own making. I did it this way to be sure I got them all.

PR #494, opened 2015-08-07
Fauxton csrf fixes

robertkowalski:
Main patch is:

```
   csrf: fix tests

    when fauxton requests session token & csrf token it handles them
    as additional data it sends to each ajax request.

    the testsuite tries to open 5984:/_alldocs and friends in several
    places right after running the test. as the csrf cookie is sent
    via ajax, a plain GET will result in a CSRF error
```

Other patches:

- add test coverage
- don't monkey patch ajax, use an event handler "hook"


rnewson:
lgtm

garrensmith:
+1 nice fix.

robertkowalski:
merged :)

PR #497, opened 2015-08-12
Add CSRF indicator

robertkowalski:
includes a lot of refactoring and bug fixes, click on each commit for less noise in the diff :)

main patch: 

```
adds a small indicator to the sidebar if we are protected
against CSRF.

to test, comment `res.setHeader('x-couchdb-csrf-valid', 'true');`
in `tasks/couchserver.js` and browse without logging into fauxton

we have to modify the dev-server to test as the dev-version of
fauxton fetches the html templates through it with ajax, which is
disturbing for the detection.

this closes COUCHDB-2769
```


garrensmith:
+1 once the tests pass.

robertkowalski:
merged

PR #499, opened 2015-08-19
[DISCUSSION] enable include_docs per default

robertkowalski:
- shows document content
- do not override the amount of displayed docs and set it to 100
  when clicking back

### DIFF
--- a/app/addons/databases/components.react.jsx
+++ b/app/addons/databases/components.react.jsx
@@ -131,7 +131,7 @@ define([
       return (
         <tr>
           <td>
-            <a href={"#/database/"+encoded+"/_all_docs"}>{name}</a>
+            <a href={"#/database/" + encoded + "/_all_docs?include_docs=true"}>{name}</a>
           </td>
           <td>{size}</td>
           <td>{row.status.numDocs()} {this.renderGraveyard(row)}</td>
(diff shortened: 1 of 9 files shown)


kxepal:
That was so long time ago initially. The problem is that indexes pages view turns into nightmare when you deal with big documents since you cannot disable this option before actual request is made. 

Imagine 4MiB documents in db. 100 per page (default) -> 400MiB of received JSON which need to parse and render. Default options shouldn't hurt users in any situations.

michellephung:
:/  maybe test for how big a document is before showing it?

kxepal:
@michellephung Good idea, and it's possible with additional HEAD request per document. 100 rows per page - 100 additional requests, but do we really need that?

michellephung:
no that wouldn't be good

elistevens:
I would like to have it in the user's hands, off by default.  Sometimes I'm going to want to see 400mb of JSON, but it will be rare.

willholley:
how about off by default but remembering the setting between visits to the page (could be per-database / view)? The large documents problem happens regardless of when the user uses include_docs so it sounds like a separate ticket / fix.

robertkowalski:
We are planning table views, see also the data importer that is about to land

garrensmith:
This will be fixed when we switch to a more default table view.

PR #506, opened 2015-08-28
Remove reference to $.clearForm

benkeen:
There's an error that appears after uploading a file. This is due
to the removal of a jQuery form lib. This PR removes the reference
to the unfound function and clears the form manually.


robertkowalski:
to test this you can just add something like

```js
  .assert.attributeEquals('#_rev', 'value', '');
```

to https://github.com/benkeen/couchdb-fauxton/blob/upload-file-fix/app/addons/documents/tests/nightwatch/uploadAttachment.js#L30

benkeen:
Good call. Will do.

benkeen:
Thanks @robertkowalski - updated.

robertkowalski:
+1

benkeen:
Thanks Robert. Merged as 0c2645a

PR #507, opened 2015-08-28
Updating Fauxton layout to use flexbox - #1

benkeen:
This is a first salvo for cleaning up the layout CSS for Fauxton.
Right now it’s kind of a mess: lots of floats, position
absolutes/fixed’s and more. This PR focuses on updating the
with_tabs_sidebar.html template so the main panel uses flexbox:
meaning the optional tab row, body content and optional footer.

All Fauxton pages should look identical to before, except it
fixes a small issue where the scrollbar for the main panel
on the Changes page (and other pages) kind of scrolls off the
top.

Benefits to using flexbox are:
- we can drop a LOT of complicated CSS once it’s all ported over
- flexbox is much easier to understand than older methods
- we can drop the position:fixed + z-indexes that cause issues
with embedding modals anywhere in the DOM (and have the
backdrop appear properly).

Affected pages:
- all pages in the Databases section
- Config page (single node)

Tested on Chrome 44, FF 40, Safari 7.05, IE 10


robertkowalski:
neat! 

also great docs in form of the commit message!

big +1

benkeen:
Thanks, Robert! Nice to get this one in. Merged as 5f52818

PR #509, opened 2015-08-31
fix login/logout with IE

robertkowalski:
disable caching for IE which prevents IE from logging in/out.

as IE starts to add cachebust-get-parameters when we disable ajax-
caching, i had to adjust the dev-server.

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -25,6 +25,11 @@ function (app, FauxtonAPI, LoadAddons) {
   // root folder to '/' by default.  Change in app.js.
   Backbone.history.start({ pushState: false, root: app.root });
 
+  // feature detect IE
+  if ('ActiveXObject' in window) {
+    $.ajaxSetup({ cache: false });
+  }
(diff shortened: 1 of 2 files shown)


benkeen:
This is pretty broad - disabling all request caching for the browser seems a little extreme. Any chance it could be made for the individual login/logout requests? I've ran into issues with IE's caching as well, but found it could be solved on a per-request level with a `whatever.fetch({ cache: false })`.

robertkowalski:
@benkeen 

when we bundle the application for production, we create on big js file that is loaded once.

disabling caching for IE in this case means that we disable it just for all API calls to couch - which is the wanted behaviour as we don't want to show outdated data on databases, too

benkeen:
This seems reasonable, thanks @robertkowalski! +1

robertkowalski:
meregd as 01650f62cacbb7110b35d368080a44c9d7d93fda

PR #510, opened 2015-08-31
Increase timeout for NW tests

benkeen:
The max timeout time for several of the NW custom commands was
set to 10 seconds. Recently NW tests have been failing a good deal
more (esp on the dash) and all appear to be due to timeouts.

This PR changes it to use the default timeout time (30 seconds). I
think this is wise in any event, since ANY effort to increase the
robustness of the tests is needed. Plus it won't slow anything
down - just make it more likely to identify legitimately failing tests.


robertkowalski:
+1

benkeen:
Merged as 8219b17. Thanks, Robert.

PR #511, opened 2015-09-01
remove broken notifications

robertkowalski:
they don't notificate any more, and they would flood couchdb-dev
if they would work.


garrensmith:
+1

robertkowalski:
merged as be90e4ccd15b9601a3a2440fd962a8fb5561b860

PR #512, opened 2015-09-01
Minor extension check

benkeen:
Small thing. If the 'mango:additionalIndexes' extension isn't
defined it throws a JS error and causes the page not to load.
This just adds a check.

### DIFF
--- a/app/addons/documents/mango/mango.helper.js
+++ b/app/addons/documents/mango/mango.helper.js
@@ -29,8 +29,11 @@ define([
     }, []);
 
     if (!nameArray.length) {
-      indexes = FauxtonAPI.getExtensions('mango:additionalIndexes')[0];
-      nameArray = indexes.createHeader(doc);
+      indexes = FauxtonAPI.getExtensions('mango:additionalIndexes');
+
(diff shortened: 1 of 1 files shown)


robertkowalski:
looks good but a unit test is missing

benkeen:
Oh rats.

robertkowalski:
i think we can close this now. imo it is good that it crashes if the extension is missing and not failing silently

benkeen:
I don't know... seems like extensions are by design *optional*, so crashing when one isn't defined seems kind of inelegant. Plus that way looking at the codebase I'd never know which extensions needed to be defined or not in order for a chunk of code to run. I think we should supply an alternate test to confirm that it's working as expected (actually, I think one may already exist).

robertkowalski:
fauxton has no optional extensions defined, and does not crash and that's cool.

right now it errors if you try to use extensions, that are registered in a wrong way. these extensions are partly registered and not set up properly. 

crashing here is better compared to failing silently which will introduce cascading errors or weird behavior where the extension should be used. it also makes debugging very hard.

benkeen:
K. Closin'.

PR #513, opened 2015-09-01
use a stricter way to determine if db exists

robertkowalski:
avoids false positives like fauxton-seleniumtest2

code is covered by the testsuite itself :)


benkeen:
+1

PR #516, opened 2015-09-08
increase timeout for notifications

robertkowalski:
We might redefine the way notifications work in the future, but it
might be also vaporware, so I'm fixing the timeout for the near
future.

This closes COUCHDB-2692


garrensmith:
+1

robertkowalski:
merged

PR #519, opened 2015-09-09
Add generation date to index.html

benkeen:
Small thing: this just adds a UTC date for the time the index.html
file was last generated. Useful so we have an easy way to figure
out when the code was last built.

Looks like: `<!-- Generated: 2015-09-09T15:36:15.417Z -->`


garrensmith:
+1

PR #520, opened 2015-09-10
use FauxtonAPI.url helper

robertkowalski:
covered by the attachment selenium test

### DIFF
--- a/app/addons/documents/doc-editor/actions.js
+++ b/app/addons/documents/doc-editor/actions.js
@@ -141,9 +141,13 @@ function (app, FauxtonAPI, ActionTypes) {
     // store the xhr in parent scope to allow us to cancel any uploads if the user closes the modal
     xhr = $.ajaxSettings.xhr();
 
+    var query = '?rev=' + params.rev;
+    var db = params.doc.getDatabase().safeID();
+    var docId = params.doc.safeID();
     var file = params.files[0];
+
     $.ajax({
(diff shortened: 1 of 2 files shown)


app/addons/documents/doc-editor/actions.js
```diff
@@ -141,9 +141,13 @@ function (app, FauxtonAPI, ActionTypes) {
     // store the xhr in parent scope to allow us to cancel any uploads if the user closes the modal
     xhr = $.ajaxSettings.xhr();
 
+    var query = '?rev=' + params.rev;
```
michellephung: tthis name (query) threw me for a loop initially. Maybe name it revision?

michellephung:
+1 code is working

app/addons/documents/doc-editor/actions.js
robertkowalski: it is a query string :)

https://en.wikipedia.org/wiki/Query_string

michellephung: it is a query string, but for attachments it will always be the document revision right? either way still +1

robertkowalski:
merged as 24953b29308682a9832870a9f33a1ad983ff5ca1 - thank you

app/addons/documents/doc-editor/actions.js
robertkowalski: right that's true

PR #521, opened 2015-09-11
Allow custom buttons to be added to the Doc Editor page

benkeen:
This feature was accidentally removed in the recent full page doc
editor refactor. This ticket adds the option back in to allow
other code to add arbitrary buttons on the doc editor page.


michellephung:
+1

PR #523, opened 2015-09-14
Specifying correct params for before/afterEach

benkeen:
The NW docs say the done() callback is the second param, not first.

### DIFF
--- a/test/nightwatch_tests/helpers/helpers.js
+++ b/test/nightwatch_tests/helpers/helpers.js
@@ -20,14 +20,16 @@ module.exports = {
   getNanoInstance: function () {
     return nano(this.test_settings.db_url);
   },
-  beforeEach: function (done) {
+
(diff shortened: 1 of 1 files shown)


robertkowalski:
+1 if green

benkeen:
I've been battling trying to get couch going all day. It's possible I've just borked something that I don't understand, but it may also be couch. Some info:

- Ysing the latest couch code for me prevents Fauxton working at all. All requests just hang. 
- reverting to an earlier couch SHA (eb34513 - the last sha that was working for me) gets Fauxton running again. However, the load balancer seems wonky: creating a new DB request only adds it to a single node, not all nodes. Thus NW tests fail erratically - in a very similar fashion to what's happening on this ticket. Could be coincidence.

benkeen:
Ignore above, it was a couch issue. Re-running CI.

PR #525, opened 2015-09-16
Better handle runtime-loaded files for bundled envs

benkeen:
Occasionally we encounter issues where a file needs to be loaded
at runtime, e.g. ZeroClipboard.swf, Ace editor. To handle this we
have some custom Grunt rules to copy them over to a /js subfolder
where their location is known. This PR removes the custom
code in grunt for Zeroclipboard and make it generic so that
any addon can define a /dependencies subfolder. Anything placed
in that location will be moved to a /js/[addonName] subfolder
for production (bundled) builds.


benkeen:
@robertkowalski, @michellephung: could I get a critical look at this? It replaces all the custom zeroclipboard handling code in grunt, couchserver, etc. to make it copied over generically for prod environments. The existing NW tests confirm zeroclipboard continued to work, but not for a prod env.

michellephung:
+1 works for me

benkeen:
Merged as f0d4518617ac58fa394f7b3db92dd605d7253c98

PR #527, opened 2015-09-17
Passes the database info to doc editor buttons

benkeen:
This just saves extensions having to wait for the doc to load to get
the database name.


michellephung:
+1

PR #528, opened 2015-09-18
Update two_pane template to use flexbox

benkeen:
This updates the CSS for two_pane to remove the old
layout CSS and use flexbox.

Impact:
- Cloudant Query page

### DIFF
--- a/app/addons/documents/mango/mango.components.react.jsx
+++ b/app/addons/documents/mango/mango.components.react.jsx
@@ -129,7 +129,7 @@ function (app, FauxtonAPI, React, Stores, Actions,
       var url = '#/' + FauxtonAPI.urls('allDocs', 'app', this.props.dbName, '');
 
       return (
-        <div className="editor-wrapper span5 scrollable">
+        <div className="editor-wrapper">
           <PaddedBorderedBox>
             <div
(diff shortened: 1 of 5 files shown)


app/addons/documents/mango/mango.components.react.jsx
```diff
@@ -129,7 +129,7 @@ function (app, FauxtonAPI, React, Stores, Actions,
       var url = '#/' + FauxtonAPI.urls('allDocs', 'app', this.props.dbName, '');
 
       return (
-        <div className="editor-wrapper span5 scrollable">
+        <div className="editor-wrapper">
```
robertkowalski: i guess this change has also to be made to the index-editor?

robertkowalski: aka view-editor

benkeen: Actually not yet. The index-editor doesn't use two_pane yet, so the flexbox refactor will change that in a later commit.

michellephung:
+1

benkeen:
Thanks @michellephung. Just rebased with master. Will merge shortly.

PR #529, opened 2015-09-18
add a wizard for setting up a cluster

robertkowalski:
how to test:

set the port in `exports.couch` to a node that will handle the
setup (the "setup-node"):

```
exports.couch = 'http://localhost:15984/';
```

if you change the port of the setup node during setup the wizard will
lose the connection and can't finish.

(this will change in a proper release where fauxton is running on each node)

how to run couch:

```
rm -rf dev/lib/ && rm -rf dev/logs/ && /dev/run --with-admin-party-please --no-join
```

deletes your dbs, the `--no-join` spawns three nodes that are not joined into a cluster

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ app/addons/*
 !app/addons/documents
 !app/addons/styletests
 !app/addons/cors
+!app/addons/setup
 settings.json*
 i18n.json
 !settings.json.default
--- a/app/addons/components/react-components.react.jsx
+++ b/app/addons/components/react-components.react.jsx
@@ -907,7 +907,7 @@ function (app, FauxtonAPI, React, Components, ace, beautifyHelper) {
   var ConfirmButton = React.createClass({
     render: function () {
       return (
-        <button type="submit" className="btn btn-success save" id={this.props.id}>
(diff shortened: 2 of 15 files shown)


robertkowalski:
note: you need to run configure & make before testing, as it requires the latest version of https://github.com/apache/couchdb-setup

robertkowalski:
scenarios:

```
rm -rf dev/lib/ && rm -rf dev/logs/ && /dev/run --with-admin-party-please --no-join -n1
rm -rf dev/lib/ && rm -rf dev/logs/ && /dev/run --admin=tester:testerpass --no-join -n1

rm -rf dev/lib/ && rm -rf dev/logs/ && /dev/run --with-admin-party-please --no-join
rm -rf dev/lib/ && rm -rf dev/logs/ && /dev/run --admin=tester:testerpass --no-join
```

michellephung:
this is working ok.

a little confusing what ports are acceptable, but not a blocker
+1 because it is working. 

one small thing: 

screen shot 2015-10-29 at 7 00 48 pm

I think this is supposed to say something other than 'undefined'.

robertkowalski:
> I think this is supposed to say something other than 'undefined'.

it goes away if you restart grunt after switching the branch, because the build system has to bake the string into the dashboard.

robertkowalski:
merged!

PR #532, opened 2015-09-21
Convert headers to use flexbox

benkeen:
This removes the old .fixed-header class on the header/div
element in the templates and cleans up the CSS a bit to use
flexbox to layout the subelements.

N.B. In fixing this I also tweaked the active tasks header
to reduce the height a little (it was affected at one point)

### DIFF
--- a/app/addons/activetasks/assets/less/activetasks.less
+++ b/app/addons/activetasks/assets/less/activetasks.less
@@ -90,7 +90,7 @@
   }
 
   #active-tasks-filter-tabs {
-    height: 100px;
+    height: 60px;
     background-color: #CBCBCB;
     padding-left: 20px;
     overflow: hidden;
@@ -103,7 +103,7 @@
 
     li {
       background-color: #eee;
-      margin-top: 65px;
+      margin-top: 25px;
       margin-left: 3px;
 
       &.active-tasks-checked {
--- a/app/addons/documents/assets/less/doc-editor.less
+++ b/app/addons/documents/assets/less/doc-editor.less
@@ -20,6 +20,11 @@
     padding-bottom: 0;
     top: 0;
   }
+
+  /* temporary wedge: can be removed at end of flexbox refactor */
+  #api-navbar {
+    float: right;
+  }
 }
 
 #editor-container {
@@ -39,11 +44,6 @@
     }
   }
 
-  .fixed-header {
-    height: 65px;
-    .box-shadow(none);
-  }
-
   #breadcrumbs {
     white-space: nowrap;
   }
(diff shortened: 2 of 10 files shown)


michellephung:
I love this.
+1

Could you change the last few lines of the css file though?


```
.polling-interval-widget {
  width: 225px;
  margin: 10px 30px 0px 0px;
  color: #666;
  float:right;

  li {
    list-style-type: none;
  }

  .polling-interval-time-label {
    display: inline-block;
    float: right;
    margin-right: 0px;
    cursor: default;
  }

  #polling-range {
    width: 225px;
  }
}

.active-tasks-loading-lines-container {

  width: 355px;

  .active-tasks-loading-lines {
    border: 0px;
    position: relative;
    top: 17px;
    float: left;
    padding-left: 20px;

    #line1, #line2, #line3, #line4 {
      background-color: #bbb;
    }
  }
}
```

should look like this, when with the above changes: 
screen shot 2015-09-25 at 1 12 36 am

michellephung:
you should be able to just copy and paste my changes into the active tasks less file in your branch,
but i could also do it after, if you prefer, just let me know

benkeen:
Brilliant, thanks @michellephung! Will look at this shortly :+1:

benkeen:
@michellephung w.r.t to this change, I don't quite see what you see... I still see the Active Tasks title appear. See screenshot: 
screen shot 2015-09-25 at 9 35 51 am

The reason I touched on the active tasks CSS was just because the layout got pushed down a bit, so while I patched that I just reduced the height of the tab section a little. Seems in the ballpark of this ticket. :) But if we're removing the title altogether, I wonder if that might be a different ticket? Seems perhaps more consistent to keep it, no?

michellephung:
Hi Ben,

So originally, I put the dark grey tab-header part a little longer so that at smaller widths the polling will have a space to move into, ala:
screen shot 2015-09-25 at 12 41 41 pm

I'll admit it did look a little weird, when full screen, but I wasn't sure what to do. 
With this PR, at smaller widths, the header will look like this:
screen shot 2015-09-25 at 12 47 25 pm

So the polling part will run into the sidebar, a little too close.

The snippet I pasted above makes the polling widget a bit smaller, and gives a bit of padding so that the loading lines aren't on top of the sidebar:
screen shot 2015-09-25 at 1 12 36 am

Normal sized, everything looks pretty much the same. :D

I thought your intention was to move the polling widget over the title, which I think is a good idea!

benkeen:
Oh, gotcha! Yeah I actually quite like that - good call. :)

benkeen:
Merged as 99dfd3f

PR #536, opened 2015-09-24
Catch all errors when failing to load doc

benkeen:
This makes the error handling for loading a doc a little more
generic. This actually caused an issue on the dash, not with
Fauxton - but good to change so that all errors are caught when
failing to load a document and the user's redirected back
to the database all docs page.

### DIFF
--- a/app/addons/documents/doc-editor/actions.js
+++ b/app/addons/documents/doc-editor/actions.js
@@ -39,10 +39,8 @@ function (app, FauxtonAPI, ActionTypes) {
         params.onLoaded();
       }
     }, function (xhr, reason, msg) {
-      if (xhr.status === 404) {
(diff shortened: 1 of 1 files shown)


garrensmith:
+1

garrensmith:
Is it work adding a test for this? Specially if its caused an issue before?

benkeen:
Thanks @garrensmith! I was thinking of adding a test, but since it was just the fail condition of the then that would be confirmed to run, I figured all it would do is confirm jQuery then() works... so ditched the test.

I actually introduced this issue recently when I did the full page doc editor rewrite. At the time I thought adding in the 404 test was a little cleaner, but I didn't consider the other failing condition states and that we'd want a redirect with those too.

benkeen:
Merged as a589e59

PR #537, opened 2015-09-25
Fix for editing View not showing correct content

benkeen:
There's a bug where if you edit one index, then edit a second,
you still see the Map Function from the first view. This PR does an
explicit reset every time you edit a view.

### DIFF
--- a/app/addons/documents/index-editor/actions.js
+++ b/app/addons/documents/index-editor/actions.js
@@ -76,6 +76,10 @@ function (app, FauxtonAPI, Documents, ActionTypes, IndexResultsActions) {
       });
     },
 
+    clearIndex: function () {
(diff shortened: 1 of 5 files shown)


benkeen:
How to reproduce:

1. Create a new view and give it a custom Map Function (like `function(doc) { emit(doc.id, 123); }`)
2. save it, the go back to the  db's all_docs page
3. Create a second new view, with a different Map Function (like `function(doc) { emit(doc.id, 456); }`)
4. Save it, go back to the all_docs page again
5. In the sidebar, click on the first View.

Result: the Map Function should should the first view's map function & not the second. Current master incorrectly shows the second map function.

michellephung:
+1

benkeen:
Merged as 010557e

PR #538, opened 2015-09-28
Fix for error caused by newlines in View search -> keys field

benkeen:
Curious, very specific bug. When doing a search on a View and entering
a value for the Keys field that contains a newline between two values,
the encoding of the URL gets all messed up and you see a
"missing_named_view" error get thrown within the UI. This ticket simply
strips out newlines in that field.

How to reproduce:
- create a new view
- click on the "Query Options" section and click on Keys to expand
that section. Enter:

["one",
"two"]

Note the newline after the "one" line.
- submit the form.

Result: JS errors and the error notification.

### DIFF
--- a/app/addons/documents/queryoptions/stores.js
+++ b/app/addons/documents/queryoptions/stores.js
@@ -211,7 +211,7 @@ function (app, FauxtonAPI, ActionTypes) {
           params.end_key = betweenKeys.endkey;
         }
       } else if (this._showByKeys) {
-        params.keys = this._byKeys;
+        params.keys = this._byKeys.replace(/\r?\n/g, '');
       }
 
       if (this._updateSeq) {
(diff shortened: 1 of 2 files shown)


app/addons/documents/queryoptions/stores.js
```diff
@@ -211,7 +211,7 @@ function (app, FauxtonAPI, ActionTypes) {
           params.end_key = betweenKeys.endkey;
         }
       } else if (this._showByKeys) {
-        params.keys = this._byKeys;
+        params.keys = this._byKeys.replace(/\n/g, '');
```
robertkowalski: that probably won't work for windows `\r\n` newlines

untested alternative: `/\r?\n/`

benkeen: Good point! That makes sense.

benkeen:
Thanks @robertkowalski. Confirmed that this now works on Windows.

robertkowalski:
+1, merge it! :)

benkeen:
Merged as b5936be

PR #539, opened 2015-09-28
Layout CSS fixes

benkeen:
This fixes a small problem with IE10 and Safari where the right
header isn't properly positioned. To see the error just load the
Databases page in one of those browsers in current master.


ocelotpotpie:
+1 @benkeen Looks good to me.

benkeen:
Thanks @ocelotpotpie :)

PR #541, opened 2015-09-30
Remove old, incorrect classes

benkeen:
Apologies @robertkowalski you were quite right on an earlier ticket:
these classes should have been removed; I didn't see the problem
in Fauxton.

To confirm: check the Edit View page layout looks normal.


michellephung:
+1

benkeen:
Merged as cde3216

PR #542, opened 2015-09-30
Add a check to confirm test exists

benkeen:
A pet peeve with running a single mocha test on the command line,
like `grunt file --file=./a/b/c.js` is that (a) it spends several
seconds compiling, copying etc. then (b) just ends. It doesn't
notify you that the file doesn't exist, you just have to infer it
or spend a few minutes being confused.

This little PR adds a test that runs up front: if the file you
entered doesn't exist, it lets you know about it and stops the
process immediately.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -364,6 +364,7 @@ module.exports = function (grunt) {
       }
     },
     gen_initialize: templateSettings,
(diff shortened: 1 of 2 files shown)


robertkowalski:
nice catch, +1

benkeen:
Merged as 6203825

PR #543, opened 2015-10-01
a table view for fauxton

robertkowalski:
this PR is intended to discuss the feature and design

the goal: display more information on less space in fauxton / increase information density

known issues: if the display width goes beyond 1365px the document search goes down. not showing it (usual responsive design) would remove the search feature for macbook 13" users at whole i guess?

known issues 2: i just copy pasted the css from the dataimporter table, we have to put it in a shared location when one of the features landed


robertkowalski:
cc @elistevens

michellephung:
:( this toggle bar looks wierd here:
screen shot 2015-10-01 at 8 10 58 pm

lightening the border makes it look *less* weird, 
screen shot 2015-10-01 at 8 11 03 pm
screen shot 2015-10-01 at 8 33 56 pm
and then taking out the background, is an improvement still:
screen shot 2015-10-01 at 9 13 42 pm

but really, the style of the toggle still doesn't fit in with the overall design of the page.

I do really like the search bar coming down. :+1:  Its okay that it overlaps the cards, it's pretty unintrusive. It is definitely an improvement on what was happening before, when it would just become unreachable at smaller widths. Lets keep this as a feature, this is not a bug to me. :D

The table is okay. Not super happy about it, but if we stick with it, we'll really have to address the include_docs=true situation, because the default currently, is really confusing:
screen shot 2015-10-01 at 8 29 19 pm

It's all scary random strings, and if even if those _ids were legible, they wouldn't really give that much information, unless include_docs=true.

I say take default  include_docs=true, AFTER we think about something to do about the databases which have way too much to show on the cards. Maybe throttle/truncate the data for those folks, by setting a max length for the cards? (just a first idea, better ideas welcome)


there was some discussion before about moving the query things into to upper space of the rightmost section where the cards are, which is a good idea.

Lets see what others have to say. 
You've got me thinking about it though, more ideas will probably formulate in the days to come.

:D looks cool though, just a little needs some polishing.

In the meantime.... 
1. I'd really love to see futon-style table inside the json cards:
screen shot 2015-09-28 at 4 29 23 am
screen shot 2015-10-01 at 9 40 23 pm
(stick with existing font-family/size/color though. I just didnt have time to change the font here^^)

sean/andrea mocked up something like this from before.  It's around somewhere. I remember sean showing it to us and i really like it. I didn't spend too much time on the colors, pick colors that match what's going already, or whatever, it'll be easy to change the colors in css. It's just the formatting of the table I'm after. This is hard to read: screen shot 2015-09-28 at 4 29 08 am. 

2. and the repeated thrice _id/_rev!!! to be taken out of the cards because they are taking up tons of space!
screen shot 2015-10-01 at 9 51 25 pm
(if you have time :sweat_smile: )

elistevens:
Thanks for working on this!

I'd like to reiterate just how usable the futon implementation is. The core UI for views is really quite good (in terms of medium-size-text _id, small-size-text _rev underneath, 2nd column for value). I would suggest starting cloning it as closely as possible, then adding a 3rd row only when include_docs=true (please not on by default). Pretty-printing the value and the doc would be nice.

robertkowalski:
hey @michellephung - thanks for your detailed feedback!

> :( this toggle bar looks wierd here:
> lightening the border makes it look less weird, 

i got the colors for the border from dante, i can change it

I like the cards proposal, but i would propose to keep this PR focussed on the table layout. the cards proposal is something that probably goes more into the direction of a feature called "redesign of the json cards". this PR is about the listing of docs, let's keep it focussed so we can deliver some value to our users soon.

robertkowalski:
hey @elistevens - thanks for the fast feedback!

this PR is about listing documents, would the "compact" -view mode resemble the listing from futon, when we reduce the paddings / space around each element? should we add the rev id?


**comparison:**

bildschirmfoto 2015-10-02 um 12 52 44

bildschirmfoto 2015-10-02 um 12 52 32

elistevens:
Ah, sorry - I misunderstood the "view" in the title.

Showing `_rev` in the all-docs page is absolutely helpful. I have documents that during the course of processing get about 65 updates. Being able to tell at a glance what revision those docs are tells me a lot about what might have happened (all at `_rev` 1? Then the task runner didn't do anything; a couple at `_rev` 13? Probably those choked on something partway through; etc.).

And I also agree with the more compact listing. 

Is having a more consistent dark-text-on-white (or vice versa) going to be part of this work?

michellephung:
Here's one idea:
all-docs-3-toggle-table
all-docs-3-toggle-table-dropdown
that doesn't break the design of the page, although....I wish we could get rid of these tray things, in general.

icons, here:
http://fortawesome.github.io/Font-Awesome/icon/list-alt/
http://fortawesome.github.io/Font-Awesome/icon/bars/
http://fortawesome.github.io/Font-Awesome/icon/th/

michellephung:
for the css table:
screen shot 2015-10-04 at 3 36 02 am

app/addons/components/react-components.react.jsx
```diff
@@ -60,6 +62,94 @@ function (app, FauxtonAPI, React, Stores, FauxtonComponents, ace, beautifyHelper
...
+    getMasterSelector: function () {
+      return (
+        <div className="group-clean bulk-actions-panel panel">
+          <input type="checkbox"
+            checked={this.props.isChecked}
+            onChange={this.props.toggleSelect}
+            disabled={this.props.disabled} />
+          <div className="separator"></div>
+          <ReactBootstrap.OverlayTrigger
+            ref='popover'
```
michellephung: thing

app/addons/documents/index-results/index-results.components.react.jsx
```diff
@@ -36,6 +37,142 @@ function (app, FauxtonAPI, React, Stores, Actions, Components, Documents) {
...
+
+    maybeGetCheckboxCell: function (i) {
+      if (!this.props.docId) {
+        return null;
+      }
+
+      return (
+        <td className="tableview-checkbox-cell" key={"tableview-checkbox-cell-" + i}>
+          <input
+            id={'checkbox-' + this.props.docIdentifier}
```
michellephung: ' '

app/addons/documents/index-results/index-results.components.react.jsx
```diff
@@ -36,6 +37,142 @@ function (app, FauxtonAPI, React, Stores, Actions, Components, Documents) {
...
+
+      var res = data.map(function (el, i) {
+
+        return (
+          <TableRow
+            key={"tableview-row-component-" + i}
+            index={i}
+            data={el}
+            docId={el.id}
+            docIdentifier={el.id || 'tableview-row-component-' + i}
```
michellephung: ' '

app/addons/documents/index-results/stores.js
```diff
@@ -199,6 +216,91 @@ function (FauxtonAPI, ActionTypes, HeaderActionTypes, Documents, MangoHelper) {
...
+      };
+    },
+
+    getHasEditableAndDeleteableDoc: function (data) {
+      var found = false;
+      var i = 0;
+      var length = data.length;
+
+      // use a for loop here as we can end it once we found the first id
+      for (; i < length; i++) {
```
michellephung: not a big deal, but it's traditional to initialize i=0 inside the for loop. This stood out for some reason. But it's not a blocker.

app/addons/documents/index-results/tests/index-results.componentsSpec.react.jsx
```diff
@@ -13,79 +13,196 @@ define([
...
+      it('you can change the default text', function () {
+        IndexResultsActions.newResultsList({
+          collection: [],
+          deleteable: true,
+          textEmptyIndex: 'I <3 Hamburg'
+        });
+
+        instance = TestUtils.renderIntoDocument(<Views.List />, container);
+        var $el = $(instance.getDOMNode());
+        assert.equal($el.text(), 'I <3 Hamburg');
```
michellephung: :heart:

michellephung:
UI looks fantastic :D 
I'm still unhappy that Collapse View and Expanded View does not give the viewer any new nor different information, which raises the question, why do we have both?

Collapsed View:
screen shot 2015-10-16 at 5 29 00 pm

Expanded View:
screen shot 2015-10-16 at 5 29 06 pm

Everything else is great.

The search bar is still disappearing at small widths, and now there is space to put it in the document cards space, but I realize that will take longer.

This is great!

+1

michellephung:
Ah, I just noticed also, 
screen shot 2015-10-16 at 5 34 04 pm

maybe have a minimum width for the columns?

app/addons/documents/index-results/index-results.components.react.jsx
robertkowalski: what do you mean?

app/addons/components/react-components.react.jsx
robertkowalski: i don't quite understand, in case you meant that the name is too generic, i changed it too `bulkActionPopover` now

app/addons/documents/index-results/stores.js
robertkowalski: good point, readability and making it easy to can the code fast is super important. fixed!

robertkowalski:
> I'm still unhappy that Collapse View and Expanded View does not give the viewer any new nor different information, which raises the question, why do we have both?

the do, when you have `include_docs` enabled

> maybe have a minimum width for the columns?

we want to implement a small dropdown/input field/thingy in the second iteration where you can select which fields you are interested in. we also want to guess the most interesting fields based on their appearance in the future. a min-width would imply that we scroll horizontally and that is something we try to avoid.

app/addons/documents/index-results/index-results.components.react.jsx
michellephung: these should be "

app/addons/components/assets/less/bulk-selector.less
```diff
@@ -0,0 +1,122 @@
...
+    &:hover {
+      background-color: #fbcece; // light red
+    }
+    &:last-child {
+      -webkit-border-bottom-right-radius: 5px;
+      -webkit-border-bottom-left-radius: 5px;
+      -moz-border-radius-bottomright: 5px;
+      -moz-border-radius-bottomleft: 5px;
+      border-bottom-right-radius: 5px;
+      border-bottom-left-radius: 5px;
```
benkeen: Be nice to move this CSS chunk to a mixin.

benkeen:
This looks SWEET. Really, really excited by this ticket.

UI remarks (for Dante?):
- I find "View" as the default text for the different views a little unclear until I click on it. We shied away from just having the three icons in a row? (each with a tooltips, say?) 
- maybe a pale grey line to the right of View, just to make it like the other header btns? 
- might be nice to change up the icon in the header based on the selected View rather than just having the eye icon all the time. 
- the clickable link locations to take you to the full page doc editor differ depending on whether you're on the all docs or a View. All docs has _id in the first column; Views have the last. Any chance that could be consistent? I'd favour the first column.
- the greys are a little weak for the bulk selector, in context of Fauxton. See: 
screen shot 2015-10-19 at 7 55 24 am

Super small issues:
- the down arrow on the bulk selector widget should show a cursor:pointer to indicate clickable. The clickable area should be the full square behind it, too. 
- when you click from page to page, the bulk selector widget noticeably checks then unchecks. A bit unsightly. 
- the garbage can icon in the bulk selector that appears is perhaps 1 or 2px left of centre.
- the "View" header button shouldn't have a 1px border left: it appears a little too thick (see screenshot above)

Other:
- selecting all rows that can be deleted doesn't check them all. How come? I'm on a default _all_docs with default query options. Seems curious you can't delete them all. 
- could we put the View selection (expanded, table, collapsed) into local storage? I think that a lot of people are going to want table view to be the default & won't want to have to re-select it after each page refresh.

benkeen:
Besides the things I mention, a big +1 from me. The UI comments are just nit-picks, I'd be very happy with it going out as is.

benkeen:
[Also, was the license for the new react lib added to couchDB?]

justin-mcdavid-ibm:
Comments below (on things design-related).

============================

**I find "View" as the default text for the different views a little unclear until I click on it. We shied away from just having the three icons in a row? (each with a tooltips, say?)**

We had (still have) a lot of problem with horizontal real estate in that header.  I believe Robert is planning to move the search box down to right next to the bulk selector, which might give us the room we'd need for that.  But, frankly, bringing the "Include Docs" check box out of the Query Options panel might be a better use of that space.

**maybe a pale grey line to the right of View, just to make it like the other header btns?**

I agree. Left justification helps, but we should give View the same button-ish treatment.

**might be nice to change up the icon in the header based on the selected View rather than just having the eye icon all the time.**

I agree. Good call, Ben. If there is a default docs display, then that means doing away with the eye icon.

**the clickable link locations to take you to the full page doc editor differ depending on whether you're on the all docs or a View. All docs has _id in the first column; Views have the last. Any chance that could be consistent? I'd favour the first column.**

I'm not sure I understand the question. I believe that we are having all table values be clickable. When ID isn't displayed, we still need a way to evoke the editor. Docs (collapsed or uncollapsed) always have doc ID at the top, and that ID is clickable.

**the greys are a little weak for the bulk selector, in context of Fauxton. Maybe they should be a hair darker?**

I agree. Or if we can dial up the weight of the lines.

**the down arrow on the bulk selector widget should show a cursor:pointer to indicate clickable. The clickable area should be the full square behind it, too.**

I agree.

I think we need to also obviate the bulk selector behavior (select all docs in the database vs. select all docs on the page). Feedback showing the number of docs currently selected would help, if it'd be possible to provide this information on the same line as the bulk selector.

robertkowalski:
@justin-mcdavid-ibm @benkeen 

the current border is like the borders of the buttons `1px solid #cccccc` -- need another color code. i tried #666 (color of the button icons in the header, but it is a bit too dark)

benkeen:
Don't shoot me, two very minor things:

screen shot 2015-10-20 at 10 05 55 am

- Could the little up arrow that appears at the top of the modal be the same colour as the grey header behind "Select all docs that can be..."? (`#f7f7f7`)
- the colour for the down arrow in the bulk selector is too light now. Looks like it's disabled.

robertkowalski:
@benkeen fixed the arrow-up issue. talked to justin -- we are disabling the overlay for now. still waiting for a color code for the borders and icons.

benkeen:
> we are disabling the overlay for now

Oh, pity! It was totally unnecessary, for sure, but it looks great :)

justin-mcdavid-ibm:
I think #aaa should do for that color, Robert.  I played around with it a bit, and it looks ok.  Thanks.

app/addons/components/assets/less/bulk-selector.less
```diff
@@ -0,0 +1,151 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+
+#bulk-action-component-popover {
+  font-size: 13px;
+
+  i {
```
michellephung: this is kind of odd give all italics a margin right

app/addons/components/react-components.react.jsx
```diff
@@ -60,6 +62,101 @@ function (app, FauxtonAPI, React, Stores, FauxtonComponents, ace, beautifyHelper
...
+
+    getMultiSelectOptions: function () {
+      if (!this.props.hasSelectedItem) {
+        return null;
+      }
+
+      return (
+        <button
+          onClick={this.props.removeItem}
+          className={'fonticon ' + this.props.bulkIcon}
```
michellephung: double quotes

app/addons/documents/index-results/index-results.components.react.jsx
```diff
@@ -75,31 +213,71 @@ function (app, FauxtonAPI, React, Stores, Actions, Components, Documents) {
...
+    render: function () {
+
+      var loadLines = null;
+      var isTableView = this.props.isTableView;
+
+      if (this.props.isLoading) {
+        loadLines = <Components.LoadLines />;
+      }
+
+      var mainview = isTableView ? this.getTablestyleView(loadLines) : this.getDocumentStyleView(loadLines);
```
michellephung: this.getTablestyleView --> getTableStyleView

michellephung: mainview --> mainView

app/addons/documents/index-results/stores.js
```diff
@@ -179,11 +180,27 @@ function (FauxtonAPI, ActionTypes, HeaderActionTypes, Documents, MangoHelper) {
     },
 
     getResults: function () {
+      var hasEditableAndDeleteableDoc;
```
michellephung: hasEditableAndDelet///e///ableDoc --> hasEditableAndDeletableDoc,

and maybe this should be a logic thing? if(isEditableDoc && isDeletableDoc)then...

app/addons/documents/index-results/stores.js
```diff
@@ -199,6 +216,100 @@ function (FauxtonAPI, ActionTypes, HeaderActionTypes, Documents, MangoHelper) {
             isEditable: this.isEditable(doc)
           };
         }, this);
+
+      hasEditableAndDeleteableDoc = this.getHasEditableAndDeleteableDoc(res);
```
michellephung: getHasEditableAndDelet//e//ableDoc

app/addons/components/assets/less/bulk-selector.less
michellephung: maybe a class would be better?

michellephung:
UI looks good, I really like how the new 'select all' functionality works.
There are some minor things that I've commented.

+1 if you want to merge quickly.

app/addons/components/react-components.react.jsx
robertkowalski: hm, we have no rule that we have to use double quotes in a computed property. i usually go with single quotes as it is a computed one, and not declarative markup

michellephung: ah that makes sense now for all the other single quotes in the pr :) +1

robertkowalski:
merged as 246057e64ee0683a44f1c9c1ca1d74f75638a740

PR #544, opened 2015-10-01
Notification Center

benkeen:
This PR adds a new notification center panel to Fauxton
that collects and displays all notifications that occur
during a users session so they're no longer lost.

A bell icon now appears at the top right of all pages 
opens the notification center. That shows you a log 
all notifications with the option to clear single ones, clear 
all, and view by group.

screen shot 2015-10-01 at 1 14 06 pm

screen shot 2015-10-01 at 1 13 55 pm

### DIFF
--- a/app/addons/fauxton/components.react.jsx
+++ b/app/addons/fauxton/components.react.jsx
@@ -34,12 +34,27 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
   var Clipboard = React.createClass({
     propTypes: function () {
       return {
-        text: React.PropTypes.string.isRequired
+        text: React.PropTypes.string.isRequired,
+        displayType: React.PropTypes.string.oneOf(['icon', 'text'])
+      };
+    },
+
+    getDefaultProps: function () {
+      return {
+        displayType: 'icon',
+        textDisplay: 'Copy'
       };
     },
 
     componentWillMount: function () {
       ZeroClipboard.config({ moviePath: getZeroClipboardSwfPath() });
+    },
+
+    getClipboardElement: function () {
+      if (this.props.displayType === 'icon') {
+        return (<i className="fonticon-clipboard"></i>);
+      }
+      return this.props.textDisplay;
     },
 
     componentDidMount: function () {
@@ -49,8 +64,8 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
 
     render: function () {
       return (
-        <a href="#" ref="copy" data-clipboard-text={this.props.text} data-bypass="true" title="Copy to clipboard">
-          <i className="fonticon-clipboard"></i>
+        <a href="#" ref="copy" className="copy" data-clipboard-text={this.props.text} data-bypass="true" title="Copy to clipboard">
+          {this.getClipboardElement()}
         </a>
       );
     }
--- a/app/addons/fauxton/tests/componentsSpec.react.jsx
+++ b/app/addons/fauxton/tests/componentsSpec.react.jsx
@@ -28,7 +28,7 @@ define([
       container = document.createElement('div');
 
       // when we want to control the diff, we have to render directly
-      trayEl = React.render(<Views.Tray className="traytest" />, container);
+      trayEl = TestUtils.renderIntoDocument(<Views.Tray className="traytest" />, container);
 
       oldToggleSpeed = FauxtonAPI.constants.MISC.TRAY_TOGGLE_SPEED;
 
@@ -109,7 +109,7 @@ define([
         }
       });
 
-      var reactEl = React.render(React.createElement(wrapper), container);
+      var reactEl = TestUtils.renderIntoDocument(React.createElement(wrapper), container);
       reactEl.runTest();
 
       React.unmountComponentAtNode(container);
@@ -135,7 +135,7 @@ define([
     });
 
     it('renders 20-wise pages per default', function () {
-      var pageEl = React.render(
+      var pageEl = TestUtils.renderIntoDocument(
         <Views.Pagination page={3} total={55} urlPrefix="?prefix=" urlSuffix="&suffix=88" />,
         container
       );
@@ -152,7 +152,7 @@ define([
     });
 
     it("can overwrite collection size", function () {
-      var pageEl = React.render(
+      var pageEl = TestUtils.renderIntoDocument(
         <Views.Pagination perPage={10} page={3} total={55} urlPrefix="?prefix=" urlSuffix="&suffix=88" />,
         container
       );
@@ -162,7 +162,7 @@ define([
     });
 
     it("handles large collections properly - beginning", function () {
-      var pageEl = React.render(
+      var pageEl = TestUtils.renderIntoDocument(
         <Views.Pagination page={3} total={600} />,
         container
       );
@@ -175,7 +175,7 @@ define([
     });
 
     it("handles large collections properly - middle", function () {
-      var pageEl = React.render(
+      var pageEl = TestUtils.renderIntoDocument(
         <Views.Pagination page={10} total={600} />,
         container
       );
@@ -189,7 +189,7 @@ define([
     });
 
     it("handles large collections properly - end", function () {
-      var pageEl = React.render(
+      var pageEl = TestUtils.renderIntoDocument(
         <Views.Pagination page={29} total={600} />,
         container
       );
@@ -204,5 +204,32 @@ define([
 
   });
 
-});
 
+  describe('Clipboard', function () {
+    var container;
+    beforeEach(function () {
+      container = document.createElement('div');
+    });
+
+    afterEach(function () {
+      React.unmountComponentAtNode(container);
+    });
+
+    it('shows a clipboard icon by default', function () {
+      var clipboard = TestUtils.renderIntoDocument(<Views.Clipboard text="copy me" />, container);
+      assert.equal($(clipboard.getDOMNode()).find('.fonticon-clipboard').length, 1);
+    });
+
+    it('shows text if specified', function () {
+      var clipboard = TestUtils.renderIntoDocument(<Views.Clipboard displayType="text" text="copy me" />, container);
+      assert.equal($(clipboard.getDOMNode()).find('.fonticon-clipboard').length, 0);
+    });
+
+    it('shows custom text if specified ', function () {
+      var clipboard = TestUtils.renderIntoDocument(<Views.Clipboard displayType="text" textDisplay='booyah!' text="copy me" />, container);
+      assert.ok(/booyah!/.test($(clipboard.getDOMNode())[0].outerHTML));
+    });
+
+  });
+
+});
--- a/app/addons/auth/routes.js
+++ b/app/addons/auth/routes.js
@@ -29,9 +29,10 @@ function (app, FauxtonAPI, Auth, AuthActions, Components, ClusterActions) {
       'login': 'login',
       'logout': 'logout',
       'createAdmin': 'checkNodes',
-      'createAdmin/:node': 'createAdminForNode',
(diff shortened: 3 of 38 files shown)


app/addons/fauxton/components.react.jsx
```diff
@@ -42,6 +54,14 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
       ZeroClipboard.config({ moviePath: getZeroClipboardSwfPath() });
     },
 
+    getClipboardElement: function () {
+      if (this.props.displayType === 'icon') {
+        return (<i className="fonticon-clipboard"></i>);
+      } else {
```
robertkowalski: you can use an early return here

app/addons/fauxton/components.react.jsx
```diff
@@ -334,13 +354,246 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
...
+            item={notification}
+            filter={this.state.filter}
+            key={notification.notificationId}
+          />
+        );
+      }, this);
+    },
+
+    selectFilter: function (e) {
+      var filter = $(e.target).closest('li').data('filter');
```
robertkowalski: jquery?

app/addons/fauxton/components.react.jsx
```diff
@@ -334,13 +354,246 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
...
+              <li className={filterClasses.success} data-filter="success" title="Success notifications">
+                <span className="fonticon fonticon-ok-circled"></span>
+              </li>
+              <li className={filterClasses.error} data-filter="error" title="Error notifications">
+                <span className="fonticon fonticon-attention-circled"></span>
+              </li>
+              <li className={filterClasses.info} data-filter="info" title="Info notifications">
+                <span className="fonticon fonticon-info-circled"></span>
+              </li>
+            </ul>
```
robertkowalski: on possible way to solve this. just add the event handler on all three nodes, react will take care of the memory management / delegation internally:

```html
            <ul className="notification-filter flex-layout flex-row">
              <li  onClick={this.selectFilter.bind(this, 'all')} className={filterClasses.all} title="All notifications">All</li>
            
   [...]
```

this also removes the need for jquery and accessing the dom directly, what we should really avoid

app/addons/fauxton/components.react.jsx
```diff
@@ -334,13 +354,246 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
...
+
+      var show = true;
+      if (this.props.filter !== 'all') {
+        show = this.props.item.type === this.props.filter;
+      }
+      if (show) {
+        console.log(this.state.elementHeight);
+        $(this.getDOMNode()).velocity({ opacity: 1, height: this.state.elementHeight }, this.props.transitionSpeed);
+      } else {
+        this.hide();
```
robertkowalski: early returns make it easier readable and you don't need to nest that much

app/addons/fauxton/components.react.jsx
```diff
@@ -334,13 +354,246 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
...
+          elementHeight: this.getHeight()
+        });
+      }
+
+      var show = true;
+      if (this.props.filter !== 'all') {
+        show = this.props.item.type === this.props.filter;
+      }
+      if (show) {
+        console.log(this.state.elementHeight);
```
robertkowalski: debug?

app/addons/fauxton/components.react.jsx
```diff
@@ -334,13 +354,246 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
...
+      // for testing purposes
+      var visible = (this.props.filter === 'all' || this.props.filter === this.props.item.type) ? 'true' : 'false';
+
+      // N.B. wrapper <div> needed to ensure smooth hide/show transitions
+      return (
+        <li data-visible={visible}>
+          <div className={classes}>
+            <span className={rowIconClasses}></span>
+            <div className="flex-body">
+              <p dangerouslySetInnerHTML={{__html: this.props.item.msg}}></p>
```
robertkowalski: we have some notifications that include user generated content, so we open a XSS vector here i guess?

app/addons/fauxton/stores.js
```diff
@@ -0,0 +1,132 @@
...
+    },
+
+    isNotificationCenterVisible: function () {
+      return this._notificationCenterVisible;
+    },
+
+    addNotification: function (info) {
+      if (_.isEmpty(info.type) || !_.contains(validNotificationTypes, info.type)) {
+        console.warn('Invalid message type: ', info);
+        return;
```
robertkowalski: <3

app/addons/fauxton/tests/componentsSpec.react.jsx
```diff
@@ -204,5 +207,169 @@ define([
...
+    beforeEach(function () {
+      container = document.createElement('div');
+    });
+
+    afterEach(function () {
+      React.unmountComponentAtNode(container);
+    });
+
+    it('shows all notification types when "all" filter applied', function () {
+      var row1 = React.render(
```
robertkowalski: this all should use `TestUtils.renderIntoDocument` i guess?

app/addons/fauxton/components.react.jsx
```diff
@@ -334,13 +354,246 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
...
+      // we can safely do this because the store ensures all notifications are of known types
+      var rowIconClasses = 'fonticon ' + iconMap[this.props.item.type];
+      var classes = 'flex-layout flex-row';
+
+      // for testing purposes
+      var visible = (this.props.filter === 'all' || this.props.filter === this.props.item.type) ? 'true' : 'false';
+
+      // N.B. wrapper <div> needed to ensure smooth hide/show transitions
+      return (
+        <li data-visible={visible}>
```
robertkowalski: it looks like this was added just for testing?

app/addons/fauxton/tests/componentsSpec.react.jsx
```diff
@@ -204,5 +207,169 @@ define([
...
+    it('shows all notifications by default', function () {
+      store.addNotification({ type: 'success', msg: 'Success are okay' });
+      store.addNotification({ type: 'success', msg: 'another success.' });
+      store.addNotification({ type: 'info', msg: 'A single info message' });
+      store.addNotification({ type: 'error', msg: 'Error #1' });
+      store.addNotification({ type: 'error', msg: 'Error #2' });
+      store.addNotification({ type: 'error', msg: 'Error #3' });
+
+      var panelEl = React.render(<Views.NotificationCenterPanel />, container);
+      assert.equal($(panelEl.getDOMNode()).find('.notification-list li[data-visible=true]').length, 6);
```
robertkowalski: what is that `data-visible` thing?

robertkowalski: it feels like an jquery application with some react on top

app/addons/fauxton/components.react.jsx
```diff
@@ -334,13 +354,246 @@ function (app, FauxtonAPI, React, ZeroClipboard) {
...
+            <span className={rowIconClasses}></span>
+            <div className="flex-body">
+              <p dangerouslySetInnerHTML={{__html: this.props.item.msg}}></p>
+              <div className="notification-actions">
+                <span className="time-elapsed">{timeElapsed}</span>
+                <span className="divider">|</span>
+                <Clipboard text={this.props.item.msg} displayType="text" />
+              </div>
+            </div>
+            <button type="button" aria-hidden="true" onClick={this.clearNotification}>×</button>
```
robertkowalski: `aria-hidden="true"` <- is it always hidden?

robertkowalski:
does the new notification center also mean we can throw away the old notification code?

robertkowalski:
code looks quite well! just spotted some jqueryisms

app/addons/fauxton/components.react.jsx
benkeen: I think it's okay in this case, no? I need to do a little hunting in the DOM to locate the parent list item based on whatever element triggered the event, so jQuery seemed the simplest.

app/addons/fauxton/components.react.jsx
benkeen: Yeah, this would work... I think whether it's an improvement is debatable, though, having to add 3X the number of event handlers. (What if there were 10 items & they couldn't be added via a loop?) But fair enough point about generally wanting to avoid accessing DOM directly.

app/addons/fauxton/components.react.jsx
benkeen: Oops.

app/addons/fauxton/components.react.jsx
benkeen: Drat, excellent point.

I liked passing in the HTML here because some messages contained links that would continue to work from within the notification centre panel. But on further thought, it could backfire... not just with the XSS - but with links that were relative, or whatever. Potentially there'd be broken links appearing in the notification centre.

I'll change this to strip out HTML + just pass the msg itself.

app/addons/fauxton/components.react.jsx
benkeen: Yeah, I do hate to do it but it's an acceptable sacrifice, I think. :( We do it all over the place with things like modals to test they're open & not run into timing issues with fade-in/outs.

app/addons/fauxton/tests/componentsSpec.react.jsx
benkeen: It's just for testing whether it's visible or not. The problem is just timing: any of the nice transition effects we do (bootstrap, velocity) take time to perform so checking visibility doesn't work immediately. This is a workaround. I *could* use a class, but I kinda prefer a data-role because to me it seems more detached from a class which could have styles. I know data-roles could be styled too, but they are less likely to & I like to draw attention to it.

app/addons/fauxton/components.react.jsx
benkeen: Good one.

benkeen:
Thanks @robertkowalski! Terrific feedback. Will apply the various fixes today.

Yes, ultimately we'll be able to drop the old notification code but not yet. I actually started doing that at the beginning of this ticket, but felt I was kinda shaving a yak, so left it alone until later. :)

michellephung:
Looks great! can we make the old notifications smaller at least?

michellephung:
+1

app/addons/fauxton/components.react.jsx
robertkowalski: i've seen that for the first time someone doing in react.

i am really against storing state in the dom like we did in jquery times, that's one of the reason why we chose react/flux, to enforce a separation of concerns. we run into two issues here:

1. adding code to production code for a test (not to confuse with making code testable)
2. storing state in the DOM

there are multiple ways to avoid this, one is testing for visibility in your tests:

```js
$('.mycomponent').is(':visible')
```

another way is to test if the class that gives visibility appears after an action

benkeen:
Thanks @michellephung! I'll merge this later today. I'm going to leave the old notifications alone for the moment + refactor them separately. I think the designers had some ideas.

michellephung:
sounds good

app/addons/fauxton/components.react.jsx
benkeen: Neither would work in this instance. The visibility is done by velocity sliding up/down the component, not by adding a class. I can add a class if you'd prefer that, but it seems the same as the data-role, only more confusing because people would think it had something to do with setting the visibility.

After the end of the action the height is zero and opacity is 0. So it should be technically visible, there - but I'll double check.

PR #546, opened 2015-10-02
Fix for Safari not recognizing autofill on login form

benkeen:
Safari doesn't trigger change events for form field autofills,
so the state of the login component doesn't updated, triggering
the rather irritating "you didn't fill in all the fields" error
when visually you sure did.

Since existing polyfill options don't work (see ticket), this fix
explicitly checks for the condition of a mismatch of
username-password form values and component state. It's a little
inelegant, but justifiable I think.

Closes COUCHDB-2829

### DIFF
--- a/app/addons/auth/components.react.jsx
+++ b/app/addons/auth/components.react.jsx
@@ -37,7 +37,11 @@ define([
 
     getDefaultProps: function () {
       return {
-        urlBack: ''
+        urlBack: '',
+
+        // for testing purposes only
+        testBlankUsername: null,
+        testBlankPassword: null
       };
     },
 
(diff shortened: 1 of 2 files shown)


garrensmith:
@benkeen this is cool, but we need some tests for this. I know its only reproduceable on Safari but hopefully you can write some unit tests to just exercise that code base.

benkeen:
Yeah, fair enough! I'll see what I can rustle up.

benkeen:
Boy, this proved to be an absolute devil to test. I tried different methods of altering the DOM to set the form field values without triggering a change, but nothing worked. That actually makes a whole lot of sense since the whole *point* of React is that there's a 1-1 map between the fields + state in cases like this.

The best I could come up with was adding a couple of component props intended to test this precise condition. The test uses them to confirm that non-state values are passed to the login action. I don't much like it, but it confirms the component works as expected for this given clause. Better ideas welcome!

garrensmith:
+1 nice work @benkeen

benkeen:
Thanks, Garren! Merged as 191ef07

PR #547, opened 2015-10-02
Add DOM check before mounting/unmounting React components

benkeen:
The recent API Bar refactor unveiled a rather neat problem that
showed up in the Dashboard only. The Replication page failed to
load and threw a React "Invariant Violation" error which we've
seen show up from time to time but couldn't yet reproduce. What
was happening was that when attempting to mount the component,
the DOM element just didn't exist. The same problem
could occur when unmounting.

Now to go down the rabbithole! This may get boring, be warned...

The reason it occurred in the dash only is because some page
layouts don't contain the #api-navbar element. It looks like
all Fauxton pages do.

Adding in a DOM check to confirm the existence of the apibar
at this point...
https://github.com/apache/couchdb-fauxton/blob/master/app/addons/fauxton/base.js#L64
... has no effect because the core routeObject uses the
`two_pane.html` layout template as a base template when loading,
and that DOES contain the element. But due to the complexity of
the way our pages load, that template gets loaded first, THEN
beforeFullRender runs, THEN the route object's template gets 
loaded, THEN the React components are rendered. So basically 
we don'tknow for sure whether the element exists until the 
moment its inserted/removed, hence the checks added in this PR.

I dedided to throw a warning in these cases. Seems like it should
be caught.

This fix should solve all of those Invariant Violation errors
of this kind, not just the problem with the API Bar. It was just
very nice having a clear reproducible instance to debug this.

### DIFF
--- a/app/core/tests/routeObjectSpec.js
+++ b/app/core/tests/routeObjectSpec.js
@@ -15,8 +15,10 @@ define([
   'testUtils'
 ], function (FauxtonAPI, React, testUtils) {
   var assert = testUtils.assert,
-  restore = testUtils.restore,
-  RouteObject = FauxtonAPI.RouteObject;
+      restore = testUtils.restore,
+      RouteObject = FauxtonAPI.RouteObject,
+      TestUtils = React.addons.TestUtils;
+
 
   describe('RouteObjects', function () {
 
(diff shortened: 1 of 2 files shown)


robertkowalski:
wow, thanks for the good writeup

+1

app/core/tests/routeObjectSpec.js
```diff
@@ -197,17 +197,6 @@ define([
           restore(React.unmountComponentAtNode);
         });
 
-        it('removes existing component via react', function () {
-          var spy = sinon.stub(React, 'unmountComponentAtNode');
-          var fakeSelector = 'remove-selector';
-          testRouteObject.reactComponents[fakeSelector] = React.createElement('div');
```
robertkowalski: you can just fix the existing test by creating the div with the fake selector

robertkowalski:
assuming there is no other test covering this codepath i would propose to fix the deleted test

app/core/tests/routeObjectSpec.js
benkeen: Bah, good point. Spent ages trying to construct a working test here.

benkeen:
Test has been patched. I'll merge once green, thanks @robertkowalski!

PR #548, opened 2015-10-05
Moving with_sidebar template to flexbox

benkeen:
This ticket updates the `with_sidebar.html` template which is used on
the account page (cluster of one) only right now, but it also touches on
other pages as well to clean up a little CSS, so a cursory look over the 
UI is probably a good idea.

After this one's merged, I'll tackle the `with_tabs_sidebar.html` template
which should allow me to remove a lot of old CSS!

### DIFF
--- a/app/addons/auth/assets/less/auth.less
+++ b/app/addons/auth/assets/less/auth.less
@@ -18,16 +18,14 @@
 
 .sidenav header {
   padding-left: 24px;
-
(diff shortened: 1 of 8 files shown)


michellephung:
+1

benkeen:
Just rebased with master after notification center merged. A few conflicts, now resolved.

PR #549, opened 2015-10-06
Improve db page error handling

benkeen:
This PR fixes an old bug with the databases page where if one
database query fails, the remaining responses that haven't been
returned don't get populated and the database table has a lot
of incomplete rows.

Now the individual failed database rows don't affect the others and
show a "This database failed to load" message to make it clear to
the user.

_______________

How to test:
- The https://issues.apache.org/jira/browse/COUCHDB-2243 ticket includes some instructions. For my part, I had Fiddler installed, so I just overwrote the `_all_dbs` request to return an array of valid database names, plus a made-up one in the middle.

_______________

Closes COUCHDB-2243

screen shot 2015-10-05 at 8 13 43 pm

### DIFF
--- a/app/addons/databases/actions.js
+++ b/app/addons/databases/actions.js
@@ -26,20 +26,36 @@ function (app, FauxtonAPI, Stores, ActionTypes, Resources) {
 
       this.setStartLoading();
       FauxtonAPI.when(databases.fetch({ cache: false })).then(function () {
-        FauxtonAPI.when(databases.paginated(page, perPage).map(function (database) {
-          return database.status.fetchOnce();
-        })).always(function () {
-          //make this always so that even if a user is not allowed access to a database
-          //they will still see a list of all databases
+
+        // if there are no databases, publish the init message anyway
+        if (!databases.paginated(page, perPage).length) {
           FauxtonAPI.dispatch({
             type: ActionTypes.DATABASES_INIT,
             options: {
-              collection: databases.paginated(page, perPage),
+              collection: [],
               backboneCollection: databases,
               page: page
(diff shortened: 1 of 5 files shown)


app/addons/databases/actions.js
```diff
@@ -26,20 +26,35 @@ function (app, FauxtonAPI, Stores, ActionTypes, Resources) {
...
-              collection: databases.paginated(page, perPage),
+              collection: [],
               backboneCollection: databases,
               page: page
             }
           });
-        }.bind(this));
+        }
       }.bind(this));
     },
```
robertkowalski: i don't understand this part of the diff

isn't it the same code as the map but written with an each and a counter that counts side effects? if so i would prefer the map back again, the counter thingy is a bit odd

to speed up the code you can also put the `if (!databases.paginated(page, perPage).length) {` above the other code and just return early

benkeen: Ah! Well spotted, this is actually the most interesting part of the whole PR. :) 

The old code was more elegant but buggy. `FauxtonAPI.when()` behaves in an unexpected way here, which is why it was refactored. When passed an array of promises it fires the `.always()` in two conditions:
1. When all promises are resolved, or,
2. When ONE fails

What was happening was that when a db request failed, it didn't fire any of the remainder requests, which is why the table would appear partly empty.

[and good point about the early return, I'll do that.]

robertkowalski: ah that makes sense now!

+1 after changing to early return and putting the description into the commit msg, in case someone wonders by reading the code in 1 year :)

PR #550, opened 2015-10-07
Grunt mocha test file existence check

benkeen:
This is a small tweak to a recent commit that does an up-front
check to confirm the mocha test exists. It had failed to check
the condition where the user was trying to run a .js mocha test
for a .jsx file that hadn't yet been compiled. This expands on the
existence check for a .jsx version of the file.


robertkowalski:
+1

PR #551, opened 2015-10-07
Fix for logout link showing up twice in footer

benkeen:
An old bug is the Logout link in the footer showing up twice. It's
a little difficult to reproduce consistently, but I've found this
works the best (Mac, chrome):
1. open dev tools (it slows down browser performance),
2. Login to Fauxton, notice the single "logout" link in footer.
3. click on the browser URL bar and click <enter>
4. Repeat the above until you see two "logout" links in the footer.

What's happening is that when you see two logout links, 
the change event on Auth.session is firing twice, each of which 
adds a new footer link.

This fix just ensures the link is removed prior to adding it.

### DIFF
--- a/app/addons/fauxton/navigation/stores.js
+++ b/app/addons/fauxton/navigation/stores.js
@@ -171,13 +171,14 @@ function (app, FauxtonAPI, ActionTypes) {
 
     dispatch: function (action) {
       switch (action.type) {
-
         case ActionTypes.ADD_NAVBAR_LINK:
           this.addLink(action.link);
         break;
+
         case ActionTypes.TOGGLE_NAVBAR_MENU:
           this.toggleMenu();
         break;
+
         case ActionTypes.UPDATE_NAVBAR_LINK:
           this.updateLink(action.link);
         break;
--- a/app/addons/auth/base.js
+++ b/app/addons/auth/base.js
@@ -29,7 +29,7 @@ function (app, FauxtonAPI, Auth) {
       title: "Login",
       href: "#login",
       icon: "fonticon-user",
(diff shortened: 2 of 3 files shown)


app/addons/fauxton/navigation/stores.js
```diff
@@ -173,6 +173,7 @@ function (app, FauxtonAPI, ActionTypes) {
       switch (action.type) {
 
         case ActionTypes.ADD_NAVBAR_LINK:
+          console.log(action.link);
```
robertkowalski: debug logging?

robertkowalski:
this needs tests ben

app/addons/fauxton/navigation/stores.js
benkeen: Oops. Will remove.

benkeen:
I thought about that, but `addHeaderLink` and `removeHeaderLink` already have tests. Testing a remove-then-add would just confirm that it's been removed before adding which is already tested. I guess I could test that `removeHeaderLink` fails nicely when it tries removing something that doesn't exist? But that may also already be tested I'll need to check.

robertkowalski:
how about taking the navbar and adding a link multiple times. assumption: the element should just be there one time then

robertkowalski:
this could be tested by emitting the change evvent multiple times on `Auth.session.on('change'`

benkeen:
Updated.

robertkowalski:
+1 after changing to single quotes :)

benkeen:
All fixed. I patched up all whitespace + quote issues in that file.

PR #552, opened 2015-10-07
make code for getting safe doc ids reusable

robertkowalski:
this puts the tools for getting a safe id for a doc into a shared
location.

part of https://github.com/apache/couchdb-fauxton/pull/543 which needs further design work but i need the safe-id functionality in other places now


michellephung:
+1

PR #553, opened 2015-10-08
Fix for header on full page doc editor

benkeen:
Drat, missed this one. This fixes a layout issue on the full
page doc editor header. Test included to confirm the breadcrumb
and API URL are always visible.


michellephung:
+1

PR #554, opened 2015-10-09
Pass data to API Bar via actions

benkeen:
The previous code used props on page load to set the API Bar
values (endpoint + doc URL). The problem was that because there
was no way to reference the APIBarController, it prohibited
on-the-fly updates to change the values without a route change.

This just tweaks it to store the API Bar data in a store that
can be updated by publishing an action. This will ultimately
allow us to do away with hacks like this:
https://github.com/apache/couchdb-fauxton/blob/master/app/addons/documents/header/header.react.jsx#L174

e.g. we could expand on this ticket and publish an
APIBAR_FADE_OUT” msg and the component would handle its own
fading out, rather than relying on other components directly
manipulating its DOM.

### DIFF
--- a/app/addons/components/react-components.react.jsx
+++ b/app/addons/components/react-components.react.jsx
@@ -14,13 +14,16 @@ define([
   'app',
   'api',
   'react',
+  'addons/components/stores',
   'addons/fauxton/components.react',
   'ace/ace',
   'plugins/beautify'
 ],
 
-function (app, FauxtonAPI, React, FauxtonComponents, ace, beautifyHelper) {
+function (app, FauxtonAPI, React, Stores, FauxtonComponents, ace, beautifyHelper) {
   var ReactCSSTransitionGroup = React.addons.CSSTransitionGroup;
+  var componentStore = Stores.componentStore;
+
 
   var ToggleHeaderButton = React.createClass({
     getDefaultProps: function () {
@@ -1076,11 +1079,30 @@ function (app, FauxtonAPI, React, FauxtonComponents, ace, beautifyHelper) {
 
   var ApiBarController = React.createClass({
 
-    getDefaultProps: function () {
+    getInitialState: function () {
+      return this.getStoreState();
+    },
+
+    getStoreState: function () {
       return {
-        endpoint: '_all_docs',
-        documentation: FauxtonAPI.constants.DOC_URLS.GENERAL
+        visible: componentStore.isAPIBarVisible(),
+        endpoint: componentStore.getEndpoint(),
+        docURL: componentStore.getDocURL()
       };
+    },
+
+    onChange: function () {
+      if (this.isMounted()) {
+        this.setState(this.getStoreState());
+      }
+    },
+
+    componentDidMount: function () {
+      componentStore.on('change', this.onChange, this);
+    },
+
+    componentWillUnmount: function () {
+      componentStore.off('change', this.onChange);
     },
 
     showCopiedMessage: function () {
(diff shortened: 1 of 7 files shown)


app/addons/components/react-components.react.jsx
```diff
@@ -1091,7 +1113,22 @@ function (app, FauxtonAPI, React, FauxtonComponents, ace, beautifyHelper) {
...
+      return (
+        <a className="help-link" data-bypass="true" href={this.state.docURL} target="_blank">
+          <i className="icon icon-question-sign"></i>
+        </a>
+      );
+    },
+
     render: function () {
+      if (!this.state.visible || !this.state.endpoint) {
+        return false;
```
robertkowalski: isn't returning `null` what we return usually?

app/addons/components/tests/storesSpec.js
```diff
@@ -0,0 +1,74 @@
...
+
+    afterEach(function () {
+      componentStore.reset();
+    });
+
+    it("sets reasonable defaults", function () {
+      assert.equal(componentStore.isAPIBarVisible(), true);
+      assert.equal(componentStore.getEndpoint(), '');
+      assert.equal(componentStore.getDocURL(), FauxtonAPI.constants.DOC_URLS.GENERAL);
+    });
```
robertkowalski: not sure how useful this test is

app/addons/components/tests/storesSpec.js
```diff
@@ -0,0 +1,74 @@
...
+      assert.equal(componentStore.getDocURL(), docURL);
+
+      var newDocURL = 'http://newwebsite.org';
+      ComponentActions.updateAPIBar({
+        docURL: newDocURL
+      });
+      assert.equal(componentStore.isAPIBarVisible(), true);
+      assert.equal(componentStore.getEndpoint(), newEndpoint);
+      assert.equal(componentStore.getDocURL(), newDocURL);
+    });
```
robertkowalski: this test just tests setter and getters

app/addons/components/tests/apiBarControllerSpec.react.jsx
```diff
@@ -0,0 +1,70 @@
...
+    it('Doesn\'t show up when set to visible BUT there\'s no endpoint defined', function () {
+      var el = TestUtils.renderIntoDocument(<ApiBarController />, container);
+      Actions.updateAPIBar({
+        visible: true,
+        endpoint: '',
+        docURL: 'http://link.com'
+      });
+      assert.equal(el.getDOMNode(), null);
+    });
+
```
robertkowalski: other tests: 

1. test the show / hide api bar actions and how the element behaves
2. test if the link doc url is rendered
3. test if the endpoint is rendered
4. test if the endpoint changes after an update and is rendered
5.test if the docUrl changes on update and is rendered

robertkowalski:
looks good, just some suggestions how to better test

benkeen:
Brilliant, thanks for the feedback @robertkowalski :) Will look at shortly.

app/addons/components/tests/storesSpec.js
benkeen: It tests `Actions.updateAPIBar()` too, to confirm that only the data being passed gets overwritten. But is testing the getters/setters bad? (Other than perhaps overkill)

app/addons/components/tests/storesSpec.js
benkeen: yeah.... you're probably right.

benkeen:
Good call on the tests. Added.

robertkowalski:
+1 merge it!

benkeen:
Thanks, Robert! Merged.

PR #556, opened 2015-10-14
Increase height of clickable space for close notification

benkeen:
There's a little "x" that appears in the notification center next
to each message that allows users to remove that single message.
On Firefox, the height of the clickable space wasn't quite large
enough so if you clicked the bottom of the x it wouldn't close.
This fixes it so that it covers the full "x".


michellephung:
+1

PR #558, opened 2015-10-15
Fix for doc typeahead field being hidden

benkeen:
When typing a doc name in the Doc ID field on the Database -> All
Docs page, it would get cut off. This PR ensures it overlaps
the body content.

Not thrilled with the extra `<span>` needed, but the position:relative 
high up in the tree caused issues, and this at least localizes the 
position:relative to the search icon.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ before_script:
   - ./node_modules/grunt-cli/bin/grunt dev &
   - sleep 25
 script:
-  - ./node_modules/grunt-cli/bin/grunt nightwatch
+  - ./node_modules/grunt-cli/bin/grunt nightwatch --file=selectDocViaTypeahead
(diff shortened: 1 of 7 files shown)


michellephung:
+1

benkeen:
Thanks for your help on this, Michelle! Just re-running one more time....

michellephung:
:D

robertkowalski:
hi ben,

i think that is not fixing the issue: https://cloudup.com/cZrYuRUH4lP

robertkowalski:
fun fact: it seems to work if you have visited the documents page before and don't start at the databases page

benkeen:
That's actually a different issue. I just fixed an issue on the database -> documents page. I think I know the cause of this, but it's an old bug.

PR #561, opened 2015-10-18
Allow tests to run on windows

sebastianrothbucher:
build-jsx also has this - just do the same for running the test(s). Won't do harm anywhere else I suppose


robertkowalski:
+1 thank you

PR #562, opened 2015-10-20
Add callback to Clipboard component

benkeen:
This just adds a callback to the `<Clipboard />` component that
gets called once the copy to clipboard action is completed.

Not testable due to Flash not being available on the testing environment.


robertkowalski:
+1

benkeen:
Merged as 230e5f7

PR #566, opened 2015-10-23
remove doc-editor check

robertkowalski:
i quite often run into the problem where i want to copy the json
of an existing doc for testing. i copy the whole json, click
"new doc" and have toremoev the _rev in order to save it.
sometimes i also want to remove the id, to get a random one. right
now i have to copy/paste the json into an offline editor before
that as the fauxton editor tells me i can't delete the rev or id
and refuses to make changes.


benkeen:
LGTM. Seems like a test to confirm that this functionality now works seems like overkill, so: +1

robertkowalski:
yes, as i deleted 34 lines of previously untested code

robertkowalski:
:)

PR #568, opened 2015-10-23
Change header z-index

benkeen:
This increases the z-index of the headers from 1 to 2. The reason
is that Ace Editor's selected row has a z-index of 1 and the API
URL bar is being incorrectly overlapped.


benkeen:
Merged as 9b99f52

PR #571, opened 2015-10-30
don't allow deletion of docs without rev

robertkowalski:
if we have no rev, bulk deletion is not possible, so we should
not try to offer it.


michellephung:
I merged this already.

e5f834e8f4eca0d3d0daa844f29dfcf9ea3f2537

PR #573, opened 2015-11-08
Allow multiple attachments to be uploaded at once.

mudcube:
The "multiple" attribute on a file input allows the user to select multiple attachments.


michellephung:
hi @mudcube,

This doesn't quite works as-is. When I select multiple attachments using this input, only one shows up in the document as an attachment.

mudcube:
Woopsidaisy, I was testing in Futon not Fauxton. I took a couple hours to attempt to figure out a way to get this working with FormData on Fauxton, but no success yet.

PR #575, opened 2015-11-14
Improve the strings

ykye:
Improve the strings.

### DIFF
--- a/app/addons/auth/components.react.jsx
+++ b/app/addons/auth/components.react.jsx
@@ -144,7 +144,6 @@ define([
     render: function () {
       return (
         <div className="auth-page">
-          <h3>Change Password</h3>
 
           <form id="change-password" onSubmit={this.changePassword}>
             <p>
@@ -154,12 +153,12 @@ define([
             <input id="password" type="password" ref="password" name="password" placeholder="Password"
               size="24" onChange={this.onChangePassword} value={this.state.password} />
             <br />
-            <input id="password-confirm" type="password" name="password_confirm" placeholder= "Verify Password"
+            <input id="password-confirm" type="password" name="password_confirm" placeholder="Password confirmation"
               size="24" onChange={this.onChangePasswordConfirm} value={this.state.passwordConfirm} />
 
             <br />
             <p>
-              <button type="submit" className="btn btn-primary">Change</button>
+              <button type="submit" className="btn btn-primary">Change Password</button>
             </p>
           </form>
         </div>
@@ -219,17 +218,18 @@ define([
     render: function () {
       return (
         <div className="auth-page">
-          <h3>Create Admins</h3>
+          <h3>Create User Accounts with ADMIN Privileges</h3>
 
           <p>
-            Before a server admin is configured, all clients have admin privileges. This is fine when
-            HTTP access is restricted to trusted users. <strong>If end-users will be accessing this
-            CouchDB, you must create an admin account to prevent accidental (or malicious) data
-            loss.</strong>
+            Before an admin user account is created for this CouchDB server instance,
+            all clients have ADMIN privileges.  <strong>If clients are not trusted users,
+            you must create an admin account to prevent accidental or malicious data loss.</strong>
           </p>
           <p>
-            Server admins can create and destroy databases, install and update _design documents, run
-            the test suite, and edit all aspects of CouchDB configuration.
+            Users with ADMIN privileges can create, modify, and delete databases and design documents;
+            run the test suite; and modify the CouchDB configuration. Users without ADMIN privileges
+            have read/write access to all databases. CouchDB can be <a href="http://docs.couchdb.org/en/1.6.1/config/auth.html"
+            target="_new" title="CouchDB link">configured</a> to block the access of anonymous users.
           </p>
 
           <form id="create-admin-form" onSubmit={this.createAdmin}>
@@ -239,9 +239,6 @@ define([
             <input id="password" type="password" name="password" placeholder= "Password" size="24"
               onChange={this.onChangePassword} />
             <p>
-              Non-admin users have read and write access to all databases, which
-              are controlled by validation functions. CouchDB can be configured to block all
-              access to anonymous users.
             </p>
             <button type="submit" id="create-admin" className="btn btn-primary">Create Admin</button>
           </form>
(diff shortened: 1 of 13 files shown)


kxepal:
@ykye why closed?

ykye:
@kxepal This is my first time to create a pull request for Fauxton. I created this PR, but I found it can be automatically merged. I wasn't very sure if the process of this PR is right. I asked @benkeen, he told me this is natural. So reopen this PR.

ykye:
@benkeen Please help have a look at.

kxepal:
@ykye Congrats with the first PR! (:
Don't worry about: automatically merged means that there are no merge conflicts that could involve people to resolve them and you can easily press merge button. However, we don't have such, so all the PR's merged manually with care and after review.

kxepal:
Also, I'm not sure if same rules applies to Fauxton ,but AFAIK you need to create PR from Cloudant fork and not use BugzID, but CouchDB JIRA issue number. Or both at least. Ask your team mates for the details, that could be important.

ykye:
@kxepal Thank you kxepal. I will ask my team mates later.

app/addons/config/routes.js
```diff
@@ -67,7 +67,7 @@ function (app, FauxtonAPI, Config, Views, CORSComponents, CORSActions, ClusterAc
       this.sidebar = this.setView('#sidebar-content', new Views.Tabs({
         sidebarItems: [
           {
-            title: 'Main config',
+            title: 'config.js File',
```
robertkowalski: config.js File?

app/addons/permissions/components.react.jsx
```diff
@@ -123,39 +123,73 @@ function (app, FauxtonAPI, React, Components, Stores, Actions) {
...
+    },
+
+    getRoleBlurb: function () {
+      var blurb = '';
+      if (this.props.section === 'admins') {
+        blurb = 'Grant database ADMIN privileges to user accounts with the following roles:';
+      } else {
+        blurb = 'Grant database member privileges to user accounts with the following roles:';
+      }
+      return (
```
robertkowalski: you can use omit the else case in both functions:

```js
var blurb = 'Grant database member privileges to user accounts with the following roles:';
if (this.props.section === 'admins') {
  blurb = 'Grant database ADMIN privileges to user accounts with the following roles:';
} 
```

app/addons/permissions/components.react.jsx
```diff
@@ -123,39 +123,73 @@ function (app, FauxtonAPI, React, Components, Stores, Actions) {
...
+            <p className="help">
+              {this.getHelp()}
+              <a className="help-link" data-bypass="true" href={getDocUrl('DB_PERMISSION')} target="_blank">
+                <i className="icon-question-sign"></i>
+              </a>
+            </p>
+          </header>
+       );
+      } else {
+        return '';
```
robertkowalski: you can use an early return instead. 

please also return `null` instead of an empty string

app/addons/auth/components.react.jsx
```diff
@@ -154,12 +153,12 @@ define([
             <input id="password" type="password" ref="password" name="password" placeholder="Password"
               size="24" onChange={this.onChangePassword} value={this.state.password} />
             <br />
-            <input id="password-confirm" type="password" name="password_confirm" placeholder= "Verify Password"
+            <input id="password-confirm" type="password" name="password_confirm" placeholder= "Password confirmation"
```
robertkowalski: `placeholder= "Password confirmation"`

one whitespace that does not belong there

app/addons/config/routes.js
ykye: image

robertkowalski:
looks good, i left a few comments

app/addons/config/routes.js
kxepal: Speaking from backend side, there are no Main config or 'config.js'. Users knowns that all configs are stored in INI files, so if you tell them about some config.js someone mind gonna blow.

"Main config" name isn't good here as well since as you can see, there is CORS menu item. Which means, that here we actually see some configuration groups, not some special configuration  From this point "General" instead of "Main config" would be more sane replacement. Or "All configuration" - also good, especially in mid-term perspective (imagine that there will be more groups like Auth, Compaction, Databases, Admins etc).

garrensmith: I think instead of `Main config` maybe something like General. Its actually all the config but CORS.

app/addons/auth/components.react.jsx
```diff
@@ -219,17 +218,17 @@ define([
...
+            Before an admin user account is created for this CouchDB server instance,
+            all clients have ADMIN privileges.  <strong>If clients are not trusted users,
+            you must create an admin account to prevent accidental or malicious data loss.</strong>
           </p>
           <p>
-            Server admins can create and destroy databases, install and update _design documents, run
-            the test suite, and edit all aspects of CouchDB configuration.
+            Users with ADMIN privileges can create, modify, and delete databases and design documents;
+            run the test suite; and modify the CouchDB configuration. Users without ADMIN privileges
+            have read/write access to all databases. CouchDB can be configured to block access by anonymous users.
```
michellephung: "CouchDB can be configured to block access by anonymous users." Can you rewrite this sentence?

kxepal: I think such sentence requires a link to somewhere to answer on upcoming question: "That's great! How?".

app/addons/auth/components.react.jsx
```diff
@@ -239,9 +238,6 @@ define([
             <input id="password" type="password" name="password" placeholder= "Password" size="24"
               onChange={this.onChangePassword} />
             <p>
-              Non-admin users have read and write access to all databases, which
-              are controlled by validation functions. CouchDB can be configured to block all
-              access to anonymous users.
```
michellephung: Hello. :) I didn't mean to delete the paragraph. I meant the new wording needs some improvements.

What was wrong with this original wording? I like it.

As Kxepal mentioned, it might be helpful to point the user to where/how they can protect their database(s). Perhaps a helpful link in a circle-question-mark-symbol, to the correct place in the docs?

ykye: I wanted to change the last sentence to "CouchDB can be configured to block the access of anonymous users." and add the link "http://docs.couchdb.org/en/1.6.1/config/auth.html" to it, but I found that it’s not necessarily related to ADMIN privileges, so I deleted the last sentence.

https://wiki.apache.org/couchdb/Security_Features_Overview

kxepal: Technically, to block anonymous users you need to fix admin party (aka create admin) and set [require_valid_user](http://docs.couchdb.org/en/1.6.1/config/auth.html#couch_httpd_auth/require_valid_user) as `true`, so everyone have to auth to send any request to CouchDB.

Humbedooh:
Abandoned issue. Closing here.

PR #576, opened 2015-11-16
fix cors validation for hosts with no tld

robertkowalski:
- allow ips and things like: http://localhost:3000
 - normalize urls in case a path is given by removing the path

note:

in general the validation should be done in the backend as single
source of truth so that api users and fauxton users get the same
results and also the interfaces are consistent.


benkeen:
Looks good. +1

robertkowalski:
merged as 8588ef7cc55bfa7a02664d6f6670275c986bb63a

PR #578, opened 2015-11-18
Cors timing

robertkowalski:
cors: use i18n system for changeable strings

cors: remove dead code

Fix timing issue for slow connections
If the conenction was slow you could enabble cors and change
options, but cors was not enabled yet in the backend


michellephung:
can we do this instead:

in components.react.jsx:
```
render: function () {
      var isVisible = _.all([this.state.corsEnabled, !this.state.isAllOrigins]);
      var className = this.state.corsEnabled ? 'collapsing-container' : '';
      var waiting = this.state.isLoading ? <ReactComponents.LoadLines /> : '';

      // if (this.state.isLoading) {
      //   return (<ReactComponents.LoadLines />);
      // }

      return (
        <div className="cors-page">
          <header id="cors-header">
            <p>{app.i18n.en_US['cors-notice']}</p>
          </header>

          <form id="corsForm" onSubmit={this.save}>
            <div className="cors-enable">
              <label className="checkbox">
                <input
                  type="checkbox"
                  checked={this.state.corsEnabled}
                  onChange={this.enableCorsChange} />
                  Enable CORS
              </label>
              
            </div>
            <div id="cors-waiting">{waiting}</div>
            <div id={className}>
              <Origins corsEnabled={this.state.corsEnabled} originChange={this.originChange} isAllOrigins={this.state.isAllOrigins}/>
              <OriginTable updateOrigin={this.updateOrigin} deleteOrigin={this.deleteOrigin} isVisible={isVisible} origins={this.state.origins} />
              <OriginInput addOrigin={this.addOrigin} isVisible={isVisible} />
            </div>

          </form>
        </div>
      );
    }
```

and in cors.less: 
```
#cors-waiting {
  position: absolute;
  left: 50%;
  top: 50%;
}
```

michellephung:
+1 if you want to merge though

robertkowalski:
no we can't as the checkbox must be also hidden by something

michellephung:
1. why do we have to hide the checkbox?

2. assume we HAVE to, can you still center the loading lines?

robertkowalski:
to avoid that people change the value while the previous request is still running

robertkowalski:
i'm not sure what you mean, they are centered i think:

bildschirmfoto 2015-11-19 um 15 16 58

michellephung:
i mean, center it vertically as well.

What happens if they want to cancel the request while waiting?
Can we disable the checkbox while it is loading?

PR #579, opened 2015-11-18
Allow db names starting with underscore

benkeen:
The JS validation unnecessarily stopped users from creating
database names with underscores. This removes that restriction.


rnewson:
but database names can't start with underscore, that's the reserved character, couchdb will prohibit it. Are you just trying to avoid duplicating the few that are allowed? (_users, _replicator).

kxepal:
May be remove this validation from Fauxton completely and let it make a decision based on response from CouchDB then?

robertkowalski:
@kxepal sounds like a good idea, +1 for a single source of truth

benkeen:
Yikes! This opened up a can of worms that I didn't anticipate. Agreed on the single-source-of-truth bit, but pity the user has to wait for a request response for something like this.

Going to close because this brooks a longer investigation. 

Thanks all.

PR #581, opened 2015-11-25
build: add json suffix to json files

robertkowalski:
adding a .json suffix to our json files enables us to use the
faster native `require` from node instead of grunts wrapper
around the file api with a json parser.

decoupling the reading from grunt enables us to slice out small
parts of our build without a hard dependency to the grunt
monolith.

notes:

 - throw instead of silently fail in case of a missing
   settings.json. if the settings.json is missing fauxton can't
   run, there is no reason to return a stub


michellephung:
+1 everything seems to work

robertkowalski:
merged as 8f84268cf9d8d16d13bc43068edc31483193b2a0

PR #582, opened 2015-11-25
build: simplify & fix

robertkowalski:
- remove git checkouts, fauxton never used git checkout and they
   add complexity
 - fix async code: an async function call returns undefined, fix
   done(local && remote)
 - fix async code: as async callbacks do not return move the
   done-callback to the point where all async operations are
   finished to avoid possible race conditions


michellephung:
+1 works on my machine

robertkowalski:
merged as 6d6dbdd366e1a42b1bd7298baf4ee4dd598dc2fd

PR #584, opened 2015-11-25
Pagination update

benkeen:
This refactors the Pagination React component to make it a bit more
generic. Namely:
- max visible pages that appear in pagination is now configurable
- optional callback function to be called when the user clicks
a pagination link, instead of just outputting an href on the <a>
tags

### DIFF
--- a/app/addons/fauxton/tests/componentsSpec.react.jsx
+++ b/app/addons/fauxton/tests/componentsSpec.react.jsx
@@ -202,6 +202,50 @@ define([
       assert.equal("30", lis[10].innerText);
     });
 
+    it('limits the number of total pages when customized', function () {
+      var maxNavPages = 15;
+      var pageEl = TestUtils.renderIntoDocument(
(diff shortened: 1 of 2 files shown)


app/addons/fauxton/tests/componentsSpec.react.jsx
```diff
@@ -202,6 +202,50 @@ define([
...
+      );
+      var links = React.findDOMNode(pageEl).getElementsByTagName("a");
+
+      TestUtils.Simulate.click(links[0]);
+      assert.ok(spy.calledWith(currentPage - 1));
+
+      TestUtils.Simulate.click(links[11]); // last index
+      assert.ok(spy.calledWith(currentPage + 1));
+    });
+
```
robertkowalski: i really like the avoidance of jquery here i am usually to lazy to do that in my tests

robertkowalski:
+1 when green

PR #585, opened 2015-11-26
Footer & Queryoptions update

robertkowalski:
three commits:

```
footer: text update
```

```
remove update seq

update_seq does not work with couchdb 2. the footer text also was
not visible because of a css issue for months.
```

```
remove broken showStale

this was broken in multiple ways (option not visible and if option
visible changing it made no effect)

remove the feature as it has 0 users.
```


app/addons/documents/pagination/pagination.react.jsx
```diff
@@ -139,16 +138,10 @@ define([
 
       pageNumber: function () {
         if (this.state.totalRows === 0) {
-          return <p> Showing 0 documents. </p>;
+          return <p>Showing 0 documents.</p>;
```
benkeen: Maybe drop the period?

app/addons/documents/queryoptions/actiontypes.js
```diff
@@ -17,7 +17,6 @@ define([], function () {
     QUERY_TOGGLE_BETWEEN_KEYS: 'QUERY_TOGGLE_BETWEEN_KEYS',
     QUERY_UPDATE_BETWEEN_KEYS: 'QUERY_UPDATE_BETWEEN_KEYS',
     QUERY_UPDATE_BY_KEYS: 'QUERY_UPDATE_BY_KEYS',
-    QUERY_TOGGLE_UPDATE_SEQ: 'QUERY_TOGGLE_UPDATE_SEQ',
     QUERY_TOGGLE_DESCENDING: 'QUERY_TOGGLE_DESCENDING',
     QUERY_TOGGLE_REDUCE: 'QUERY_TOGGLE_REDUCE',
     QUERY_UPDATE_SKIP: 'QUERY_UPDATE_SKIP',
```
benkeen: I think we can remove QUERY_SHOW_STALE from this file too.

benkeen:
Other than the minor things above, code looks good + works fine. I'm all for removing features that are simply not used. I wonder if people will notice that the stale option is no longer there and pipe up. 

But I think +1.

robertkowalski:
fixed and merged :)

PR #586, opened 2015-11-30
build: build less/css faster & without grunt

robertkowalski:
this slices our first build step out of the monolithic grunt build.

improvements:

 - faster less build
 - no useless css files for modules in debug folder, just one
   `index.css`
 - multiple targets possible (so we don't need to build from
   `dist/debug` in the future

performance for less build & concat:

```
old:
 - 762ms -- 757ms + 5ms (separate concat)

new:
 - 597ms
```

why is it faster?

1. previous buildstep wrote css files on the disk to read & concat
it later

2. grunt.readfile / grunt.readjson are slow compared to the native
require function which also supports caching

needs node 4.x like our travis builds


build-helper/less.js
```diff
@@ -0,0 +1,69 @@
+'use strict';
```
benkeen: This necessary? Just curious - I thought all parsing of the JS files was jshinted - or is that only the application code?

benkeen:
This looks great, Robert. Works fine. Nice to see some ES6 code in Fauxton. :+1: +1

build-helper/less.js
robertkowalski: yes, some ES6 features like const just work in strict mode right now

robertkowalski:
merged as a0c0b166d85d5e3db30342949290b33107e9304e - yay!

PR #587, opened 2015-12-02
cors: ui improvements

robertkowalski:
- display loading lines at the bottom
 - use a button for enabling / disabling

### DIFF
--- a/app/addons/cors/assets/less/cors.less
+++ b/app/addons/cors/assets/less/cors.less
@@ -14,6 +14,18 @@
 @import "../../../../../assets/less/variables.less";
 @import "../../../../../assets/less/bootstrap/mixins.less";
 
+.cors-page {
+  .loading-lines {
+    margin-top: 20px;
+  }
+  .enable-disable.btn {
+    margin-top: 10px;
+  }
+
+}
+
+
+
 #cors-header {
   margin: 18px 30px 0 30px;
 }
--- a/app/addons/cors/components.react.jsx
+++ b/app/addons/cors/components.react.jsx
@@ -178,8 +178,8 @@ define([
       }
 
       return (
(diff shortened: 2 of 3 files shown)


benkeen:
Looks good! Just one thing: could we give the LoadLines a 20px margin-top, so there's some space between it and the line separating the top? It looks a bit odd (and inconsistent with other pages) being flush with the line: 

screen shot 2015-12-02 at 3 42 01 pm

benkeen:
Other than that +1

kxepal:
I may be too late, but why Enable CORS is cyan? It looks alien for Fauxton color scheme. Everywhere else green buttons were used.

benkeen:
Really? I don't mind it myself. Red and green don't seem applicable for this, and we've used that colour in the compaction and notification panel (class `.btn-info`). But I'm totally game for a colour tweak - not sure what it should be, but I'd argue against red or green which imply error/success to me.

kxepal:
@benkeen Green buttons are used not only for success, but for general actions (see database security configuration) while red ones are just used on dark background (see configuration add option). So I don't think there is any rule.

As for notification panel, it has same alien design that fits more Cloudant style rather than CouchDB Fauxton one.

Anyway, I just get caught by your screenshot and thought it worth to point on this issue. I see there is no colour definition on this PR, so this topic is definitely not for here (:

benkeen:
Oh I don't know, it seems applicable! :) I'm definitely game for overriding it for something more Fauxton-y (just don't ask me to be the one to pick the colour...), but yeah - it can be handled in a separate ticket for sure.

robertkowalski:
agreed @kxepal 

changed the button to green and added margin.

good point about the notification center.

kxepal:
@robertkowalski thanks! (:

PR #589, opened 2015-12-04
views: remove non existent _id attribute from result

robertkowalski:
this removes an addionial _id attribute that was added, even if it
was not part of the response.

regression was introduced in december with #c7fa6fad7c411


garrensmith:

PR #595, opened 2016-01-04
Degruntify cssmin & uglify js. Separate testserver from building releases

robertkowalski:
I am cleaning up our build further, by removing more of the grunt tools that have no real need

This PR saves us 21K by changing/optimizing the compression tools we have. 

Additionally I made first progress in separating the des/debug folder, used by the testserver we spin up and a production release. the index.css and the compiled require.js files are not copied from the debugserver.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -178,16 +178,6 @@ module.exports = function (grunt) {
       }
     },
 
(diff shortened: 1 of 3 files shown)


benkeen:
+1

PR #596, opened 2016-01-06
fix: cssclean should not rebase font urls

robertkowalski:
this leads to fonts not being loaded


benkeen:
Nice. +1

robertkowalski:
merged as 7642cadb1e8d771307dc0b64ef10509ff440f5fe :)

PR #598, opened 2016-01-07
remove useless grunt defaults

robertkowalski:
the defaults are not used at all as all setting.json files
implement them, but make it harder for everyone to understand what
is happening.

additionally configuration should not be mixed into the
code as it turned out worthwile to don't mix config and code.


benkeen:
True, this chunk of code isn't being used, but I disagree it's useless. We always override it, true, but it looks like we don't need to for Fauxton, no? (i.e. we could rely on the development or production defaults to take care of providing those default values). But besides that, I kind of like being able to see at a glance exactly what info grunt it looking for to create the build.

Instead, in addition to dropping the code like you've done, could we add a validation check to ensure the required params are there: src, dest, and the required vars? That could output a nice juicy "such and such is missing" error to ensure grunt gets what it wants. I realize this isn't strictly needed now, but for people editing / creating new settings files it may be helpful.

robertkowalski:
>  We always override it, true, but it looks like we don't need to for Fauxton, no?

we have to override them for fauxton:

```
          "requirejs": "./js/{REQUIREJS_FILE}",
           "css": "./css/{CSS_FILE}",
```
is wrong, it will never work (the curly braces are not replaced)

> Instead, in addition to dropping the code like you've done, could we add a validation check to ensure the required params are there: src, dest, and the required vars?

the feature for checking the settings.json for required fields is a nice idea. i would like to keep that feature separated from the cleanup.

the idea is nice, but so far i am not aware that people had issues when creating those files. i personally would like to wait until it gets an issue for us. i am also sure it will get outdated very soon, as people add more things to the settings.json, but forget to update the check.

but i also don't want to block you, if you think it is something you should add right now, feel free to add it.

benkeen:
Cool, thanks Robert! It’s fine - it was just a niggle I had with the ticket. It's fine as it is, thanks for the explanation. +1

robertkowalski:
merged as 57883a2b4f97c82288e4c589857879eede4861ff

PR #599, opened 2016-01-08
use require.js on the cli

robertkowalski:
makes it composable, so we can have different targets


benkeen:
Looks good to me. No problems building or running. +1

robertkowalski:
merged as 5559ed2c2f26b5d7f8359f87146ff8d1bef603ae

PR #600, opened 2016-01-09
Sidebar component updates

benkeen:
This makes a couple of changes to the DesignDoc and IndexSection
components used in the sidebar.

1. `<IndexSection />` knew too much! It had some hardcoded
references to the structure (prop names) of the design docs and
how it was being called. Now it’s more generic and passed
everything it needs to know via props, so it’s a little bit
dumber + will work for any new type of design doc contents.

2. The design doc item in the sidebar will now only list
Search Indexes or other subsections if they have content. This
won’t impact Fauxton as is, but for scripts that wrap this
code, it’ll cut down on visual clutter in the UI. [This is
mostly for UX.]

### DIFF
--- a/app/addons/documents/sidebar/sidebar.react.jsx
+++ b/app/addons/documents/sidebar/sidebar.react.jsx
@@ -142,18 +142,18 @@ function (app, FauxtonAPI, React, Stores, Actions, Components, DocumentViews) {
 
   var IndexSection = React.createClass({
 
-    getDefaultProps: function () {
-      return {
-        indexTypeMap: {
-          views:   { icon: 'fonticon-sidenav-map-reduce', urlFolder: '_view', type: 'view' },
-          indexes: { icon: 'fonticon-sidenav-search', urlFolder: '_indexes', type: 'search' }
-        }
-      };
+    propTypes: {
+      urlNamespace: React.PropTypes.string.isRequired,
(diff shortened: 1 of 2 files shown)


app/addons/documents/sidebar/sidebar.react.jsx
```diff
@@ -205,19 +212,23 @@ function (app, FauxtonAPI, React, Stores, Actions, Components, DocumentViews) {
...
   var DesignDoc = React.createClass({
 
-    createIndexList: function () {
-      var sidebarListTypes = FauxtonAPI.getExtensions('sidebar:list');
-
-      if (_.isEmpty(sidebarListTypes) ||
-        (_.has(sidebarListTypes[0], 'selector') && sidebarListTypes[0].selector !== 'views')) {
-        sidebarListTypes.unshift({
+    componentWillMount: function () {
+      this.sidebarListTypes = FauxtonAPI.getExtensions('sidebar:list');
```
robertkowalski: can we pass the sidebarListTypes as a prop, too? this would also make testing easier as we would not have to register extensions in order to test the functionality of the sidebar.

benkeen: Good idea. Will do.

robertkowalski:
+1 great!

benkeen:
Merged as 8a7a0f9

PR #601, opened 2016-01-12
serve assets from dashboard.asset

robertkowalski:
- remove app.bundled workaround, dependencies are now served
transparantly for debug and production

- put all dependencies into dashboard.assets: this makes routing
a lot easier, as a dot is not allaowed in a database name for
couchdb. cloudant supports for a long time and i see no reason to
make routing for couchdb also easier


benkeen:
LGTM. Didn't spot any problems. +1

PR #602, opened 2016-01-13
do not depend on second alert.info

robertkowalski:
the two deletes make two notification boxes appear and disappear.

if the first one is removed in an unlucky timeframe, selenoum can't
find it any more with the current selector.

as the notification is not necessary for the test, remove it.


benkeen:
+1

robertkowalski:
merged as 54427aa2449e0229051fba48663ec04ffc928648

PR #603, opened 2016-01-13
fix production build

robertkowalski:
- prepend and append additional dependencies


benkeen:
Thanks for sticking your neck out on improving the build process @robertkowalski. Not the most enjoyable work, I know! +1

robertkowalski:
merged as 75b0b11da3fe63c0cdef223311072c72d5660b90

PR #607, opened 2016-01-15
footer updates

robertkowalski:
- add dot behind every sentence
 - add text "showing x columns if all colums are shown"
 - swap order between "show all" and amount of displayed elements
 - add border between sections
 - fix alignment of label/checkbox element

### DIFF
--- a/app/addons/documents/assets/less/documents.less
+++ b/app/addons/documents/assets/less/documents.less
@@ -171,13 +171,24 @@ button.string-edit[disabled] {
   .footer-table-control label {
     margin: 0;
     padding: 0;
-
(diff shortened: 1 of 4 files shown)


benkeen:
Code looks good. Really not a fan on the `.` periods though. Looks weird and we don't add periods to one-liner text strings like this anywhere else.

robertkowalski:
i don't have a very strong opinion regarding the periods. hmm...

benkeen:
I'll defer to Justin on that. +1

robertkowalski:
merged as 4627fea6adb59ed02d772e5964dca922514f7643

PR #609, opened 2016-01-15
transpile with babel

robertkowalski:
Hello ES6 and React 14! :sunny: 

Additional changes:

 - optimize watch task: run stylecheck before to provide faster
   feedback


Gruntfile.js
```diff
@@ -376,39 +383,26 @@ module.exports = function (grunt) {
...
+    var targetFolder = folder;
 
-    // if the JS file that just changed was outside of Fauxton, copy it over
-    if (isJS && !(/^app\/addons/.test(filepath))) {
-      config.copy.changedFiles.files = [{
-        src: filepath,
-        dest: filepath.replace(/.*\/addons/, 'app/addons')
-      }];
-      grunt.task.run(['copy:changedFiles']);
-    }
```
robertkowalski: i am not sure about the intention of the code code and it does not seem to make sense with all required js and compiled js files in the new dev server directory any more - can you confirm that @benkeen ?

benkeen: I'm not sure. The purpose of this code was for local dev to ensure any changed JS files in the dash get copied over immediately (i.e. as you change them). Won't this condition still apply with the new folder structure?

benkeen:
Super exciting. :D Can't wait to get React 14 in there.

From what I can see, this look good. It does break the dash, though ['cannot read property jQuery of  undefined` / `cannot set property Backbone` of undefined].

robertkowalski:
@benkeen i had to disable the ES6 transpiling as it seems to break with phantomjs... so far haven't found out the issue.

it looks like nothing breaks now, can you retest?

benkeen:
Looks good from here. There's a wayward console.log() in `routes-mango.js` but it looks like that snuck in  in another ticket. 

+1

garrensmith:
+1 looks great.

robertkowalski:
merged as cfc664d

PR #611, opened 2016-01-18
Minor CSS tweaks

benkeen:
After speaking with Justin, I made a small tweak to the
header CSS which ensures the spacing between the API and
OPTIONS header buttons and their icons was consistent
regardless of viewport size. That lead to a few other small
CSS tweaks to ensure it works on all screen sizes (e.g. 
the views page looked a bit off).


robertkowalski:
+1

benkeen:
Just rebased. Will merge when green.

benkeen:
Merged as 9e5cf9f

PR #613, opened 2016-01-19
fix fauxton server for module

robertkowalski:
- use native node methods instead of custom solutions

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -18,6 +18,8 @@
 /*jslint node: true */
 "use strict";
 
+const path = require('path');
+
 module.exports = function (grunt) {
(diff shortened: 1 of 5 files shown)


benkeen:
+1. 

Just a thought: might be worth noting in the doc somewhere that we need a recent(ish) version of Node. Couple of people have run into issues running a pre-ES6 version where it complains about `const`.

robertkowalski:
@benkeen i decided to create a executable documentation which will notice the users even if they don't read the docs or know where they are located.

benkeen:
Nice, love it!

robertkowalski:
merged as 16a81431a89f2dbc5886855013c0b2bcb3acff72

PR #614, opened 2016-01-19
fix red test

robertkowalski:
this fixes the red master


benkeen:
Curious - what changed with this code other than spacing & removing that customDocURL var? I'm sure this was green locally (it was on the branch prior to merging). 

Wonder if it's a load-order thing...

robertkowalski:
haven't investigated it further - test went suddenly green while working on it. looks like master is green now - weird. closing...

PR #616, opened 2016-01-19
always update api url

robertkowalski:
also update api url when routes are in the same route object.

steps to reproduce:

1. click on api bar in _find area for mango
2. click edit
3. click api bar button on _index page
4. url should point to _index


benkeen:
Bit scary - we have a load of legacy logic for updating the API Bar all over the place. But I couldn't find any regression issues, it fixes the issue you noted & seems very sensible. +1

robertkowalski:
merged as 0a03e2e1a600fcd6af27bcb96fcf0bf3bc643236

PR #617, opened 2016-01-19
React 14 update

benkeen:
- update React to the most recent version (`0.14.6`)
- compatibility update for React Bootstrap (latest version: `0.28.2`)
- various code updates to use ReactDOM instead of React.

N.B. I ran into an issue with testing React Bootstrap modal components. Testing them was problematic in the first place because they work by creating the modals at the top level of the DOM *outside* the wrapper component. As such, you need to locate the elements in a different fashion (see test code). This seems to have been broken with the latest updates to the lib / React itself. I spent a long time investigating a workaround but couldn't find a solution. I'll create this PR in the meantime & return to it when I can.


benkeen:
I opened up a ticket w.r.t. to the issue I'm having accessing the react bootstrap Modals. Hopefully it's a straightforward fix.
https://github.com/react-bootstrap/react-bootstrap/issues/1617

robertkowalski:
hey ben,

i would delete the tests "event methods called" -- they don't test our logic, they try to test the react bootstrap component.


leaves us just with the "onSave" tests

benkeen:
Cool, ok.

robertkowalski:
hi been i took anothe rlook it looks like the component is not rendered and that means it does not get a ref.

things like 
```
console.log("->", $(el).html());
console.log($(ReactDOM.findDOMNode(el)).html());
```

return undefined or empty dom nodes

hope that helps!

app/addons/components/tests/stringEditModalSpec.react.jsx
```diff
@@ -86,11 +88,13 @@ define([
         );
 
         var modal = TestUtils.findRenderedComponentWithType(el, Modal);
-        var modalEl = React.findDOMNode(modal.refs.modal);
+        var modalEl = ReactDOM.findDOMNode(modal.refs.modal);
 
         TestUtils.Simulate.click($(modalEl).find('#string-edit-save-btn')[0]);
         assert.ok(spy.calledOnce);
```
robertkowalski: this test makes no sense and has to get fixed

test title is: "replaces "\\n" with actual newlines"

the actual test asserts that a spy was called once, nothing more.

app/addons/components/tests/stringEditModalSpec.react.jsx
```diff
@@ -70,7 +72,7 @@ define([
           container
         );
         var modal = TestUtils.findRenderedComponentWithType(el, Modal);
-        var modalEl = React.findDOMNode(modal.refs.modal);
+        var modalEl = ReactDOM.findDOMNode(modal.refs.modal);
 
         TestUtils.Simulate.click($(modalEl).find('#string-edit-save-btn')[0]);
```
robertkowalski: the only way how i found th button is:

```js
console.log($('body').find('#string-edit-save-btn'));
```

benkeen: Thanks Robert, yeah I'm going to use this technique instead. It's just a test: doesn't have to be beautiful, as long as it's tested!

benkeen:
Alright! This should finally be green, @robertkowalski. :)

I moved the newline test code into a separate helper. Ace wouldn't run in mocha environment, so this allows it to be tested in isolation. Beyond that it now uses simple $('body') tests to locate the react modals, and I patched up a few other minor things here and there.

app/addons/documents/doc-editor/tests/doc-editor.componentsSpec.react.jsx
```diff
@@ -146,13 +147,13 @@ define([
...
 
-      var attachmentNode = $(el.getDOMNode()).find('.view-attachments-section .dropdown-menu li')[0];
+      var attachmentNode = $(ReactDOM.findDOMNode(el)).find('.view-attachments-section .dropdown-menu li')[0];
       var attachmentURLactual = $(attachmentNode).find('a').attr('href');
 
       assert.equal(attachmentURLactual, "../../id/_design/test-doc/one.png");
     });
 
-
+/*
```
robertkowalski: this should get uncommented or deleted

benkeen:
Good catch, @robertkowalski . I uncommented and fixed those modal tests in a rather kludgy manner, but hey! Should be green again. :)

robertkowalski:
+1 great work!

PR #625, opened 2016-01-22
remove warnings on console

robertkowalski:
the warnings are displayed per default even for a valid release.

the warnings confuse people as they things something is broken,
but it is not.

this leads to a mindset were warnings are generally ignored as
the warning don't say anything about the status of the
application. this patch removes the warnings as debug logging can
get inserted again when there is something to debug.

more background info in this 5 minute lightning talk:  https://vimeo.com/146218883  (first talk)


benkeen:
I agree. +1

robertkowalski:
merged as f6d85b4a4675201d613feab8d66271c5bedce8e1

PR #628, opened 2016-01-26
don't hide api-bar

robertkowalski:
the tabs where hiding the api bar in case it was opened


benkeen:
Where can I see this bug?

benkeen:
Ah, FF! +1 works great.

robertkowalski:
merged as 31f9fb1638816c6281928b6971780b6cc85ffa5d

PR #632, opened 2016-01-27
make dbs deletable from db list

robertkowalski:
offer a modal for comfortable deletion from the database list.

additionally factor out `isSystemDatabase` and test it.


justin-mcdavid-ibm:
I recommend increasing spacing between the buttons (10px would be better).

Change the text for the warning modal to:

Delete Database

Warning: This action will permanently delete the <name> database.

To confirm the deletion of the database and all of the database's documents, you must enter the database's name.


Otherwise, looks good.

benkeen:
LOVE this improvement.

Small bug though. To reproduce:
1. delete a database via the new modal on the Databases page.
2. Create a new database via the Create Database header item, then click the `<` icon to return to the database list.

Result: you see the delete database modal again when you shouldn't.

Looks like the visibility state for the modal just isn't being cleared or something.

robertkowalski:
thanks, good catch!

it should be fixed now @benkeen

benkeen:
Great stuff! I having checked shared DBs, but I assume it's not added there since the feature wouldn't be applicable. +1

benkeen:
I'm just doing some stuff for Justin that involved moving the Compact + Clean options out of the database sidebar and into the Actions column on the databases page. Since that functionality isn't needed everywhere, the cleanest (and by far the simplest) solution is probably to just hide these buttons depending on the context with CSS. Otherwise we'd need to load the contents of the Actions column via extensions, which gets fussy fast.

Just a thought. You could do the same on shared DBs to omit the Delete Database button on that page.

robertkowalski:
merged as 63796145e1a9abac8b5d7ace8e721b379d9e6431 - great!

PR #633, opened 2016-01-30
Font icon fixes

benkeen:
This PR mops up a few remaining issues with the font icons. 
- After this ticket is merged there will now be *ZERO* regression with updating the fonts. They can be added safely and in isolation without fear of aggressive browser caching. 
- couple of new icons added

More details added to the individual commit messages.


robertkowalski:
+1

PR #638, opened 2016-02-02
fix flaky test / race condition

robertkowalski:
we init with an empty array, but the code expects a backbone
collection. if the connection is slow, we run into race conditions
as the store was not updated yet with a valid backbone collection.


benkeen:
+1

robertkowalski:
merged as 478a3b3 - yay!

PR #639, opened 2016-02-03
fix warning

robertkowalski:
reason is another race condition between the store's init and the
FauxtonAPI.url helper


benkeen:
+1

robertkowalski:
merged :)

PR #640, opened 2016-02-04
react-dom-server@0.14.7

benkeen:
React 14 moved the `renderToStaticMarkup` to react-dom-server.
This is still a helpful method on the client, like for example,
funnelling markup into React Bootstrap components which require
markup as properties on DOM elements.


robertkowalski:
+1

PR #641, opened 2016-02-04
Webpack

garrensmith:
This replaces require.js with webpack. This gives us a lot more flexibility in terms of creating builds. It also allows us to use npm to manage the majority of our packages. 
With Webpack the build is now at 1.97 MB, that includes the ace editor. Which is a definite improvement.
The dev server is now using `webpack-dev-server` which manages the proxy, reloading and everything else. And its pretty fast.

To test:
* npm install
* npm run dev for dev mode


benkeen:
Cool. I know this is early days, but just FYI - Installation + running went fine (very fast!), but I get the following errors when loading it up in my browser:
screen shot 2016-02-23 at 4 01 22 pm

.gitignore
```diff
@@ -36,3 +36,5 @@ test/nightwatch_tests/selenium/*
 *.react.js
 selenium-debug.log
 npm-debug.log
+test/bundle.js
+test/templates.js
```
robertkowalski: the both belong into the `dist/debug` folder

garrensmith: This is a tricky one. If we move them to dist then we cannot have the dev server running and run the tests at the same time.

benkeen:
This is great, Garren. Let's merge this sucker. +1

PR #642, opened 2016-02-09
trays: use higher order components to connect them to stores

robertkowalski:
this is the first step to get rid of `<Tray />` which uses
`FauxtonAPI.Events` and additionally  manages its own state by
accessing the DOM which leads to subtle bugs.

we use composition/higher order components as all trays share the
same wrapper, but have different stores connected.

 - change test urls to example.com, which is reserved for testing
 - use React.findDOMNode
 - trays don't touch the DOM directly any more

COUCHDB-2943


app/addons/components/actions.js
```diff
@@ -16,30 +16,33 @@ define([
 ],
 function (FauxtonAPI, ActionTypes) {
 
-  function showAPIBar () {
-    FauxtonAPI.dispatch({ type: ActionTypes.SHOW_API_BAR });
+  function showAPIBarButton () {
```
benkeen: Nice.

app/addons/components/react-components.react.jsx
```diff
@@ -1203,6 +1205,183 @@ function (app, FauxtonAPI, React, ReactDOM, Stores, FauxtonComponents, Helpers,
...
+        stores.forEach(function (store) {
+          store.off('change', this.onChange);
+        }.bind(this));
+      },
+
+      getInitialState: function () {
+        return getStateFromStores(this.props);
+      },
+
+      onChange: function () {
```
benkeen: connectToStores is nice and clean, very cool. 

Do we need an `this.isMounted()` check here?

benkeen:
Other than that one isMounted check, this looks good. +1

app/addons/components/react-components.react.jsx
robertkowalski: good call!

robertkowalski:
@benkeen fixed and rebased

benkeen:
Sweet. +1

robertkowalski:
merged as 40d956f0787087f47332e11e40f4b65cc9d0ce4a

PR #643, opened 2016-02-09
Design tweaks

robertkowalski:


app/addons/documents/queryoptions/queryoptions.react.jsx
```diff
@@ -173,8 +173,8 @@ function (app, FauxtonAPI, React, ReactDOM, Stores, Actions, Components) {
         <div className="query-group" id="query-options-key-search">
           <div className="add-on">Keys</div>
           <div className="btn-group toggle-btns row-fluid">
-            <label id="byKeys" onClick={this.toggleByKeys} className={byKeysButtonClass}>By Key(s)</label>
-            <label id="betweenKeys" onClick={this.toggleBetweenKeys} className={betweenKeysButtonClass}>Between Keys</label>
+            <label style={{width: '101px'}} id="byKeys" onClick={this.toggleByKeys} className={byKeysButtonClass}>By Key(s)</label>
+            <label style={{width: '101px'}} id="betweenKeys" onClick={this.toggleBetweenKeys} className={betweenKeysButtonClass}>Between Keys</label>
```
robertkowalski: i like the use of react-inline-styles here as this is style which is specific to this item

benkeen: However, there are a lot of other styles particular to this component that aren't inline. For consistency, if they're in both places I'd prefer keeping it in just the one. Just my preference - I don't object to this.

benkeen:
Looks good, +1 when green.

PR #644, opened 2016-02-09
Misc database sidebar updates

benkeen:
This contains various updates to the database sidebar, as per
Justin's feedback.
- Order and labels of buttons are changed.
- a few unclear icons (map reduce, Views) have been removed
- properly highlights what page you're on.
- Now allows us to select any nav item in the sidebar
regardless of nesting to highlight the nav item. This also
now works on full page refreshes.

Note this feature paves the way for the next thing I'll work
on: updating Views so that they retain the sidebar. Once
that's changed, it will need to clearly highlight what
View you're on in the sidebar.


app/addons/documents/assets/less/sidenav.less
```diff
@@ -16,176 +16,142 @@
...
+      width: @sidebarWidth - 30px;
+      display: block;
+      padding: 10px 13px 10px 36px;
+      text-overflow: ellipsis;
+      overflow: hidden;
     }
   }
 }
+
+.index-group-header {
```
robertkowalski: maybe namespace it somehow, e.g by nesting and/or naming?

app/addons/documents/sidebar/sidebar.react.jsx
```diff
@@ -380,24 +403,28 @@ function (app, FauxtonAPI, React, ReactDOM, Stores, Actions, Components, Documen
     },
 
     onChange: function () {
-      this.setState(this.getStoreState());
+      if (this.isMounted()) {
```
robertkowalski: cool, we should add those conditionals more often

app/addons/documents/sidebar/stores.js
```diff
@@ -42,62 +54,89 @@ function (app, FauxtonAPI, ActionTypes) {
...
+    getToggledSections: function () {
+      return this._toggledSections;
+    },
 
+    getDatabaseName: function () {
+      if (this.isLoading()) {
+        return '';
+      }
       return this._database.safeID();
     },
```
robertkowalski: maybe make the model have an property called 

`this._database.safeId` which is a string, and until it is populated an empty string

or a variation:

`this._database.name`

don't like the fact that we pass the model to the view and the view calls getter on the model, if the view thinks it has finished loading.

robertkowalski:
<3 the CSS refactor & also the new design.

the pink feels a bit strange, maybe Justin can give us a hint, he usually has really good ideas for colors

apart from the namespacing thing, +1

app/addons/documents/assets/less/sidenav.less
benkeen: Fair point! But I'd argue it's okay, actually. It's a pretty descriptive (specific) class name and it's the namespacing that ends up getting us in trouble with deep selectors + difficulty overriding them later. I did a search over the code base to ensure uniqueness (I always do when adding new selectors).

benkeen:
Yeah, let me run the colour by Justin one more time. I tweaked the pink a little to make it less awful but it could still use some love.

benkeen:
Thanks @robertkowalski. Justin had a number of other small changes which I'll add shortly.

app/addons/documents/assets/less/sidenav.less
benkeen: EDIT: I've namespaced this in .sidenav, just to get this ticket merged.

PR #645, opened 2016-02-09
Move compact-clean

benkeen:
This PR moves the Cleanup Views functionality to the databases page's Actions column, and removes the non-functionality Compaction functionality altogether.

Was unable to determine the reason for not being able to access the React components + Actions in the componentsSpec test.


benkeen:
Closing. After speaking with @robertkowalski, we're just going to remove the compaction addon altogether since it no longer works.

PR #646, opened 2016-02-12
Remove compaction addon

benkeen:
The compaction and view cleanup functionality stopped no longer
works within a cluster setup. This removes the old non-functional
addon.


robertkowalski:
+1

benkeen:
Merged as a62af89

PR #648, opened 2016-02-16
No Docs Found screen style update

benkeen:
Small aesthetic thing. The "no documents found" page had a
hardcoded minimum height which often caused the panel to
have the text hidden offscreen and a scrollbar to appear.

This just vertically centers the watermark and label so it works
on any screen size.

Also now exports the NoResultsScreen component for use elsewhere.

screenshot

### DIFF
--- a/app/addons/documents/assets/less/index-results.less
+++ b/app/addons/documents/assets/less/index-results.less
@@ -28,17 +28,28 @@
   }
 }
 
+.no-results-screen {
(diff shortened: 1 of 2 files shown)


robertkowalski:
+1

benkeen:
Thanks, @robertkowalski!

benkeen:
Merged as 5bc8bcd

PR #649, opened 2016-02-17
Truncate large docs in JSON view

benkeen:
Currently there's no restriction on the size of documents that
we attempt to display in JSON view. If the user uses
include_docs=true it can actually crash people's browsers if the
documents are too large.

This PR adds in a truncation option to the Document component,
defaulting to enabled, and 500 lines. It displays a simple
"truncated" note at the bottom of the result.

Example (with less than 500 rows, just for illustration purposes): 
screen shot 2016-02-16 at 4 04 49 pm

### DIFF
--- a/app/addons/components/react-components.react.jsx
+++ b/app/addons/components/react-components.react.jsx
@@ -17,7 +17,6 @@ define([
   'react-dom',
   'addons/components/actions',
   'addons/components/stores',
-
   'addons/fauxton/components.react',
   'addons/documents/helpers',
   'ace/ace',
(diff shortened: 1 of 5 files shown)


app/addons/components/react-components.react.jsx
```diff
@@ -1046,13 +1055,28 @@ function (app, FauxtonAPI, React, ReactDOM, Stores, Actions,
...
+      // if need be, truncate the document
+      var content = this.props.docContent;
+      var isTruncated = false;
+      if (this.props.truncate) {
+        var lines = this.props.docContent.split('\n');
+        if (lines.length > this.props.maxRows) {
+          isTruncated = true;
+          lines = lines.slice(0, this.props.maxRows);
+          content = lines.join('\n');
+        }
```
robertkowalski: can we move the logic to the store / data model?

robertkowalski:
wow, that is a super cool idea! super cool!

app/addons/components/react-components.react.jsx
benkeen: Yeah, this feels like the wrong place, hey. Almost feels like a helper function. I'll move it.

benkeen:
Thanks, Robert! I moved the truncation and added some tests.

robertkowalski:
+1

PR #650, opened 2016-02-19
Make DesignDocSelector into a dumb component

benkeen:
The dumbifies the DesignDocSelector to remove all ties a to
store and passes everything via props so we can get more use out
of it. Specifically, I need this for my next PR which includes
a new, generic Clone Index modal that'll use this component
in different contexts (i.e. all index types).

Also includes validation in the component to make it more
self-contained and allows for better UX by focusing on error fields.


benkeen:
Hey @robertkowalski, there's a failing test on this one, but it may have uncovered a related problem so I wanted to confirm the behaviour before fixing it. 

Here's the scenario (do this on master): 
1. editing a view and select "_sum" as the optional reduce option.
2. result: it shows a single doc (the summed doc).

So far, this seems correct to me. 

3. now click back to all docs, then click on the View that you just added `_sum` to.
4. That page then shows all docs in the view, not a single _sum result. 

Is this behaviour correct? Seems strange to me... shouldn't that View now *always* just show the single document when you've selected _sum for a View, like the test is confirming?

robertkowalski:
> Is this behaviour correct? Seems strange to me... shouldn't that View now always just show the single document when you've selected _sum for a View, like the test is confirming?

yes that sounds like a bug

benkeen:
Great, thanks Robert. I'll patch that up with this ticket and write a test for it.

benkeen:
Okay, I pared back this ticket a bit. I was a little overambitious in my refactoring. The next PR [https://github.com/apache/couchdb-fauxton/pull/651] moves the Views to their own page and will remove the need for some of the logic to dynamically reload the result list.

app/addons/documents/index-editor/components.react.jsx
```diff
@@ -326,10 +351,12 @@ function (app, FauxtonAPI, React, Stores, Actions, Components, ReactComponents)
         return;
       }
 
+      var doc = this.refs.designDocSelector.getDesignDoc();
```
robertkowalski: hmmm, not feeling comfortable by calling a getter on a view to get data. we used to do that with backbone a lot and it got us really into trouble.

maybe we can put that into a store and let the component fire an action? the dumb component does not need to be connected directly to the store, it could get the data updates as props from the parent component

benkeen: Blerg. I agree. Let me see what I can do.

benkeen:
Here we go. I slimmed down the component further to remove the getter and the Backbone props (Document + design doc collection).

robertkowalski:
+1 after using this.state

app/addons/documents/index-editor/components.react.jsx
```diff
@@ -330,6 +331,7 @@ function (app, FauxtonAPI, React, Stores, Actions, Components, ReactComponents)
         database: this.state.database,
         newView: this.state.isNewView,
         viewName: this.state.viewName,
+        designDoc: indexEditorStore.getSaveDesignDoc(),
```
robertkowalski: `this.state.saveDesignDoc`

saves performance & is in general nicer :)

benkeen:
Nice, thanks Robert! Yeah I stuck the store call at that location because technically it was only needed at that point: and calling it *always* would call it before the data was actually ready. Not a big deal at all, I can just add in a clause to clear it up, and I agree it's cleaner to put them all in state.

PR #654, opened 2016-02-23
Trays: finish softmigration

robertkowalski:
Fix bug where the query panel stayed open after multiple clicks:

When opening the Options Tray in the all_docs view, clicking query
and then cancel, clicking Cancel in the Query Options panel
doesn't close it.

uses flux for state management of the tray.

closes COUCHDB-2943


app/addons/documents/queryoptions/queryoptions.react.jsx
```diff
@@ -292,93 +294,126 @@ A key value is the first parameter emitted in a map function. For example emit("
...
+
+    propTypes: {
+      contentVisible: React.PropTypes.bool.isRequired
+    },
+
+    runQuery: function (e) {
+      e.preventDefault();
+
+      Actions.runQuery(this.props.queryParams);
+      Actions.toggleQueryBarVisibility(!this.props.contentVisible);
```
benkeen: Maybe just call `this.toggleTrayVisibility()` instead? Bit nicer than having the same Actions call twice.

app/addons/documents/queryoptions/queryoptions.react.jsx
```diff
@@ -292,93 +294,126 @@ A key value is the first parameter emitted in a map function. For example emit("
...
+    },
+
+    toggleTrayVisibility: function () {
+      Actions.toggleQueryBarVisibility(!this.props.contentVisible);
     },
 
     componentDidMount: function () {
-      store.on('change', this.onChange, this);
+      $('body').on('click.QueryTray', function (e) {
+        if ($(e.target).parents('#query-options').length) {
```
benkeen: `.closest()` is a little more performant (just finds the first parent that matches the selector, rather than all.

benkeen:
Works great; code looks great too! +1

app/addons/documents/queryoptions/queryoptions.react.jsx
robertkowalski: cool thank you

robertkowalski:
merged as b7c7865865a402bad836bc1f116184ad25b98a6a

PR #657, opened 2016-03-01
react-migration: replace delete-db-modal with react version

robertkowalski:
1. replaces old backbone delete-db-modal with backbone version
  2. renames `showModal` to `modalProps` for easier understanding
  3. removes softmigration paths for delete-db-modal


app/addons/components/react-components.react.jsx
```diff
@@ -1440,35 +1430,39 @@ function (app, FauxtonAPI, React, ReactDOM, Actions, Stores, FauxtonComponents,
...
+      var val = encodeURIComponent(e.target.value.trim());
+
       this.setState({
-        inputValue: e.target.value
+        inputValue: val
       });
 
       this.setState({
-        disableSubmit: e.target.value.trim() !== this.props.showModal.dbId.trim()
+        disableSubmit: val !== this.props.modalProps.dbId.trim()
```
benkeen: Tempting to put the `this.props.modalProps.dbId.trim()` into a `getDbId()` helper since it's done several times.

benkeen:
+1 looks good & works great.

app/addons/components/react-components.react.jsx
robertkowalski: good call, changed

robertkowalski:
merged as 98a546245eb5ddb25d2f7649c37baab7c8de6f61 and 0c07ddc432c12895310592b0c46729feaac958db

PR #659, opened 2016-03-02
[WIP] Clipboard.js

benkeen:
This PR was to switch out ZeroClipboard for Clipboard.js, but I ran into a problem with elegantly downgrading for Safari in all instances. The only way to gracefully degrade for Safari is to select the text in the page and alert a message saying "Press ⌘-C to copy to your clipboard". But this won't work for cases where the clipboard text isn't visible in the page.

I don't terribly want the code to get wasted; I'll post it here. Once clipboard.js supports Safari, this can be applied.


garrensmith:
@benkeen we have implemented clipboard.js, this PR helped. Could you close this PR

PR #661, opened 2016-03-08
Minor text/btn changes

benkeen:
This makes a few further additions to buttons / text / styles,
as per Justin's recommendations.

### DIFF
--- a/app/addons/documents/doc-editor/components.react.jsx
+++ b/app/addons/documents/doc-editor/components.react.jsx
@@ -143,12 +143,14 @@ define([
     },
 
     render: function () {
+      var saveButtonLabel = (this.props.isNewDoc) ? 'Create Document' : 'Save Changes';
+
       return (
         <div>
           <div id="doc-editor-actions-panel">
             <div className="doc-actions-left">
               <button className="save-doc btn btn-success save" type="button" onClick={this.saveDoc}>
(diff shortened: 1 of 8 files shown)


app/addons/fauxton/components.react.jsx
```diff
@@ -377,6 +378,8 @@ function (app, FauxtonAPI, React, ReactDOM, ZeroClipboard, ReactBootstrap) {
     },
 
     render: function () {
+      var btnClasses = 'btn js-btn-success ' + this.props.buttonClass;
```
robertkowalski: `js-btn-success` i think this class is unused

app/addons/documents/doc-editor/components.react.jsx
```diff
@@ -143,12 +143,14 @@ define([
     },
 
     render: function () {
+      var saveButtonLabel = (this.props.isNewDoc) ? 'Create Document' : 'Save Changes';
```
robertkowalski: minor: no parens needed, but feel free to leave it like it is, no hard opinions here from my side

robertkowalski:
good work! +1 after removing the dead class

benkeen:
Merged as bd7205f45c2c2448da7e8e06f743f2d0444177f9

PR #662, opened 2016-03-08
Revert "JS validation added for Add Database"

robertkowalski:
This reverts commit b8bbb87c6919469521ff277694fb75e898157a7c.

The commit was added in https://github.com/apache/couchdb-fauxton/pull/468
but the validation has to happen in the backend. The current
validation does not cover databases which are created via curl or
other apps. It also prevents users from creating an accidentally
deleted `_users` database.

The fix for the root cause has to happen in the backend.

We should avoid trying to add validations in multiple, different
places.

COUCHDB-2748

### DIFF
--- a/app/addons/databases/actions.js
+++ b/app/addons/databases/actions.js
@@ -81,9 +81,9 @@ function (app, FauxtonAPI, Stores, ActionTypes, Resources) {
     },
 
     createNewDatabase: function (databaseName) {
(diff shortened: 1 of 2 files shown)


benkeen:
C'est la vie! +1

kxepal:
+1

robertkowalski:
merged as b182c03e8969ea553f16e9afbe328e981e86130e

PR #663, opened 2016-03-08
Style update for metadata page

benkeen:
The metadata page was kinda ugly, and Tim White recently made
a few suggestions to me on how to improve the page. This implements 
many of his ideas and generally smartens up the page a bit.

#### Old page
metadata-page-old

#### New page
metadata-page-new

### DIFF
--- a/app/addons/documents/assets/less/documents.less
+++ b/app/addons/documents/assets/less/documents.less
@@ -88,7 +88,39 @@ button.beautify {
 }
 
 .metadata-page {
+  font-size: 14px;
   padding: 20px;
+
+  .preheading {
+    font-size: 14px;
+  }
+  h2 {
+    margin: 5px 0 15px;
+  }
+  h3 {
+    font-size: 22px;
+    margin-top: 12px;
+  }
+  header {
+    border-bottom: 1px solid #cccccc;
+    padding-bottom: 10px;
+  }
+  .icon-question-sign {
+    margin-left: 4px;
+  }
+  section {
+    line-height: 21px;
+    width: 100%;
+
+    ul {
+      list-style-type: none;
+      margin: 0 0 15px;
+      padding: 0;
+    }
+    .item-title {
+      margin-right: 6px;
+    }
+  }
 }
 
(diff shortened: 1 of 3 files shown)


justin-mcdavid-ibm:
How about something like this (and hear me out)?

- Detailed field labels will help in understanding the metadata.
- The sub-headings don't currently provide much logical organization.
- Consistent justification (and units) should help in understanding values at a glance. And there shouldn't be massive orders of difference between actual data sizes and size on disk.
- The signature isn't going to mean anything on visual glance, so we can stick it under the copy-to-clipboard action.

metadata

benkeen:
Cool!
- I don't mind dropping the headings at all. Not sure they added much. But the gaps between the sections seem odd now - could we remove them? It's all just info, anyway.
- wonder if `language: javascript` is useful? Is a doc ever not in JS? (cc @robertkowalski)
- I miss the horizontal rule. The page just looks a lot smarter with it, I think. We use horizontal rules elsewhere.
- The copy-to-clipboard signature is very clever :) I like it. 
- I personally don't like the right alignment of the second col. I don't find it more readable and to me it feels out of place somehow (do we do this anywhere else in the application at all?) But if you feel strongly on it, I'm game.

justin-mcdavid-ibm:
I feel strongly about consistent alignment  (and wish it was used throughout our application, too). :)

I'd like to see spacing used to separate different types of values.  If you remove the one between purge sequence and data, I'd on board, but I think the other spaces allow you to visually group certain data types.

I accidentally covered the horizontal rule, and am completely fine with it coming back.

Thanks much, Ben.

garrensmith:
I'm getting this error

`bundle.js:135902 Uncaught TypeError: Cannot read property 'toLocaleString' of undefined`

benkeen:
Curious. Couldn't reproduce, but I made the code a bit more defensive.

garrensmith:
+1 working for me now.

benkeen:
Thanks @garrensmith. Merged as f6a1f82

PR #664, opened 2016-03-09
QueryButtons component now passed close handler

benkeen:
Needed to be able to reuse the component elsewhere.

### DIFF
--- a/app/addons/documents/queryoptions/queryoptions.react.jsx
+++ b/app/addons/documents/queryoptions/queryoptions.react.jsx
@@ -274,8 +274,12 @@ A key value is the first parameter emitted in a map function. For example emit("
   });
 
   var QueryButtons = React.createClass({
+    propTypes: {
+      onCancel: React.PropTypes.func.isRequired
+    },
+
     hideTray: function (e) {
-      Actions.toggleQueryBarVisibility(false);
+      this.props.onCancel();
     },
 
     render: function () {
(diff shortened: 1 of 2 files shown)


app/addons/documents/queryoptions/queryoptions.react.jsx
```diff
@@ -343,6 +347,10 @@ A key value is the first parameter emitted in a map function. For example emit("
       Actions.toggleQueryBarVisibility(!this.props.contentVisible);
     },
 
+    closeTray: function () {
+      Actions.toggleQueryBarVisibility(false);
```
robertkowalski: question: there are a few other occurences, like https://github.com/apache/couchdb-fauxton/pull/664/files#diff-a9b641a64002ae88871f5b368a99f1beL351 -- i think we have to replace them, too in order to make the callbacks called uniform?

benkeen: Those are actually okay. The other occurrences of that method are specific to the standard query tray used in Fauxton, so that's fine: they can keep their own event handler. 

This ticket's just so I can re-use the QueryButtons component elsewhere.

millayr:
LGTM

benkeen:
Thanks, Ryan. :)

benkeen:
Merged as aa52e40d432ecaeeaf04e31ed2597efd6d51078c

PR #665, opened 2016-03-09
Handle empty view attribute and does not set state for unmounted reduce

millayr:
See https://cloudant.fogbugz.com/f/cases/58137/Cannot-add-a-view-to-an-existing-design-doc-with-no-views-property

This should handle instances where users create a ddoc outside of the dashboard editor and fail to include a "views" attribute.


garrensmith:

PR #668, opened 2016-03-15
Notifications update

benkeen:
This updates the global notifications to use the same store as the notification panel and drop Backbone.


benkeen:
The license has been added to Couch here: https://github.com/apache/couchdb/pull/392

app/templates.js
```diff
@@ -0,0 +1,623 @@
+this["JST"] = this["JST"] || {};
```
garrensmith: This file shouldn't be here. you will have to do a `git rm`

garrensmith:
This looks really cool, so awesome to see another piece of backbone disappear. Two things

* The animation looks a little jagged when a notification appears.
* The one test is failing - I think its worth just restarting the tests to see if its a flaky test.

Otherwise +1

app/templates.js
benkeen: Ah ok, thanks.

benkeen:
@garrensmith - worrying about the animation looking jagged. Did it always look jagged for you, or just now and then? You're on Chrome/FF, I guess.

garrensmith:
I'm on chrome. It always looked jagged.

garrensmith:
+1

PR #669, opened 2016-03-16
Replication page update

benkeen:
This PR refactors the replication page for a simpler wizard-like structure. Advanced replication options are coming down the pipe in a different PR.

Replication scenarios now working:

```
local -> new local db 
local -> existing local db
local -> new remote db
local -> existing remote db
remote -> remote new DB
remote -> remote existing DB
remote -> new local db
remote -> existing local db
```

Yaaaaaaaayyyyyyyyy.


benkeen:
TODO - notes from Justin:

- In-field explanatory text should be gray +1
- Text on the replication button should be “Start Replication” +1
- The in-field explanatory text for Remote Sources and Targets should be https://$REMOTE_USERNAME:$REMOTE_PASSWORD@$REMOTE_SERVER/$DATABASE - 

[discussion: oooffff... that's awful verbose. I think REMOTE_* is implicit, no? I mean you already picked "remote source" / "remote target". Don't want to bludgeon them over the head with it. Oddly clarity becomes less clear here.
I'd be worried about some mistakenly thinking they needed to enter the current account's username and password. Maybe just drop the REMOTE_ on SERVER?]

- Increase the widths of all dropdowns. “Existing remote database” gets cut-off after selection (in FF, but not Chrome, I see). +1
- The replicate button activates after fields have been clicked on, but not before information is actually entered. Can we have the replication button activate only after the fields are populated? +1
-  There doesn’t appear to be any checking for the pre-existence of a “new” database. I selected “new local database,” and used the same DB for source and target.  Can we put in check in to make sure that users aren’t accidentally overwriting existing DBs? I’m assuming that there are back-end safeguards, but my test-replication job didn’t throw any errors. +1
- Can we leave the replication button inactive if the DB listed as an existing local source doesn’t actually exist?  I can enter an db that doesn’t exist and still trigger replication. Same for non-existing local targets. +1

Other:
- A source database that is open to/readable by the public does not require $USERNAME and $PASSWORD in its URL, right? Or do they all now require authentication (I remember something about this in your chats with Bob)? - Not sure... this bit's hard. :) In my trial and error I found you usually had to enter it, but I'll confirm.

justin-mcdavid-ibm:
Hey Ben,

I'm testing this, and it's awesome. Great work on this.  It's a big improvement.

There are a couple of things I'm seeing that we probably should tweak.

The "Select source" and "Select target" in-field explanatory test should be gray, though values selected through those dropdowns should remain the color of regular text.

When you specify a custom replication doc ID, it has to be a new ID (folks might think they can specify an existing doc-- which causes a conflict), so we should elaborate on the explanatory text as [Custom, new ID (optional)]. Also, because you can't use the same custom ID for multiple jobs, can we add a clear (x) inside the custom ID field, so that after starting a replication job, users can easily clear just that field if they're reusing selected dbs or remote creds?

If possible, we should remove the verification step where the account password is entered.  I'm not sure why it's a requirement.  If we do keep it, the button should be "Start Replication" rather than "Continue Replication," and we should specify which password is required, because it could be interpreted as a remote account's password.

Icing, but after selecting from a dropdown, can we make focus automatically change to the following object?

If there are illegal characters (spaces, prohibited symbols), or if the db name starts with something other than a lower-case letter in new-database-name field, can we disable the replicate button? I was able to "trigger" a replication from a local db to a new one called "spa ces" which then failed.

Thanks for your work on this.

benkeen:
Thanks @justin-mcdavid-ibm! Just emailed you - sorry, didn't realize you responded here.

benkeen:
Hey @garrensmith / @robertkowalski  - took a thousand years, but could I get one of you to review this sucker? Hope all is well on your ends.

app/addons/auth/actions.js
```diff
@@ -28,88 +28,135 @@ var errorHandler = function (xhr, type, msg) {
...
-    promise.then(function () {
-      FauxtonAPI.addNotification({ msg: FauxtonAPI.session.messages.loggedIn });
-      if (urlBack) {
-        return FauxtonAPI.navigate(urlBack);
-      }
-      FauxtonAPI.navigate('/');
-    });
-    promise.fail(errorHandler);
-  },
+  promise.done(function () {
```
garrensmith: `promise.done` should be `promise.then` that is more semantic.

app/addons/components/react-components.react.jsx
```diff
@@ -1160,6 +1171,61 @@ const ConfirmButton = React.createClass({
   }
 });
 
+
+// A simple, unstyled typeahead widget that can be used for typeahead on any arbitrary strings.
+// This should replace the more specific <JumpToDatabaseWidget />
+var TypeaheadField = React.createClass({
```
garrensmith: We have started using React-Select now instead. Could you rather use that.

app/addons/replication/actions.js
```diff
@@ -0,0 +1,90 @@
...
+function replicate (params) {
+  var promise = $.ajax({
+    url: window.location.origin + '/_replicator',
+    contentType: 'application/json',
+    type: 'POST',
+    dataType: 'json',
+    data: JSON.stringify(params)
+  });
+
+  var source = Helpers.getDatabaseLabel(params.source);
```
garrensmith: Can you make these const's

app/addons/replication/actions.js
```diff
@@ -0,0 +1,90 @@
...
+  var target = Helpers.getDatabaseLabel(params.target);
+
+  promise.then(function () {
```
garrensmith: You can also make these all `() =>` functions

app/addons/auth/actions.js
```diff
@@ -28,88 +28,135 @@ var errorHandler = function (xhr, type, msg) {
...
+    FauxtonAPI.addNotification({ msg: FauxtonAPI.session.messages.adminCreated });
+    if (loginAfter) {
+      FauxtonAPI.navigate('/');
+    } else {
+      FauxtonAPI.dispatch({ type: ActionTypes.AUTH_CLEAR_CREATE_ADMIN_FIELDS });
+    }
+  });
 
-  updateChangePasswordConfirmField: function (value) {
+  promise.fail(function (xhr, type, msg) {
```
garrensmith: Could you rather use the second function in the `then` promise to handle the fail case.

app/addons/replication/components.react.jsx
```diff
@@ -0,0 +1,527 @@
...
+  }
+
+  getAuthHeaders () {
+    const username = this.getUsername();
+    return {
+      'Authorization': 'Basic ' + base64.encode(username + ':' + this.state.password)
+    };
+  }
+
+  submit () {
```
garrensmith: Could you break this `submit` up a bit into different functions depending on if its local etc. This is a pretty large function. Its a little tricky to follow.

app/addons/replication/components.react.jsx
```diff
@@ -0,0 +1,527 @@
...
+        });
+        return false;
+      }
+    }
+
+    return true;
+  }
+
+  getReplicationSourceRow () {
+    const { replicationSource, databases, sourceDatabase, remoteSource } = this.state;
```
garrensmith: Can you break these replication source row into their own react components.

app/addons/replication/components.react.jsx
```diff
@@ -0,0 +1,527 @@
...
+class ReplicationTarget extends React.Component {
+  getOptions () {
+    const options = [
+      { value: '', label: 'Select target' },
+      { value: Constants.REPLICATION_TARGET.EXISTING_LOCAL_DATABASE, label: 'Existing local database' },
+      { value: Constants.REPLICATION_TARGET.EXISTING_REMOTE_DATABASE, label: 'Existing remote database' },
+      { value: Constants.REPLICATION_TARGET.NEW_LOCAL_DATABASE, label: 'New local database' },
+      { value: Constants.REPLICATION_TARGET.NEW_REMOTE_DATABASE, label: 'New remote database' }
+    ];
+    return _.map(options, function (option) {
```
garrensmith: You can just do `options.map` here.

test/nightwatch_tests/nightwatch.json.underscore
```diff
@@ -16,7 +16,7 @@
     "cli_args" : {
       "webdriver.chrome.driver" : "",
       "webdriver.ie.driver" : "",
-      "webdriver.firefox.profile" : ""
+      "webdriver.firefox.profile" : "default"
```
garrensmith: Why do you need to set this as default?

garrensmith:
Hey @benkeen this is making good progress. The basic functionality looks great. Just some code clean ups needed. We will need some nightwatch tests before we can merge this in.

benkeen:
Nice, thanks @garrensmith!

test/nightwatch_tests/nightwatch.json.underscore
benkeen: Doh.

app/addons/components/react-components.react.jsx
```diff
@@ -1160,6 +1171,61 @@ const ConfirmButton = React.createClass({
...
+      autoFocus: false
+    };
+  },
+
+  componentDidMount: function () {
+    this.updateTypeahead();
+  },
+
+  updateTypeahead: function () {
+    $(ReactDOM.findDOMNode(this.refs.field)).typeahead({
```
robertkowalski: we are removing the jquery typeahed and are using react-select instead.

it is very easy to integrate, here is an example: https://github.com/apache/couchdb-fauxton/blob/5be261085761c5f1d0510f2d28117775ea3e0da2/app/addons/documents/components/jumptodoc.react.jsx

robertkowalski:
good work! agree with garren on smaller components.

using `const` would be great

app/addons/replication/components.react.jsx
benkeen: Done.

benkeen:
Thanks again for the feedback, guys. All fixed up except for the NW test. @garrensmith did you ever get NW running locally? Not sure what I'm missing but I get a "Connection refused! Is selenium server started?" error when running `grunt nightwatch` locally.

garrensmith:
Hey @benkeen, yes its working. We have changed how it works though. It now runs through docker. So you need to install docker. See (here)[https://github.com/apache/couchdb-fauxton/blob/master/tests.md] and (here)[https://medium.com/@garrensmith/consistent-selenium-testing-with-docker-f2d5a24a1bc5]

app/addons/auth/actions.js
```diff
@@ -28,88 +28,133 @@ var errorHandler = function (xhr, type, msg) {
...
+  var promise = FauxtonAPI.session.login(username, password);
+
+  promise.then(function () {
+    FauxtonAPI.addNotification({ msg: FauxtonAPI.session.messages.loggedIn });
+    if (urlBack) {
+      return FauxtonAPI.navigate(urlBack);
+    }
+    FauxtonAPI.navigate('/');
+  });
+  promise.fail(errorHandler);
```
garrensmith: This needs to be changed to the .then

app/addons/auth/actions.js
```diff
@@ -28,88 +28,133 @@ var errorHandler = function (xhr, type, msg) {
...
-  },
+  promise.then(() => {
+    FauxtonAPI.addNotification({ msg: FauxtonAPI.session.messages.changePassword });
+    FauxtonAPI.dispatch({ type: ActionTypes.AUTH_CLEAR_CHANGE_PWD_FIELDS });
+  });
 
-  changePassword: function (password, passwordConfirm) {
-    var nodes = nodesStore.getNodes();
-    var promise = FauxtonAPI.session.changePassword(password, passwordConfirm, nodes[0].node);
+  promise.fail(errorHandler);
```
garrensmith: We mustn't use `promise.fail` we cant to eventually move away from jquery and use something else.

benkeen:
Too cool about Docker. Sheesh! Step away for 2 months and all this activity happens when you're gone. Can't say I approve! ;)

app/addons/auth/actions.js
benkeen: Yeah totally - this is old code, but I can update that too.

benkeen:
Hey @robertkowalski or @garrensmith, I think I need an extra pair of eyes on this. This ticket has dragged onnnnn and on and I have supremely limited time to work on it. I just can't get the NW tests to run reliably. Running just the one replication.js file works fine both locally & on travis, but when running all of them, it fails every time on Travis. Possibly it's actually a couch error silently failing to create the new replicated DB...? Maybe you can see something I can't. 

I left all the debugging code in.

garrensmith:
@benkeen I've continuing the work in #761

garrensmith:
@benkeen this worked was merged in with the other PR. Could you close this PR.

benkeen:
No problem. Hope all's going well, @garrensmith!

PR #671, opened 2016-03-23
Update NW to 0.8.18; asyncHookTimeout increased

benkeen:
This updates NW to use the latest version. Occasionally async
requests can fail to return by the default NW 10 second timeout;
this increases the value to 20 seconds to reduce the likelihood
of that occurring.

Note: the other refactoring was due to "this" no longer 
being consistent in the scenarios where it was called and 
helpers no longer being extended with all the runtime settings.


garrensmith:
+1 nice

benkeen:
Thanks! Merged as adaf1d30018a5aa4d4cc69262bb2e10f4b2d9a45

PR #672, opened 2016-03-24
remove comment

benkeen:
This just removes a console.log that's causing a lot of noise on the command line when running the dev server.


garrensmith:
Sorry I thought I removed that. +1

PR #676, opened 2016-03-31
Deleting cors domain moved to modal

benkeen:
This is part of the death-to-all-window.confirm()'s that Justin's
leading. The delete origin is now a modal.

Two related things:
- The cors page wasn't scrolling, hence the addition of the flex-body
class on the page element.
- the ConfirmationModal component now lets you pass in a class
for the main button to change the style.

screen shot 2016-03-31 at 2 34 28 pm


kxepal:
The message is awesome! It seems I can delete google.com now (:

app/addons/fauxton/components.react.jsx
```diff
@@ -403,7 +405,7 @@ function (app, FauxtonAPI, React, ReactDOM, ZeroClipboard, ReactBootstrap) {
             {content}
           </Modal.Body>
           <Modal.Footer>
-            <button className="btn btn-success js-btn-success" onClick={this.props.onSubmit}>
+            <button className={btnClasses} onClick={this.props.onSubmit}>
               <i className="fonticon-ok-circled"></i> {this.props.successButtonLabel}
             </button>
```
robertkowalski: we have a component for that:

https://github.com/apache/couchdb-fauxton/blob/master/app/addons/components/react-components.react.jsx#L1119-L1149

benkeen: Yeah, I guess I can refactor that component to allow passing the custom classes.

benkeen: Just FYI, I did try using the ConfirmButton component but ran into a circular reference problem where one file required the other. As discussed, way back when, I'm going to combine those two files + this problem will go away.

app/addons/components/react-components.react.jsx
```diff
@@ -1147,17 +1149,18 @@ define([
     },
 
     render: function () {
+      const { onClick, buttonType, id, style, text } = this.props;
```
robertkowalski: <3

robertkowalski:
+1

benkeen:
Thanks Robert. Merged as 30fdd49

PR #678, opened 2016-04-05
Editor: don't throw if document has no id

robertkowalski:
Right now the editor throws if you try to create a doc that does
nto have an id, instead of doing a POST to the database, which
will cause CouchDB to create an id for us.

Sometimes you want to be able to not use the pre-generated id from
Fauxton. Use cases are copy pasting a JSON object into the editor
or if you want to let the server take care of creating the id.

This patch lets fixes an invalid POST to `$DB/undefined` in cases
where the id wasn't specified.

Thanks to Will Holley for reporting!

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,6 @@
 {
   "presets": [
-    "react"
+    "react",
+    "es2015"
   ]
 }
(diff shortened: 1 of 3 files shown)


garrensmith:
+1 once tests pass

robertkowalski:
merged as d44aa93a92a200d763f9ca4632f2c0abc7a5f91d

PR #681, opened 2016-04-05
More minor fixes

benkeen:
- minor style fix for view attachments dropdown
- autofocus for delete db panel input field
- changes clipboard item to show notification when copied to
clipboard
- Now the API Bar tray only closes once you click outside of
it (not within it)

### DIFF
--- a/app/addons/documents/changes/components.react.jsx
+++ b/app/addons/documents/changes/components.react.jsx
@@ -273,6 +273,7 @@ define([
     getStoreState: function () {
       return {
         changes: changesStore.getChanges(),
+        loaded: changesStore.isLoaded(),
         databaseName: changesStore.getDatabaseName(),
         isShowingSubset: changesStore.isShowingSubset()
       };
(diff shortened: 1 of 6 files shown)


app/addons/documents/changes/components.react.jsx
```diff
@@ -351,6 +360,18 @@ define([
       };
     },
 
+    onClipboardClick: function (target) {
+      var msg = 'The document ID has been copied to your clipboard.';
```
robertkowalski: in ES 2015 we should use `let` if `const` is not possible because of a rebinding

app/addons/documents/changes/components.react.jsx
```diff
@@ -300,6 +301,14 @@ define([
     },
 
     getRows: function () {
+      if (!this.state.changes.length && this.state.loaded) {
+        return (
+          <p>
+            No document changes have occurred in this database.
+          </p>
+        );
+      }
```
robertkowalski: test coverage?

robertkowalski:
a test would be cool, aside from that +1

app/addons/documents/changes/components.react.jsx
benkeen: True, there's lot of instances where let will be better. Just wondering what you mean here though: wouldn't they be the same in this usage? I need msg to be defined in the function block to allow overriding. If I needed the var only in the `if` I could define it as a `let`, but I don't.

app/addons/documents/changes/components.react.jsx
benkeen: Oops.

PR #689, opened 2016-04-08
node5

robertkowalski:


robertkowalski:
we have problems with a flaky testsuite right now after moernizing our stack, one idea is that it might relate to an outdated node version.

> I curious how we would upgrade for Erlang 19 in this way (:

how do you mean?

kxepal:
> how do you mean?

I mean that you'd submitted 5 PR for node 5 update. Extrapolate that for Erlang 19 (;

janl:
LOL, obviously *something* went wrong ;D

screen shot 2016-04-08 at 15 04 13

PR #696, opened 2016-04-11
Upgrade PhantomJS to 2.x to fix Bitbucket errors

robertkowalski:
Most of our tests currently fail because the Bitbucket CDN, source
for PhantomJS errors. This upgrades the Phantom package to the
latest version where the issue got fixed in:
https://github.com/Medium/phantomjs/issues/161

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "mocha": "^2.4.5",
     "mocha-phantomjs": "~4.0.2",
     "nightwatch": "~0.8.18",
-    "phantomjs": "^1.9.1",
+    "phantomjs": "^2.0.1",
     "react-addons-test-utils": "^0.14.7",
     "sinon": "git+https://github.com/sinonjs/sinon.git",
     "url-loader": "^0.5.7"


robertkowalski:
example error: https://travis-ci.org/robertkowalski/couchdb-fauxton/builds/122250790#L1968

robertkowalski:
test plan from today:

- reset couchdb core to a version from january (try to fix 503 errors for setting up a test db on selenium setup) -- result: 5 / 5 fail
 - revert nightwatch upgrade (try to fix "random" fails in the middle of a testsuite run) -- result: 5 / 5 fail
 - set wait-time after boot of couchdb to 120 secs (try to fix 503 errors for setting up a test db on selenium setup)  -- result: 4 / 5 fail
 - revert webpack (long shot, but fits in the timeframe when errors occured)  -- result: 5 / 5 fail
 - upgrade to phantom 2 to avoid 403 errors (bitbucket cdn gives 403 errors for phantomjs 1.9) result: 2/5 failed, no 503 errors any more

benkeen:
Great that that 503 error is solved. I wonder, however, if it's worth merging this in until we solve the other issues too? What I mean is that we may find we need to downgrade other things to get it more stable - and possible downgrade phantom again in the process. I was having *some* success with PR 684 - certainly less than 2/5 were running.

benkeen:
Yeah, let's merge this guy and go from there. I personally like the `--suiteRetries 1` compromise in the other ticket. +1

robertkowalski:
merged! :)

PR #698, opened 2016-04-12
fixes fauxton to work as an npm release again

garrensmith:
Fix the package.json to get the correct npm modules for a `--production` install. Also set the `./bin/fauxton` server to look in the `/dist/release` folder for all assets.


robertkowalski:
+1

PR #700, opened 2016-04-12
React 15

robertkowalski:
Fixes and upgrades for React 15


app/addons/documents/index-results/index-results.components.react.jsx
```diff
@@ -21,14 +21,18 @@ define([
...
 
   '../../../../assets/js/plugins/prettify'
 ],
 
 
-function (app, FauxtonAPI, React, Stores, Actions, Components, Documents, FauxtonComponents, ReactBootstrap, Autocomplete, ReactCSSTransitionGroup) {
+function (app, FauxtonAPI, React, Stores, Actions, Components, Documents, FauxtonComponents,
+  ReactBootstrap, ReactSelect, ReactCSSTransitionGroup) {
+
+  require('react-select/dist/react-select.css');
```
garrensmith: Why not just add this to the other defines?

garrensmith:
+1 once tests pass.

app/addons/documents/index-results/index-results.components.react.jsx
robertkowalski: ah good call!

PR #703, opened 2016-04-15
update react-addons-test-utils@15.x

robertkowalski:


garrensmith:
@robertkowalski I think we are already now using 15.0.1. Is this PR still needed?

robertkowalski:
yo are right!

PR #706, opened 2016-04-25
Tweak tableViewConflicts NW test

benkeen:
This test fails occasionally. This softens it a little to
ensure the expected parent DOM element is there before
asserting stuff on it.


garrensmith:
+1 on condition the tests pass

PR #708, opened 2016-05-11
A more flux way of hiding the navbar

garrensmith:
This uses the store to hide and show the navbar. We still have to use a little jQuery to do this which is very frustrating. But we can only fix that once we have fully moved away from Backbone.


benkeen:
Looks and functions great. Be nice to update the store's private vars to use the underscore prefix, for consistency. But other than that +1

garrensmith:
Thanks. I've updated and added the `_` for the private variables.

PR #710, opened 2016-05-16
Fauxton tweaks

benkeen:
Two things:
- standardized button/link order in modals (cancel link first, action button second)
- adds filter icon. This'll be used for any "filter" input fields.


garrensmith:
+1

PR #713, opened 2016-05-19
Update the Changes page

benkeen:
- removes the filter tab and shows filter at top, always.
- updates to ES6
- spiffs up the styles a bit


app/addons/documents/changes/components.react.jsx
```diff
@@ -21,171 +21,177 @@ define([
...
 
-    render: function () {
-      var tabContent = '';
-      if (this.state.showTabContent) {
-        tabContent = <ChangesTabContent key="changesFilterSection" />;
+    showingSubsetMsg () {
+      const { isShowingSubset, changes } = this.state;
+      var msg = '';
+      if (isShowingSubset) {
+        var numChanges = changes.length;
```
robertkowalski: use let, it will not bleed out of the block

app/addons/documents/changes/components.react.jsx
```diff
@@ -21,171 +21,177 @@ define([
...
+      store.off('change', this.onChange);
+    }
 
-    render: function () {
-      var tabContent = '';
-      if (this.state.showTabContent) {
-        tabContent = <ChangesTabContent key="changesFilterSection" />;
+    showingSubsetMsg () {
+      const { isShowingSubset, changes } = this.state;
+      var msg = '';
```
robertkowalski: let

app/addons/documents/changes/components.react.jsx
```diff
@@ -21,171 +21,177 @@ define([
...
+          </p>
+        );
+      }
 
-  var ChangesHeaderTab = React.createClass({
-    propTypes: {
-      onToggle: React.PropTypes.func.isRequired
-    },
+      return _.map(changes, function (change) {
+        var key = change.id + '-' + change.seq;
```
robertkowalski: key can just be the index i a map provides, right?

```
changes.map((change, i) => {

})
```

app/addons/documents/changes/components.react.jsx
```diff
@@ -370,38 +313,41 @@ define([
...
+
+    render () {
+      const { codeVisible } = this.state;
+      const { change, databaseName } = this.props;
 
-    render: function () {
-      var jsonBtnClasses = 'btn btn-small' + (this.state.codeVisible ? ' btn-secondary' : ' btn-primary');
-      var wrapperClass = 'change-wrapper' + (this.props.change.isNew ? ' new-change-row' : '');
+      var jsonBtnClasses = 'btn btn-small' + (codeVisible ? ' btn-secondary' : ' btn-info');
+      var wrapperClass = 'change-wrapper' + (change.isNew ? ' new-change-row' : '');
```
robertkowalski: const

app/addons/documents/changes/components.react.jsx
```diff
@@ -413,47 +359,42 @@ define([
...
-      if (this.props.deleted) {
+  class ChangeID extends React.Component {
+    render () {
+      const { deleted, id, databaseName } = this.props;
+      if (deleted) {
         return (
-          <span className="js-doc-id">{this.props.id}</span>
+          <span className="js-doc-id">{id}</span>
         );
       } else {
```
robertkowalski: else not needed

app/addons/documents/changes/components.react.jsx
```diff
@@ -413,47 +359,42 @@ define([
...
-        var link = '#' + FauxtonAPI.urls('document', 'app', this.props.databaseName, this.props.id);
+        var link = '#' + FauxtonAPI.urls('document', 'app', databaseName, id);
```
robertkowalski: const

app/addons/documents/changes/components.react.jsx
benkeen: Haha we should add a lint rule to look for `var`'s. Save you having to do the role. ;-)

app/addons/components/react-components.react.jsx
```diff
@@ -112,7 +113,7 @@ var ToggleHeaderButton = React.createClass({
 
   render: function () {
     var iconClasses = this.props.iconDefaultClass + ' ' + this.props.fonticon + ' ' + this.props.innerClasses,
```
garrensmith: This can be simpler using es6 string templates:

app/addons/components/react-components.react.jsx
```diff
@@ -1,3 +1,4 @@
+
```
garrensmith: So minor. But I don't think we should have a line break here.

app/addons/components/react-components.react.jsx
```diff
@@ -85,7 +86,7 @@ var Badge = React.createClass({
         <span className="label label-info">{this.props.label}</span>
         <a
           href="#"
-          className="label label-info js-remove-filter"
+          className="label label-info remove-filter js-remove-filter"
```
garrensmith: Is the `js-remove-filter` still needed?

app/addons/components/react-components.react.jsx
```diff
@@ -111,7 +112,7 @@ var ToggleHeaderButton = React.createClass({
   },
 
   render: function () {
-    var iconClasses = this.props.iconDefaultClass + ' ' + this.props.fonticon + ' ' + this.props.innerClasses,
+    var iconClasses = `${this.props.iconDefaultClass} ${this.props.fonticon} ${this.props.innerClasses}`,
```
garrensmith: using `const` here would be cool. Also there are quite a few string concatenations in this function. I think we could use string templates to make them more readable. What do you think?

app/addons/documents/changes/actions.js
```diff
@@ -66,7 +61,7 @@ export default {
     var query = $.param(params);
     var db = app.utils.safeURLName(changesStore.getDatabaseName());
 
-    var endpoint = FauxtonAPI.urls('changes', 'server', db, '');
+    var endpoint = FauxtonAPI.urls('changes', 'server', db, '?' + query);
     currentRequest = $.getJSON(endpoint);
     currentRequest.then(_.bind(this.updateChanges, this));
```
garrensmith: Does `updateChanges` actually need to be bound?

app/addons/documents/changes/components.react.jsx
```diff
@@ -21,168 +21,174 @@ import ReactComponents from "../../components/react-components.react";
...
-  componentDidMount: function () {
-    changesStore.on('change', this.onChange, this);
-  },
+  }
 
-  componentWillUnmount: function () {
-    changesStore.off('change', this.onChange);
-  },
+  componentDidMount () {
+    store.on('change', this.onChange, this);
```
garrensmith: Do we need the `this`. I think each method should be bound already.

app/addons/documents/changes/components.react.jsx
```diff
@@ -199,165 +205,102 @@ var AddFilterForm = React.createClass({
...
-  },
-
-  onChangeFilter: function (e) {
-    this.setState({ filter: e.target.value });
-  },
+  }
 
-  inputClassNames: function () {
+  inputClassNames () {
     var className = 'js-changes-filter-field';
```
garrensmith: using the `js-` prefix was from Backbone days. Do we still need to do that?

garrensmith:
+1. Some small es6 suggestions. But other than that. It works nicely

app/addons/documents/changes/actions.js
benkeen: Yeah, `.then()` robs the `this` context so updateChanges can't reference `this.getLatestChanges()`.

app/addons/components/react-components.react.jsx
benkeen: Agreed. This whole method can be cleaned up,

app/addons/documents/changes/components.react.jsx
benkeen: Actually not in this instance. React doesn't handle all the binding for us, so we need to explicitly bind it so that onChange can reference this.getStoreState().

app/addons/documents/changes/components.react.jsx
benkeen: The rationale was that the `js-` prefixes would indicate the class only has significance for the JS code - especially test code - and not be used for any styling. It seemed a good idea at the time :)- but it is being used by a bunch of tests.

benkeen:
Thanks! Will merge when green.

PR #719, opened 2016-05-30
convert all files to use import

garrensmith:
Replaces all `define` amd requires with es6 `import


benkeen:
This is terrific. Found a couple of errors but nothing related to this, I'll fix them separately. +1

PR #721, opened 2016-05-31
Style fix for doc ID block

benkeen:
When you typed in something in the document ID typeahead field
it would look a little odd if there were no results (the grey
block would be fixed at 291px). This changes that to a max height
so it only ever appears as tall as it needs to be (and looks nice
when you're typing and honing in on a doc).

If I'm off the mark on this & there was a reason for the min-height,
just let me know.

Before: 
screen shot 2016-05-31 at 1 46 43 pm

After: 
screen shot 2016-05-31 at 1 51 54 pm

### DIFF
--- a/app/addons/documents/assets/less/jumptodoc.less
+++ b/app/addons/documents/assets/less/jumptodoc.less
@@ -20,3 +20,20 @@
   overflow: visible;
   position: absolute;
 }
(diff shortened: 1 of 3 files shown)


garrensmith:
+1 nice work

PR #726, opened 2016-06-06
only set query options tray visibility on initialize

benkeen:


robertkowalski:
+1 merge after green

garrensmith:
+1 once tests pass

PR #727, opened 2016-06-09
Jump to docs fetches from the server

garrensmith:
I haven't added any tests because this is already tested in nightwatch.


app/addons/documents/actions/actions.js
```diff
@@ -0,0 +1,39 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+import $ from 'jquery';
+import FauxtonAPI from "../../../core/api";
+
+export default {
+  fetchAllDocsWithKey: (database) => {
+      return (id, callback) => {
```
benkeen: extra space here. Also, should this file be in an `actions` subfolder?

benkeen:
Works great. +1

PR #729, opened 2016-06-15
Remove selector

garrensmith:
This is a major performance hit and is causing issues with how flex
works on some of the components


garrensmith:
Closing this for now. We will fix this later

PR #731, opened 2016-06-17
provide debugging help for monkey patched established methods

robertkowalski:
inheritance is so ridiculous. inheritance in javascript is
just monkey patching. today i spent 8 hrs to finding a bug where
a promise failed silently in a monkey patched route object.

i found it by stringifying a function at runtime with
`console.log(this.establish.toString());` in the renderpipeline
which calls establish on the routeobjects.

the promise in the establish fails, the whole renderpipeline
fails silently and the breadcrumbs component is not rendered.

this commit will make finding those issues in the future a lot
easier


garrensmith:
+1.

PR #741, opened 2016-07-02
show vertical scrollbar on attachment list when needed

samk1:
The max-height attribute of the ul element now scales to the size of the browser
window. overflow-y is set to auto so that a scrollbar appears and the user is always
able to view all attachments.

Closes COUCHDB-2993


garrensmith:
@samk1 this has a failing unit test. Could you take a look. Then we can review it.

samk1:
Thanks, unit test has been fixed.

app/addons/documents/doc-editor/components.react.jsx
```diff
@@ -218,44 +218,137 @@ var AttachmentsPanelButton = React.createClass({
...
+  },
+
+  getDefaultProps: function () {
+    return {
+      filter: this._filter
+    };
+  },
+
+  onChange: function (event) {
+    Stores.attachmentFilterStore.updateFilter(event.target.value);
```
garrensmith: With our react components we are using the [flux pattern](https://facebook.github.io/flux/) so a store must never have any values passed to it from any components. Rather in a component you would call an action e.g `updateFilter` and then that would dispatch an event with the details. The store would then react to that event.

app/addons/documents/doc-editor/components.react.jsx
```diff
@@ -218,44 +218,137 @@ var AttachmentsPanelButton = React.createClass({
...
+  },
+
+  stopClick: function (el) {
```
garrensmith: This is not the ideal solution. Could you rather add a onClick event to the `dropdown-menu` component and if its the filter it doesn't close it otherwise if its a link it opens it up in another tab. Even better would be to use the dropdown from [React-bootstrap](https://react-bootstrap.github.io/components.html#btn-dropdowns). It has an example with a filter. You don't need to make it as complicated as there example. But you just don't need to do any of these jquery event overrides.

app/addons/documents/doc-editor/components.react.jsx
```diff
@@ -218,44 +218,137 @@ var AttachmentsPanelButton = React.createClass({
...
 });
 
+var AttachmentsPanelFilter = React.createClass({
+  propTypes: {
+    filter: React.PropTypes.string
+  },
+
+  getDefaultProps: function () {
+    return {
+      filter: this._filter
```
garrensmith: `this._filter` is never defined in the beginning. I think you can actually make it state rather than a prop. And then just maintain it there.

app/addons/documents/doc-editor/components.react.jsx
```diff
@@ -218,44 +218,137 @@ var AttachmentsPanelButton = React.createClass({
...
     }
 
+    var ulStyle = {
+      maxHeight: this.state ? this.state.maxHeight : "none"
+    };
+
     return (
-      <div className="panel-section view-attachments-section btn-group">
+      <div className="panel-section view-attachments-section btn-group" ref={this.updateMaxHeight}>
         <button className="panel-button dropdown-toggle btn" data-bypass="true" data-toggle="dropdown" title="View Attachments"
```
garrensmith: Using `ref` like this isn't actually correct. I'm not actually sure why you need to change the height?

app/addons/documents/doc-editor/components.react.jsx
```diff
@@ -218,44 +218,137 @@ var AttachmentsPanelButton = React.createClass({
...
+            <a href={url} target="_blank" data-bypass="true"> <strong>{filename}</strong>
+              <span className="attachment-delimiter">-</span>
+              <span>{item.content_type}, {Helpers.formatSize(item.length)}</span>
+            </a>
+          </li>
       );
-    });
+    }
+
+    function filter(predicate) {
```
garrensmith: I think it would be better to do the filtering in the store. Here is a nice [example](https://github.com/apache/couchdb-fauxton/blob/master/app/addons/documents/changes/components.react.jsx#L117) of how we have done it before.

app/addons/documents/doc-editor/components.react.jsx
```diff
@@ -218,44 +218,137 @@ var AttachmentsPanelButton = React.createClass({
...
+
+  onChange: function () {
+    this.setState({
+      maxHeight: this.state.maxHeight,
+      filter: Stores.attachmentFilterStore.filter()
+    });
+  },
+
+  getMaxHeight: function () {
+    var $btn = $(this._button);
```
garrensmith: We React we shouldn't really have to ever use jQuery. I'm not sure why we need this.

app/addons/documents/doc-editor/components.react.jsx
```diff
@@ -218,44 +218,137 @@ var AttachmentsPanelButton = React.createClass({
...
+    };
+  },
+
+  componentDidMount: function () {
+    $(window).on('resize.attachment-filter', this.updateMaxHeight);
+    Stores.attachmentFilterStore.on('change', this.onChange);
+  },
+
+  componentWillUnmount: function () {
+    $(window).off('resize.attachment-filter', this.updateMaxHeight);
```
garrensmith: I think we could probably resize the attachment list using media queries. I'm guessing you want to limit how many attachments are visible depending on the screensize. If that is the case I would prefer a css solution to this. @robertkowalski might be able to recommend an approach.

app/addons/documents/doc-editor/components.react.jsx
```diff
@@ -218,44 +218,137 @@ var AttachmentsPanelButton = React.createClass({
     };
   },
 
+  getInitialState: function () {
+    return {
+      filter: '',
```
garrensmith: I think this filter should be kept in the store and the list of documents is retrieved from the store. So that the `getAttachmentsList` function goes back to just being a simple `map` function.

app/addons/documents/doc-editor/tests/doc-editor.componentsSpec.react.jsx
```diff
@@ -149,6 +149,24 @@ describe('DocEditorController', function () {
     assert.equal(attachmentURLactual, '../../id/_design/test-doc/one.png');
   });
 
+  it('view attachements dropdown max-height is not set to none', function () {
+    debugger;
```
garrensmith: Can you remove the debugger

garrensmith:
@samk1 functionality wise this is great and a really useful feature. Thank you. I think the key now is doing it in a more React way. I've tried to give some pointers. Please let me know where I can explain more or help more. I'm also `garrensmith` on IRC if you want to chat there.

app/addons/documents/doc-editor/components.react.jsx
robertkowalski: the results of `updateMaxHeight` etc are not used later (they are put into a ref) so i am assuming it is a leftover. the function can get removed together with the references to `maxHeight` and state.maxheight

justin-mcdavid-ibm:
This great functionality to add, @samk1.  Thanks for your time in working on this.

While testing, I found that file information is disappearing behind the scroll bar (Chrome 51), though.

Design-wise, I there are a few things you might consider:

You'd be able to show visual hierarchy better if you have the drop-down panel end some point before the bottom of the page.  Parallel scroll bars can be a bit disorienting if you can't easily intuit which panel is supposed to be on top.

Dialing-up the padding (from 6px to 12px) for the filter so that the field isn't shorter than drop-down entry fields will help draw attention to the functionality.

Making the filter case-insensitive would probably provide a bit more utility.

Adding an ability to easily clear a filter, though a clickable "clear" icon on the right side of the filter field would be helpful, but far from required.

Again, thanks for the work.

samk1:
Hello all.

Thanks very much for the advice and positive feedback. 

I have changed the dropdown to use react-bootstrap which seems to be a lot cleaner.

I am still having some issues though:

- the dropdown is closing at unwanted times - when you click on the filter and when you try and type anything. I have set the input to autofocus when the dropdown opens so clicks closing the dropdown is mitigated a little bit, but would still like to clear this up properly. closing the dropdown as soon as you type anything is obviously a bit of a show stopper.
- still not sure how to properly implement the max-height on the attachments dropdown. I have removed the hacks that were there before.
- there seems to be something wrong with the URL in the links. have not investigated this properly yet.

Anyway, I have mainly put this commit in to check with @garrensmith  if I have the data flow correct now. It does simplify the code a lot to do it this way.

I have seen some answers on stackoverflow that show how to prevent a dropdown with an input in it from closing [here](http://stackoverflow.com/questions/32510976/react-bootstrap-dropdown-with-input-wont-stay-open). I don't feel that any of these are the 'right' way though.

The dropdown filter example in the react-bootstrap docs is good but it removes the ability to close the dropdown when clicking outside of the dropdown, which I think we should still have.

The default react-bootstrap Dropdown implementation works by using RootCloseOverlay (in react-overlays) on Dropdown.Menu which adds event listeners to the document onclick and on onkeydown. for onkeydown, it closes the dropdown whenever escape is pressed. In this case the dropdown is closed when any key is pressed. The onclick event is ignored if it comes from inside the dropdown menu - so if you click inside the dropdown, the dropdown shouldn't close. You can see this if you click on the filter icon.

I think it is something to do with the input. I will keep on trying to find the root of this problem but would appreciated any help/suggestions in the meantime. I have not yet tried just removing the input to see if keypresses still close the dropdown.

@justin-mcdavid-ibm these are really good suggestions and I look forward to implementing them once I fix the problem with the dropdown closing before you can filter anything.

garrensmith:
Nice work. I'll try and take a look today.

garrensmith:
@samk1 I took a look at this, I cannot figure out why the input is causing the dropdown to disappear. I'll try and look over the next few days again.

samk1:
I have found one way to partially fix it using this [NativeListener](https://github.com/erikras/react-native-listener) component. Still having a problem where that input loses focus as after the attachment list updates though.

samk1:
The focus is getting stolen by the code editor because it's set to autoFocus. I have set it up so that when the attachments dropdown filter is toggled open the autoFocus changes to false on the code editor.

Still having a problem where the filter input does not have focus when the dropdown is first opened though.

Still need to work on conditionally styling the max height if there are more than n attachments. 'n' is going to depend on the browser window size.

samk1:
I've fixed the problems now. The only annoyance is the input is not focused after opening the dropdown. Everything else seems to be working fine.

garrensmith:
Hi @samk1 looks like you are making progress. But I'm still having an issue with this. When I click on view attachment it shows and then immediately disappears. 

This is a difficult problem. Can I suggest you park this PR for a bit and maybe work on the config section. Then we can come back to this. I need to find sometime to really dive into this issue.

samk1:
Hello Garren

Yep sure thing. I have almost finished the config section, I have written everything except for the tests which I intend to get done this weekend. The popover for adding a new option using react.js has given me an idea on how to make the attachments panel work better so I will apply that knowledge to this PR once the config section is complete.

Thanks for taking a look.

garrensmith:
Awesome. That is great.

garrensmith:
@samk1 could you take another look at his now. We have migrated fully to react. So the dropdown should work correctly now.

samk1:
Hi Garren

Sure thing. I will take another look.

Thanks for getting in touch again.

Regards, Sam

samk1:
@garrensmith It's been a long time since i've worked on this so master has moved a long way forward, any tips on how to best start working on this again with Git? I tried rebasing onto master but my branch is still missing all of the new stuff from master.

garrensmith:
You can try doing a cherry pick on top of master. Otherwise squish your commits to one and then rebase

On Sat, Mar 18, 2017 at 2:01 AM +0200, "samk1" <notifications@github.com> wrote:












@garrensmith It's been a long time since i've worked on this so master has moved a long way forward, any tips on how to best start working on this again with Git? I tried rebasing onto master but my branch is still missing all of the new stuff from master.

garrensmith:
I'm closing since this has fallen behind quite a bit. Please feel free to reopen if you able to work on it again.

PR #742, opened 2016-07-04
database ID gets URI encoded twice on url()

samk1:
Closes COUCHDB-3051

### DIFF
--- a/app/addons/databases/resources.js
+++ b/app/addons/databases/resources.js
@@ -111,7 +111,8 @@ Databases.Changes = FauxtonAPI.Collection.extend({
 
 Databases.Status = FauxtonAPI.Model.extend({
   url: function () {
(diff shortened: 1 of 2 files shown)


robertkowalski:
thanks for the PR! 

sadly it does not solve the root issue (try to create a database with the name `a%24` and another with the name "a$"

the root cause of the problem is a backend issue. in new versions, couchdb will reject those database names: https://issues.apache.org/jira/browse/COUCHDB-2748

samk1:
Ok, I think I get it - currently couchdb throws an error if you create a database called a$, but creates the database anyway. In new versions will it not create the database at all?

Similarly with a$-, it creates the database except fauxton is unable to get the status due to the double encode. When the backend is fixed, such a database will not be able to exist.

PR #743, opened 2016-07-07
display not exist error just on 404

robertkowalski:
just in case of a 404 the notification "The document does not
exist" is relevant.


garrensmith:
+1

PR #746, opened 2016-07-18
Show unreachable docs in jump to docs

garrensmith:
This fixes the issues where some docs were not reachable in the scrollable list.


robertkowalski:
+1

PR #754, opened 2016-07-27
testsuite: autocheck if document appears after creation

robertkowalski:


garrensmith:
Nice. +1

robertkowalski:
merged as 6a618eb

PR #755, opened 2016-07-28
	remove polling feature

robertkowalski:
there is no real world use case for watching the changes feed in
production. as the feature has a bug, we decided removing it would
make more sense.


garrensmith:
+1

garrensmith:
Looks cool. +1

robertkowalski:
merged as #7308368 - thanks!

PR #759, opened 2016-08-09
set jump-to-docs to be case sensitive

garrensmith:
Previously you could not search for a document with capitals in it e.g `MY_DOCUMENT`. This sets React-select to be case sensitive


millayr:
LGTM +1

PR #555, opened 2015-10-13
Sidebar automatically updates after design doc deleted

benkeen:
After deleting a design doc, the sidebar didn't refresh, so you'd
see the old data until the next repaint. This explicitly updates
the sidebar after all document deletes (React will handle the
nasty diffing for us, so it'll be performant).

PR #761, opened 2016-08-15
New replication

garrensmith:
This is the continuation of @benkeen's work in #669


app/addons/replication/assets/less/replication.less
```diff
@@ -11,187 +11,102 @@
...
-  &.showHeader {
-    li.header {
-      display: block;
-      border: none;
-    }
-    ul {
-      border:1px solid @greyBrownLighter;
-    }
+
+body .Select div.Select-control {
```
michellephung: hey :D 

I think the css specified in `body .Select div.Select-control` should move into the `.replication-page` brackets, it's nudging the jump-to-db input box on the all_dbs page over:

screen shot 2016-08-15 at 10 27 30 pm
 

also if you have time, move the other css into there too ;) so they can all be together

michellephung: OK i did it in a gist, since i don't know how to add to this this pr. You can copy and paste it:

https://gist.github.com/michellephung/71b8658d38f37d67331c240e366fbf3b

garrensmith: Awesome thanks Michelle. I'll make the changes.

garrensmith:
I've done a few fixes to the code and css, just to add consistency to where we moving with our React codebase. I think its working really nicely. I just need to write some unit tests for the React components. Just to note that local replication won't work in dev server mode. You can hack it by doing local replication then editing the url's in the `_replicator` document to point to the couchdb instance instead of the dev server.

@benkeen and @michellephung, I would love your thoughts.

benkeen:
Just looking over the code the changes look really tight, great stuff!

michellephung:
hi garren! hi ben! :D :D !

i will look at this over the weekend. looks cool so far

justin-mcdavid-ibm:
Hey Garren, I've given things a look.  Thanks for tag-teaming this with Ben.

Here're my comments:

## General Testing

I was able to replicate into an **existing** remote database using **target: New Remote Database**. That should fail.

Forward slashes are allowed in db names, but replication fails (on both remote-local and local-local) when a slash is included. I'm assuming that this might be an issue for remote sources, as well.  Also, replication messages aren't picking up the full db name.  Replicating (locally) from cat/dog to fish/bird triggers a message that 'Replication from "dog" to "bird" has begun.' 

Related FB ticket on the Cloudant side: https://cloudant.fogbugz.com/f/cases/66917/
Tangentially related (?) ticket: https://cloudant.fogbugz.com/f/cases/45464/
 
 "source": {
   "headers": {
    "Authorization": "Basic am1tY2Rhdmk6Z2V0aW4xMDFDbG91ZGFudA=="
   },
   "url": "https://jmmcdavi.cloudant.com/conflicts"
  },
  "target": "https://justinmcdavidibm:***@justinmcdavidibm.cloudant.com/conflict/testing",
  "create_target": true,
  "continuous": false,
  "owner": "jmmcdavi",
  "_replication_state": "error",
  "_replication_state_time": "2016-08-19T18:12:36+00:00",
  "_replication_state_reason": "{db_not_found,<<\"could not open https://justinmcdavidibm:*****@justinmcdavidibm.cloudant.com/conflict/testing/\">>}",
  "_replication_id": "7ebd686ea179ba68db907a776a85b74f"

## Usability

Are we unable to separate the database name from the remote-host info in the form (making two separate fields), and then recombine them for the actual request we're sending off?  Were I to want to trigger a number of replications in a row, it would be great if I could just change a database-name field, rather than modifying the end of the https://$REMOTE_USERNAME:$REMOTE_PASSWORD@$REMOTE_SERVER/$DATABASE string before clicking Start Replication again.  Fewer keystrokes and less mousing.

The Start Replication button is inactive until an "Existing Local Database" is selected, but for the other three options (existing remote, new local, new remote), the button activates before the fields are populated. The button should be in an inactive state until all required fields are populated.

To reduce triggering of multiple jobs (and the errors that would arise), after a Replication is triggered, the "Start Replication" button should go into an inactive state until a change has been made to the form. 

Along the same lines as the above, can we provide some validation for the fields (illegal characters, spaces, databases not following naming rules, replicating to and from the same database, replicating system databases that shouldn't be replicated (or should everything be up for grabs?)? Or are we leaving all of that to the server? The current lagtime and architecture between sending off replicate jobs that we know would fail leaves me leaning toward providing our own validation where we can.

Can the "Document Update Conflict" message that appears when you attempt to specify an existing replication document as the Custom ID be made more specific, so that users know that it's a conflict between replication documents, rather than a conflict arising out of the actual replication process?

## Styles

With the different dropdowns/fields, we have two different font families (Source Sans Pro and Helvetica Neue) and a couple different font sizes (14px-16px).  Let's make all the text (labels/fields) in that form 14px Source Sans Pro (so as to match the localdb-selection dropdown). And just for consistent units, let's change the example text for remote sources/targets from 9pt to 12px.

The confirmation dialogue styles appear broken, padding-wise.

The inactive "Start Replication" button has a hover state that should be removed.

## Strings
Change "e.g." to "Example:" in the case of remote-db example text.

(apologies to @benkeen ), but in the Replication doc field, I've had a change of heart, and think that just "Custom ID (optional)" will work as in-field text.

On the confirmation dialogue:
Change the dialogue's title to "Enter Account Password."
Change the explanatory test to "Replication requires authentication on your local account."
Change the in-field text to "Password"
Change the text on the button to "Start Replication," and add the check icon that we have on the main page's "Start Replication" button.

## Bloat
I don't have the Replication status/error pages in the Cloudant beta, can we implement those in Fauxton as part of this work?

app/addons/auth/components.react.jsx
```diff
@@ -302,9 +303,72 @@ var CreateAdminSidebar = React.createClass({
...
+
+  render () {
+    return (
+      <Modal dialogClassName="enter-password-modal" show={this.props.visible} onHide={() => this.props.onClose()}>
+        <Modal.Header closeButton={true}>
+          <Modal.Title>Enter Password</Modal.Title>
+        </Modal.Header>
+        <Modal.Body>
+          {this.props.modalMessage}
+          <input type="password" placeholder="Enter your password" autoFocus={true} value={this.state.password}
```
michellephung: i think you can add some newlines in here

app/addons/replication/components.react.jsx
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    return (
+      <div>
+        <div className="row">
+          <div className="span3">Database URL:</div>
+          <div className="span7">
+            <input type="text" className="connection-url" placeholder="https://" value={remoteSource}
+              onChange={(e) => onChange(e.target.value)} />
+            <div className="connection-url-example">e.g. https://$REMOTE_USERNAME:$REMOTE_PASSWORD@$REMOTE_SERVER/$DATABASE</div>
```
michellephung: maybe we should put this on 3 lines? it's long

app/addons/replication/components.react.jsx
```diff
@@ -0,0 +1,204 @@
...
+    // new and existing remote DBs show a URL field
+    if (replicationTarget === Constants.REPLICATION_TARGET.NEW_REMOTE_DATABASE ||
+        replicationTarget === Constants.REPLICATION_TARGET.EXISTING_REMOTE_DATABASE) {
+      targetLabel = 'Database URL';
+      remoteHelpText = 'https://$REMOTE_USERNAME:$REMOTE_PASSWORD@$REMOTE_SERVER/$DATABASE';
+
+      field = (
+        <div>
+          <input type="text" className="connection-url" placeholder="https://" value={remoteTarget}
+            onChange={(e) => Actions.updateFormField('remoteTarget', e.target.value)} />
```
michellephung: new lines in here would help readability

app/addons/replication/components.react.jsx
```diff
@@ -0,0 +1,204 @@
...
+          <div className="connection-url-example">e.g. {remoteHelpText}</div>
+        </div>
+      );
+
+    // new local databases have a freeform text field
+    } else if (replicationTarget === Constants.REPLICATION_TARGET.NEW_LOCAL_DATABASE) {
+      field = (
+        <input type="text" className="new-local-db" placeholder="Database name" value={targetDatabase}
+          onChange={(e) => Actions.updateFormField('targetDatabase', e.target.value)} />
```
michellephung: ditto here

app/addons/replication/components/options.js
```diff
@@ -0,0 +1,92 @@
...
+const ReplicationDoc = ({value, onChange}) =>
+<div className="replication-section">
+  <div className="replication-input-label">
+    Replication Document:
+  </div>
+  <div className="replication-doc-name">
+    <span className="fonticon fonticon-cancel replication-doc-name-icon" title="Clear field"
+      onClick={(e) => onChange('')} />
+    <input type="text" className="replication-doc-name-input" placeholder="Custom, new ID (optional)" value={value}
+      onChange={(e) => onChange(e.target.value)}/>
```
michellephung: starting to notice a pattern... is there some reason the inputs properties are inline?

app/addons/replication/components/source.js
```diff
@@ -0,0 +1,164 @@
...
+  onChange: React.PropTypes.func.isRequired
+};
+
+const ReplicationSourceRow = ({replicationSource, databases, sourceDatabase, remoteSource, onChangeRemote, onChangeLocal}) => {
+  if (replicationSource === Constants.REPLICATION_SOURCE.LOCAL) {
+    return <LocalSourceInput
+      value={sourceDatabase}
+      databases={databases}
+      onChange={onChangeLocal}
+           />;
```
michellephung: ws

app/addons/replication/components/source.js
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    return <ReplicationSourceRow
+      replicationSource={replicationSource}
+      databases={databases}
+      sourceDatabase={sourceDatabase}
+      remoteSource={remoteSource}
+      onChangeLocal={onLocalSourceChange}
+      onChangeRemote={onRemoteSourceChange}
+           />;
```
michellephung: ws

app/addons/replication/components/source.js
```diff
@@ -0,0 +1,164 @@
...
+
+  render () {
+    const {replicationSource, onSourceSelect, sourceDatabase, remoteSource, databases} = this.props;
+    const Actions = {};
+    return (
+      <div>
+        <ReplicationSourceSelect
+          onChange={onSourceSelect}
+          value={replicationSource}
+        />
```
michellephung: the ws on this one looks good :)

app/addons/replication/components/target.js
```diff
@@ -0,0 +1,202 @@
...
+  value: React.PropTypes.string.isRequired,
+  onChange: React.PropTypes.func.isRequired
+};
+
+const RemoteTargetReplicationRow = ({onChange, value}) => {
+  return (
+    <div>
+      <input type="text" className="replication-remote-connection-url" placeholder="https://" value={value}
+        onChange={(e) => onChange(e.target.value)} />
+      <div className="replication-remote-connection-url-text">e.g. https://$REMOTE_USERNAME:$REMOTE_PASSWORD@$REMOTE_SERVER/$DATABASE</div>
```
michellephung: maybe put onto 3 lines

app/addons/replication/components/target.js
```diff
@@ -0,0 +1,202 @@
...
+    return null;
+  }
+  let input;
+
+  if (replicationTarget === Constants.REPLICATION_TARGET.NEW_LOCAL_DATABASE) {
+    targetLabel = 'New Database:';
+    input = <NewLocalTargetReplicationRow
+      value={localTarget}
+      onChange={onLocalTargetChange}
+            />;
```
michellephung: ahh i get it

app/addons/replication/components/source.js
michellephung: ok nvm

app/addons/replication/components/source.js
michellephung: nvm

app/addons/replication/tests/replicationSpec.js
```diff
@@ -9,30 +9,205 @@
...
-import testUtils from "../../../../test/mocha/testUtils";
-var assert = testUtils.assert,
-    ViewSandbox = testUtils.ViewSandbox,
-    viewSandbox;
-
-describe('Replication Addon', function () {
-  describe('Replication View', function () {
-    var view = new Views.ReplicationForm({
-      collection: new Replication.DBList()
+/*import React from 'react';
```
michellephung: ?

michellephung:
hey garren, code looks really good :) 

I'm having trouble when i try to prepublish it as a package though. I'll try and ping you in the morning.

justin-mcdavid-ibm:
With regard to trying to start a replication job that uses a custom doc ID that conflicts with that of an existing document, we should be providing validation of some sort, and then giving the user a choice on what action to take.

Currently, if I fill out details for a remote replication job, but make a small mistake on the url/password, the replication document will be created, but the replication job will fail.  If I want to go in an make a small tweak to fix the details, but use the same custom ID, I'll get a doc-conflict error warning, and will then either need to manually delete the first document or change the custom doc ID to something new (at which point I'd probably want to delete the doc for the failed job just to keep things tidy).

My recommendation is to provide a validation step in between clicking the "Start Replication" button and the replication document actually being created/sent to the server.  If validation fails, give the user an opportunity to delete the existing, conflicting replication document before creating the new one.
artboard 1 copy 4

justin-mcdavid-ibm:
For replication activity I recommend that we implement something along the lines of the following mockup:
replicator db activity

justin-mcdavid-ibm:
1. Because the current view can only show replication jobs for which a replication document exists, the section should probably have a more specific title like "Replicator DB Activity."  It's not replication history-- cancelled continuous jobs will not appear, because we currently delete a continuous job's replication doc to stop it.  "Replication Activity" implies active jobs-- But the view shows errored/completed jobs.
2. We should implement the bulk-doc selector to give users an opportunity clean up completed jobs from the view.
3. Add sortable columns so that jobs can be organized.
4. Add a filter box so that jobs can be filtered.
5. Add "State" and "State Time" columns so that jobs can be ordered by recency. Failed jobs should be in our warning red, and completed jobs can be in green.
6. Improve access to local dbs in either of the "Source" or "Target" columns by making the db name clickable.  Because "Replicator DB Activity" is the top-level tab in the "Replication" section, it will be easy to navigate back view-- Meaning don't bother trying to implement any sort of complicated back functionality, depending on how a user gets to a db.
7. Create an action icon (wrench) that will take the user to the simple replication form, but pre-populated with details from the replication job.
8. Create an action icon (document) that will take the user to the listed job's replication document. Again, no need to implement complicated back functionality-- The activity view will be easy to return to. It would also be great if hovering over the icon could create a pop-up similar to the one we have for the copy icon table view of documents, and if the custom document ID, while not appearing as a value in the table, was still associated (metadata?) with the row, allowing users to "find" a custom doc through the filter.
9. Create an action icon (trashcan) that will delete a replication job's document (after action verification through a pop-up-- screenshot attached).
10. Implement an action icon (red triangular warning) for replication errors. The icon should only appears when errors exist.  I recommend lifting the code we use for warehousing error messages in Cloudant to display the actual errors. (screenshots attached).
11. The type column should be either "Continuous" or "One time," rather than true/false
error popup
error hover
delete verification

justin-mcdavid-ibm:
Hey Garren-- This is coming along great.  Comments below.

# Activity Page:

1. When I click on the column header, it changes the active-state coloring to orange, but I need to click a second time for it to do the actual sorting.
2. Is it possible to just highlight the sorting triangle, rather than the entire column-header text?
3. Change the Activity tab label to "Replicator DB Activity."
4. Add some padding (try 8px?) between the funnel icon and the Filter Replications entry box
5. Increase the text size of and padding around action icons.  Warehousing uses 16px for the icons. _all_dbs uses 19px (+enclosure box). I think 16px font and 8px left/right padding should suffice with our current information density.
6. Encoding for db names containing forward slashes still appears to be broken for source/target displays in Activity. The broken encoding also extends to the linking to the actual databases.
7. The action to edit the Replication document for a job listed in the activity section doesn't appear to be working yet.
8. Same with the action to view the Replication document.
9. The Verify Deletion dialogue does not differentiate between single-doc deletions, and multi-doc deletions (performed through the bulk-selector).  It would probably be better if the dialogue triggered through the bulk selector had it's own (pluralized) text and could reference the actual number of documents being deleted. (Instead of "You are deleting a replication document." make it "You are deleting <number> replication documents."
10. When does the Activity page refresh?  I'm wondering if we should have polling/refresh controls on the page, similarly to how we have them for Monitoring and Active Tasks.
11. Replication jobs that fail to create a new local database still have active links to the non-existant database.


# New Replication Page

1. The Start Replication button is not quite aligned with the field above it. If they could be left-aligned, it would be great.
2. Padding and password-entry-box length on the replication dialogue are broken. Also, what are requirements around this? How regularly should it be asking for password, if you're triggering multiple jobs.  Can we have permission extend over a session? 
3. Infield-text on the dialogue's password-entry box should be change to "Local-account password."  Apologies for the confusion, but please remove the period from the end of the dialogue box's title, as well.
4. I got a "Database target may not start with an underscore." error when attempting to replicate to an existing db named "a_c." So it looks like the validation might be a bit overzealous there. Also, is that message new? And a way to prevent replicating system dbs?  I'm a little gunshy to test this, but I see that the "Start Replication" button is active if I select my local _replicator as both source and target. Perhaps we can strip dbs starting with an underscore from the source/target dropdowns, as well?  Also, now I'm still getting the underscore error, but none of the information in the form contains underscores, so it looks like something buggier might be afoot.
5. Related to above: Did you add validation to prevent replicating a database to itself?
6. Change the "Database target may not start with an underscore." message to "The names of database sources and targets cannot begin with underscores." (and then you can reuse this message for underscored sources, as well.
7. I can still use local -> new remote replication to replicate into an **existing** remote database.  Same goes for replicating something to a "new" database that actually already exists. There should be an error for remote-instance-overwrite attempts, and validation for local-instance-overwrite attempts.
8. To reduce triggering of multiple jobs (and the errors that would arise), after a Replication is triggered, the "Start Replication" button should go into an inactive state until a change has been made to the form.
9. Can we implement the replication-doc conflicting handling described here: https://github.com/apache/couchdb-fauxton/pull/761#issuecomment-249692858

# General

1. The "Replication" breadcrumb for the replication section takes you to New Replication, rather than the Activity page; however, unless Advanced Replication is going to happen sometime soon (which it doesn't look like it will), I think it would be better to follow the same model we have for Warehousing, where there is a Create Replication button on the Activity page, and then the creation page has breadcrumbs and no tabs.
2. The message upon triggering a replication job is not quite correct.  It states that "Replication from <db1> to <db2> has begun." which isn't actually the case if the job errors.  That message should be "Replication from <db1> to <db2> is triggered." Though we could say "scheduled" instead of triggered, if you want to anticipate the dbnext-change rollout.

justin-mcdavid-ibm:
Hey Garren-- This is looking great.

Beyond the changes we talked about on video conference, here are some things I've noticed while playing around.  And I'm having problems getting remote->remote replication to work (I'm getting permissions errors that I'm not expecting), and local->remote are failing to complete.  So I'm not able to test some thing at this time.

# Activity Page

1. In-field text for "Filter replication" is 16px. Please reduce to 14px.  And let's get rid of the input box's padding; the box doesn't need to be so visually dominant.
2. Remove the "Ok" button on error-message dialogue boxes
3. To slightly improve readability, can we strip the first set of curly brackets off of displayed error messages?
4. The Edit Replication (wrench) functionality breaks if the replication document ID starts with a forward slash.
5. Can we have the Activity page column-sort default to State Time, descending from new-to-old?
6. When you add the refresh options to the header, could you make sure that column-sort selection persists over refreshes or when the back-button is clicked? When I refresh via the browser, it's resetting my sort selection.

# New Replication Page

1. The "Clear" replication-clear-link text is 16px. Please reduce to 14px.  Also increase the left-padding of replication-clear-link from 12px to something like 12px 0 12px 24px (ow whatever will move it a similar distance to the right).
2. On hover for truncated long values in dropdowns, can we have the full value display?
3. Remover the API URL button from the header on this page, too.
4. Change the title of the Replication Error dialogue to "Fix Document Conflict." Capitalize the first letter of the first paragraph in the message. Add the eraser icon (http://fontawesome.io/3.2.1/icon/eraser/) to the "Overwrite Existing Document" button.
5. Malformed remote source/target URLs prevent a replication from being sent off when the Start Replication button is clicked, but there's no indicator of why the job failed.  Can we add highlighting to URL fields that are malformed once the Start Replication button is clicked?

# Component issues?

1. Long db names in the local-db dropdown list are hidden behind the slider (this is a component issue?). We should either give them some right-padding so that the slider doesn't cover them, or allow horizontal scroll in the window.
2. The ellipses in truncated long values are running into the arrow on the right side of dropdowns. Truncating a character or two earlier would likely fix the issue. 
3. In the fields of the form, font colors are not consistent between selected text (#333) and entered text (#555). Let's make them a consistent #333, to match the label color.
4. Related, the ellipses for truncated values should also be #333 (they're currently picking up the color of in-field text).
5. The .Select-arrow-zone icons are not consistently colored, and don't have consistent color states. They're #000 with no hover state before a value is selected. After a value is selected, they change to #999, with a hover state involving an odd border-top color-change. Hovers of #F3812D in Cloudant and #E73D34 for CouchDB would be better for all dropown-select arrows, as well as for the field-clearing Xs, as we have in the Replication document field.

# Odd Ways to Break Things.

1. If I create a job and specify a custom replication doc of /// the Replication Activity section fails to to load. If I go into _replicator and delete the doc, it comes back.  A couple of hours later, I now can't reproduce this, so it might have been a system hiccup. Or an odd bug that just happens on first /// creation.
2. If I create a job and specify a custom replication doc of """ the replication doc never updates, even though the replication job completed. 

Thanks again for all the work on this.

justin-mcdavid-ibm:
Hi Garren,

1. Looking at replication, and as we discussed, we should combine the "New remote database" and "Existing remote database" options into a single "A remote database" option. We can't currently validate that a "new" database doesn't already exist, and back out of the replication process.

2. After chatting about the remote source/target issues, I think we should clarify our examples.  I don't want to clutter up the page with too much text, but I think we should do some handholding here, just to clarify. I propose that we simplify the URL example to:

Example: https://$USERNAME:$PASSWORD@$REMOTE_SERVER/$DATABASE

and then add a ? icon to the "Example:" label with the following explanatory text:

"If you have username and password for the remote account, use the remote username and remote password.

If your local account has permissions on the remote database, you can use your local username and local password.

If the remote database has granted permissions to "everybody," you do not need to include a username and password."

justin-mcdavid-ibm:
This is looking better and better.  There are a few bugs we still want to squish, but I think we're almost there.

# Create Replication Page

1. The Create Replication page should not have the polling or refresh components.
2. The dropdown arrows and the X for clearing the custom-doc field are picking up the Fauxton red hover state, rather than the Cloudant orange.
3. Most selected/entered text in the completed form is now #555555, However text in the Source Name dropdown is #333333.  I'd previously was hoping that all that text could be consistently #333333. If that's too much of a pain, I'd settle for consistently #555555.

# Activity Page

1. The hyperlink highlighting for local dbs listed in either source or target columns is being picked up by the final forward slash before the database name starts.  Might be tricky to fix, as I see that line-breaks are also happening before that final slash.
2. The down arrow for column sort is picking up Fauxton red, rather than Cloudant Orange

# Dialogues

1. In the last sentence of the Fix Document Conflict Dialogue, please change "Replicationed documents" to "replicated documents."
2. Remove the unstyled "Close" button on the error-message dialogue.
screen shot 2016-11-10 at 2 14 50 pm
screen shot 2016-11-10 at 2 54 51 pm
screen shot 2016-11-10 at 2 59 14 pm
screen shot 2016-11-10 at 1 57 39 pm

robertkowalski:
tests are red

app/addons/auth/actions.js
```diff
@@ -28,88 +28,130 @@ var errorHandler = function (xhr, type, msg) {
...
+  changePassword,
+  updateChangePasswordField,
+  updateChangePasswordConfirmField,
+  createAdmin,
+  authenticate,
+  updateCreateAdminUsername,
+  updateCreateAdminPassword,
+  selectPage,
+  showPasswordModal,
+  hidePasswordModal
```
robertkowalski: is this part of the replication work or a refactor? did something change, except the function type?

app/addons/auth/components.react.jsx
```diff
@@ -302,9 +305,87 @@ var CreateAdminSidebar = React.createClass({
...
+  // clicking <Enter> should submit the form
+  onKeyPress (e) {
+    if (e.key === 'Enter') {
+      this.authenticate();
+    }
+  }
+
+  // default authentication function. This can be overridden via props if you want to do something different
+  authenticate () {
+    const username = app.session.get('userCtx').name; // yuck. But simplest for now until logging in publishes the user data
```
robertkowalski: agreed

app/addons/replication/tests/helpersSpec.js
```diff
@@ -0,0 +1,41 @@
...
+      };
+
+      const dbName = helpers.getDatabaseLabel(db);
+      assert.deepEqual('fancy', dbName);
+
+    });
+
+  });
+
+});
```
robertkowalski: nice testing

test/dev.js
```diff
@@ -13,7 +13,7 @@
 
 // This will search for files ending in .test.js and require them
 // so that they are added to the webpack bundle
-var context = require.context('../app/', true, /[Ss]pec/);
+var context = require.context('../app/addons/replication', true, /[Ss]pec/);
```
robertkowalski: debug issue?

app/addons/replication/tests/controllerSpec.js
```diff
@@ -0,0 +1,27 @@
...
+import Controller from '../controller';
+import Constants from '../constants';
+
+const store = Stores.replicationStore;
+const assert = utils.assert;
+
+describe('Replication Controller', () => {
+
+
+});
```
robertkowalski: i think this can get deleted. i think the controler is coverd by selenium tests

app/addons/replication/route.js
```diff
@@ -10,48 +10,67 @@
...
+const ReplicationRouteObject = FauxtonAPI.RouteObject.extend({
+  layout: 'empty',
+  hideNotificationCenter: true,
+  hideApiBar: true,
+
   routes: {
-    "replication": 'defaultView',
-    "replication/:dbname": 'defaultView'
+    'replication/create': 'defaultView',
+    'replication/:dbname': 'defaultView',
```
robertkowalski: it looks like you can't edit a database with the name `create` using this routing

app/addons/replication/assets/less/replication.less
```diff
@@ -11,187 +11,288 @@
...
-      padding-right: 5px;
-      min-width: 130px;
-      padding-left: 0px;
-    }
-    input[type=radio],
-    input[type=checkbox] {
-      margin: 0 0 2px 0;
-    }
-  }
+.replication-seperator {
```
robertkowalski: wouldn't it be  `.replication__seperator` or am i missing someting?

robertkowalski:
issue 1

1) take a cloudant accoutn and database
2) put your cloudant account into the url field, e.g. "https://foo:bar@example.com/mydatabase
3) select target: local database enter name
4) click replicate
5) enter password
6) nothing happens - expected "password is wrong message"
7) i'm quite sure my password  is right?

image

app/addons/auth/actions.js
garrensmith: I just did a quick clean up here as I am using the password modal

app/addons/replication/assets/less/replication.less
garrensmith: Yes you are correct. I'll make the fix

garrensmith:
@robertkowalski with the admin party issue, I think the best fix there is to not show a password if the user is in admin party mode

garrensmith:
@robertkowalski tests are green 🎉

justin-mcdavid-ibm:
Hi Garren,

1. As we just discussed, the combining of the two remote options will not be feasible (thanks for the clarification), but we can add a caveat to the form so that users will not have incorrect expectations about system behavior (replicating into a pre-existing db when they think they're creating a new database). So, when the user selects "New remote database" from the dropdown, let's add the following statement below the example string.

   If a "new" database already exists, data will replicate into that existing database.

2. Also, when a user selects "New remote database," the middle sentence in the embedded help panel should not appear, (as db permissions cannot be set on not-yet-existing dbs).

3. On looking at the example string, I think the icon in the middle of the example isn't working as well as it could. I think the following changes to that string would be improvements: 
  - Remove the "Example:" label. I think the combination of grayed-out text plus the icon make it implicitly clearer as to what it is.
  - Move the hover help icon to the end of the example string.
  - Keep the string grayed out, but change the icon color to #08c
  - Give the icon an orange (#F3812D) hover in addition to the evoking of the help panel.

4. I noticed that hover-state for the clear-field X in the dashboard is now Cloudant-orange but that the drop-down arrows are still Fauxton red.

5. Add the following sentence to the Replication page when there are not any replication jobs showing:

   There is no replicator-db activity or history to display.

Component fixes (not necessary to finish this feature):

1. I'm still seeing a bit of styling-weirdness that happens when horizontal scrolling happens in the dropdown component.  But I think that's a component fix, and not necessarily for this feature.

2. The bulk selector checkbox remains selected after bulk-delete happens. Not a big thing,

3. The ellipses on long values selected in dropdown boxes still overlap with dropdown arrows, and are of a different text color than the selected values are.

Once we have those final changes (1-5), I believe we're good to go.  Thanks for all your work on this.

screen shot 2016-11-14 at 8 19 16 am
screen shot 2016-11-14 at 11 09 13 am
screen shot 2016-11-14 at 11 40 14 am

robertkowalski:
when i forget to add a database name to the url, it will still accept the url:

remote target:
```
https://mypassword@rockoartischocko.cloudant.com
```

expected: error that i have to specify a database name

robertkowalski:
screen shot 2016-11-15 at 15 21 00


typo: Replcator DB Activity.

robertkowalski:
wow @garrensmith and @justin-mcdavid-ibm that is really cool stuff i am super excited

robertkowalski:
update: discussed with garren, the validation shoudl happen on the serverside.

robertkowalski:
+1 after fixing the typo

justin-mcdavid-ibm:
Hey Garren,

Great work.  Thanks for all the effort that you put into this.

The UI looks ready to go. Everything appears to be behaving as it should.

+1

garrensmith:
Merged!!!!. Thanks @benkeen and @justin-mcdavid-ibm for the awesome help in this.

PR #774, opened 2016-09-14
remove sourcemaps in webpack

garrensmith:
We don't use this functionality so its just creating an unnecessary request.


robertkowalski:
+1

PR #775, opened 2016-09-26
fixes the typeahead tests

garrensmith:
Sometimes selenium tries to use the keys to trigger the options before they have loaded. This waits until their are options before selecting the doc.


robertkowalski:
+1

garrensmith:
merged

PR #776, opened 2016-09-26
remove update sequence from db list

robertkowalski:
right now the update sequence display is broken. they also make
no sense in a clustered setup.

this patch removes them test-wise. we can readd them if there is
a lot demand for seeing them, but as we got no bug reports in
almost 2 years it seems it is a dead feature.


garrensmith:
+1

robertkowalski:
merged, thx

PR #779, opened 2016-09-27
Fix design doc with include_docs

garrensmith:
This fixes the issue that if you are viewing the design docs for a
database and then click on the query options and click include_docs and
then run query it would stop showing the design docs.


robertkowalski:
+1

PR #765, opened 2016-08-24
Breadcrumb mango final

robertkowalski:
reactify former "breadcrumb"-area.

PR #780, opened 2016-09-27
views: remove fauxton side check for js warnings / errors

robertkowalski:
these should be checked in one single place, the database.

this way we don't have to maintain two separate whitelists for
errors.


robertkowalski:
note to myself: conflicts with https://github.com/apache/couchdb-fauxton/pull/765

garrensmith:
+1 nice idea.

PR #782, opened 2016-09-30
Encoding issues: fix for database names and refactor of db list

robertkowalski:
this fix for encoded database names also includes a big refactor of the database page, pulling logic out of the views and gets rid of backbone models
- no backbone models for the db list any more
- replaces the jquery tooltips on the db list with react tooltips
- avoiding safeURLName which is broken by design
- no method calls on passed data in views any more (e.g. `{this.props.db.get('name')}` - focus on plain JSON data from flux stores


robertkowalski:
note to self: clicktest db page, create new db

robertkowalski:
this fix for encoded database names also includes a big refactor of the database page, pulling logic out of the views and getting rid of backbone models

robertkowalski:
```
this fix for encoded database names also includes a big refactor of the database page, pulling logic out of the views and gets rid of backbone models

 - no backbone models for the db list any more
 - replaces the jquery tooltips on the db list with react tooltips
 - avoiding safeURLName which is broken by design
 - no method calls on passed data in views any more (e.g. `{this.props.db.get('name')}` - focus on plain JSON data from flux stores
```

```

robertkowalski:
@garrensmith yes, those are existing issues in the documents section we would have to fix in the future, too. 

they are also on master.

garrensmith:
ok then I'm +1

garrensmith:
Need to check the new replication section before merging. Or just after.

garrensmith:
If I click on a shared db's database I get `base.js:243 Uncaught TypeError: Cannot read property 'server' of undefined(…)`

garrensmith:
If I click on the x icon in shared db's I get `components.react.jsx:213 Uncaught TypeError: Cannot read property 'get' of undefined(…)`

PR #783, opened 2016-10-06
devserver: readd/fix csp header

robertkowalski:
The CSP headers were not set, as directly interacting with express
in `webpack-dev-server` to set the headers does not work.


garrensmith:
+1

PR #785, opened 2016-10-13
react-select: disable cache

robertkowalski:
react-select caches the results which leads to database names
appearing in the list of document names after you opened the
dropdown on the _all_dbs page.


robertkowalski:
no test as we would just test react-select as a library.

garrensmith:
+1

robertkowalski:
merged as 5a8a3fb

PR #789, opened 2016-10-18
3045 Replace safeURLName in code

millayr:
**Task description from ticket:**
We have a still some usages of safeURLName where we manually slice of the _design prefix for design docs. Later we glue the prefix on again. We have a helper function that handles that already, which is also available in the utils.js file.

Task: find appearances of safeURLName with code search, understand how it is used, remove custom safeURLName shenanigans with their design doc handling and replace it with other helper function. 

**Change Description**
I located two places where it made sense to update safeURLName to getSafeIdForDoc.  While there were several places we use safeURLName in Fauxton, not all of them needed to be changed.  I'm open to feedback, however, if others feel as though I missed some instances where it did make sense to update.

### DIFF
--- a/app/addons/documents/resources.js
+++ b/app/addons/documents/resources.js
@@ -80,8 +80,7 @@ Documents.DdocInfo = FauxtonAPI.Model.extend({
   // treated separately. For instance, we could default into the
   // json editor for docs, or into a ddoc specific page.
   safeID: function () {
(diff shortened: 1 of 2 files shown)


robertkowalski:
tests are red

robertkowalski:
tests are red @millayr - can you see why?

PR #791, opened 2016-10-21
add script to create/delete a lot dbs

robertkowalski:
this is useful to create a lot of dbs for testing and later to
get rid of them

### DIFF
--- a/bin/create-n-dbs
+++ b/bin/create-n-dbs
@@ -1,0 +1,131 @@
+#!/usr/bin/env node
+
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License. You may obtain a copy of
+// the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+
+const optimist = require('optimist');
+const nano = require('nano');
+
+const {parallelLimit} = require('async');
+
+
+const NAMESPACE = 'test-db-';
+const PARALLEL_LIMIT = 50;
+const URL = 'http://localhost:5984';
+
+optimist
+  .usage('Create a certain amount of dbs \nUsage: $0')
+
+  .options('help', {
+    alias: 'h',
+    describe: 'Show this message'
+  })
+  .options('number', {
+    alias: 'n',
+    describe: 'The amount of databases to create'
+  })
+  .options('delete', {
+    alias: 'd',
+    default: false,
+    describe: 'Delete databases'
+  });
+
+const argv = optimist.argv;
+
+if (argv.help || argv.h) {
+  return optimist.showHelp(console.log);
+}
+
+const isDelete = (argv.d || argv.delete);
+const isCreate = (argv.n || argv.number);
+
+if (!isDelete && !isCreate) {
+  console.error('[ERROR] Please provide -d or -n to delete or create test-databases \n');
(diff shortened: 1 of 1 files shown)


garrensmith:
+1 nice

bin/create-n-dbs
```diff
@@ -0,0 +1,131 @@
...
+const NAMESPACE = 'test-db-';
+const PARALLEL_LIMIT = 50;
+const URL = 'http://localhost:5984';
+
+optimist
+  .usage('Create a certain amount of dbs \nUsage: $0')
+
+  .options('help', {
+    alias: 'h',
+    describe: 'Show this message'
```
millayr: Seems a bit unhelpful ;).  Perhaps update this string to show a usage overview?

bin/create-n-dbs
```diff
@@ -0,0 +1,131 @@
...
+if (!isDelete && !isCreate) {
+  console.error('[ERROR] Please provide -d or -n to delete or create test-databases \n');
+  optimist.showHelp(console.log);
+  process.exit(1);
+}
+
+const nClient = nano(URL)
+
+if (isCreate) {
+  return createNdbs(argv.n || argv.number);
```
millayr: do you need "return" here?  doesn't look like createNdbs returns anything.

bin/create-n-dbs
```diff
@@ -0,0 +1,131 @@
...
+}
+
+if (isDelete) {
+  return deleteDbs();
```
millayr: same as above

bin/create-n-dbs
```diff
@@ -0,0 +1,131 @@
...
+    cb(null, list);
+  });
+}
+
+function createNdbs (number) {
+  getExistingDbs((err, dbs) => {
+    const offset = dbs.length;
+
+    const newDbs = [];
+    for (let i = offset; i < offset + number; i++) {
```
millayr: What's the purpose of the offset? Why not just iterate from 0 to the number?

bin/create-n-dbs
robertkowalski: it ends the execution path of the script

bin/create-n-dbs
robertkowalski: it would crash when runnign the script for a second time and you already have databases created. this way it will just add more databases (e.g. 1000 more)

bin/create-n-dbs
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    const tasks = newDbs.map((db) => {
+      return function task (cb) {
+        nClient.db.create(db, (err, body) => {
+          if (err) {
+            console.error(err);
+          }
+
+          cb(null);
```
millayr: What does this do?

bin/create-n-dbs
robertkowalski: the pattern is called an early return

bin/create-n-dbs
millayr: Yes, but what about "./bin/create-n-dbs -h"?

bin/create-n-dbs
robertkowalski: http://blog.timoxley.com/post/47041269194/avoid-else-return-early

bin/create-n-dbs
robertkowalski: its part of the help message that gets displayed.

bin/create-n-dbs
millayr: gotcha, makes sense

bin/create-n-dbs
millayr: ah, I see.  it's for naming purposes.  got it.

bin/create-n-dbs
millayr: Ok, makes sense now

millayr:
+1

robertkowalski:
merged, ty

PR #792, opened 2016-10-24
fix failing tests

garrensmith:


garrensmith:
Merged since this is just test fixes and passing tests prove this has worked.

PR #794, opened 2016-10-27
76046 - The Documentation Link in the Query Options Panel Points to Futon Documentation

millayr:
This is from Cloudant ticket https://cloudant.fogbugz.com/f/cases/76046/.
Corresponding Cloudant dashboard PR:  https://github.com/cloudant/dashboard/pull/619

The documentation link still pointed to Fauxton docs instead of Cloudant docs.  The getDefaultProps() call was happening before FauxtonAPI.constants.DOC_URLS.GENERAL was tweaked for Cloudant specific values.

The getDefaultProps() function is now removed and the docURL is passed as a prop to MainFieldsView.  The docURL is now a required property for the React class per advice from Robert Kowalski.

### DIFF
--- a/app/addons/documents/queryoptions/queryoptions.react.jsx
+++ b/app/addons/documents/queryoptions/queryoptions.react.jsx
@@ -30,13 +30,7 @@ var MainFieldsView = React.createClass({
     reduce: React.PropTypes.bool.isRequired,
     toggleReduce: React.PropTypes.func,
     updateGroupLevel: React.PropTypes.func,
-    docURL: React.PropTypes.string
-  },
-
-  getDefaultProps: function () {
-    return {
-      docURL: FauxtonAPI.constants.DOC_URLS.GENERAL
-    };
+    docURL: React.PropTypes.string.isRequired
   },
 
   toggleIncludeDocs: function (e) {
@@ -411,7 +405,8 @@ var QueryTray = React.createClass({
             reduce={this.props.reduce}
             toggleReduce={Actions.toggleReduce}
             groupLevel={this.props.groupLevel}
-            updateGroupLevel={Actions.updateGroupLevel} />
+            updateGroupLevel={Actions.updateGroupLevel}
+            docURL={FauxtonAPI.constants.DOC_URLS.GENERAL} />
           <KeySearchFields
             key={1}
             showByKeys={this.props.showByKeys}


app/addons/documents/queryoptions/queryoptions.react.jsx
```diff
@@ -30,13 +30,7 @@ var MainFieldsView = React.createClass({
     reduce: React.PropTypes.bool.isRequired,
     toggleReduce: React.PropTypes.func,
     updateGroupLevel: React.PropTypes.func,
-    docURL: React.PropTypes.string
-  },
-
-  getDefaultProps: function () {
```
garrensmith: This `getDefaultProps` should stay in.

app/addons/documents/queryoptions/queryoptions.react.jsx
```diff
@@ -411,7 +405,8 @@ var QueryTray = React.createClass({
             reduce={this.props.reduce}
             toggleReduce={Actions.toggleReduce}
             groupLevel={this.props.groupLevel}
-            updateGroupLevel={Actions.updateGroupLevel} />
+            updateGroupLevel={Actions.updateGroupLevel}
+            docURL={FauxtonAPI.constants.DOC_URLS.GENERAL} />
```
garrensmith: This should be changed to `docUrl={this.props.docURL}` then you can override this via `FauxtonAPI.constants.DOC_URLS.GENERAL` or via the `props` for this component.

app/addons/documents/queryoptions/queryoptions.react.jsx
millayr: If the docURL is always passed in as a prop, I'd argue keeping `getDefaultProps` results in redundant updates.  No?

app/addons/documents/queryoptions/queryoptions.react.jsx
millayr: The `QueryTray` class doesn't expect docURL as a prop at the moment.  I can update it so that it does, but then it seems as though we're just pushing this same behavior further up the chain of nested React classes.

Cloudant doesn't render `QueryTray` outside of fauxton code, so setting docURL differently (i.e. props versus default value) for each type seems like a lot of effort for minimal return.

garrensmith:
+1 I've merged it

PR #795, opened 2016-10-31
fix flaky nw test

robertkowalski:
turns out the dom element in the sidebar was hidden by a parent
dom element.

see: https://cloudup.com/ch2kUJzrmAy


garrensmith:
+1

PR #796, opened 2016-11-01
Create React Layout

garrensmith:
This creates a <Onepane/> layout that can be used instead of the
one_pane.html.

This is that start of removing the backbone.layoutmanager components.

### DIFF
--- a/app/addons/activetasks/components.react.jsx
+++ b/app/addons/activetasks/components.react.jsx
@@ -24,11 +24,11 @@ import ReactCSSTransitionGroup from "react-addons-css-transition-group";
 const TabElementWrapper = Components.TabElementWrapper;
 const TabElement = Components.TabElement;
 
-var activeTasksStore = Stores.activeTasksStore;
+const activeTasksStore = Stores.activeTasksStore;
 
-var ActiveTasksController = React.createClass({
+export const ActiveTasksController = React.createClass({
 
-  getStoreState: function () {
+  getStoreState () {
     return {
       collection: activeTasksStore.getCollection(),
       searchTerm: activeTasksStore.getSearchTerm(),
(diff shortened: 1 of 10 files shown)


app/addons/activetasks/layout.js
```diff
@@ -0,0 +1,39 @@
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
```
robertkowalski: `app/addons/activetasks/layout.js` -> `app/addons/activetasks/layout.jsx`

robertkowalski: shouldn't it be a .jsx file?

app/addons/components/assets/less/layouts.less
```diff
@@ -0,0 +1,4 @@
+
+.template {
+  height: 100%;
+}
```
robertkowalski: license header missing

app/addons/activetasks/components.react.jsx
```diff
@@ -221,7 +218,7 @@ var ActiveTasksTableHeader = React.createClass({
     });
   },
 
-  render: function () {
+  render () {
```
robertkowalski: i'm a fan of separating refactor step from the actual commit as they clutter the diff, but i think i lost that fight already long time ago haha

app/addons/fauxton/base.js
```diff
@@ -33,50 +33,51 @@ Fauxton.initialize = function () {
...
+      }
     }
 
     if (!routeObject.get('hideNotificationCenter')) {
       routeObject.setComponent('#notification-center-btn', NotificationComponents.NotificationCenterButton);
     }
 
-    FauxtonAPI.masterLayout.removeView('#breadcrumbs');
-
+    //FauxtonAPI.masterLayout.removeView('#breadcrumbs');
```
robertkowalski: remove please

app/addons/activetasks/routes.js
```diff
@@ -15,29 +15,24 @@ import FauxtonAPI from "../../core/api";
 import ActiveTasksResources from "./resources";
 import ActiveTasksComponents from "./components.react";
 import Actions from "./actions";
+import Layout from './layout';
 
 var ActiveTasksRouteObject = FauxtonAPI.RouteObject.extend({
   selectedHeader: 'Active Tasks',
-  layout: 'one_pane',
+  layout: 'empty',
+  hideBreadCrumbs: true,
```
robertkowalski: the way the new crumbs work now is a bit different, they just don't show up if no crumbs are provided. i don't like the implicity, but it helps to declutter route objects from functionality in order to get rid of them soon.

I see that the new <CrumbsWrapper /> supports the same functionality, too:

```js
const CrumbsWrapper = ({crumbs}) => {
   if (!crumbs || crumbs.length === 0) {
     return null;
   }

```

so i'm not sure if this is needed here and part of a new migration strategy?

app/addons/components/layouts.js
```diff
@@ -0,0 +1,94 @@
...
+        contentVisible={false}
+      />
+  </div>
+  );
+};
+
+const CrumbsWrapper = ({crumbs}) => {
+  if (!crumbs || crumbs.length === 0) {
+    return null;
+  }
```
robertkowalski: not a blocker, just sharing my thoughts.

in general i'm not a big fan of giving the responsibility to an element to decide if they should render or not. i prefer to have as much  logic outside as possible:

```js
{crumbs ? <CrumbsWrapper crumbs={crumbs} /> : null}
```

i see that this way to handle it can be convenient, just wanted to share my thoughts. :)

app/templates/layouts/empty.html
```diff
@@ -0,0 +1,2 @@
+<div class="template">
+</div>
```
robertkowalski: missing license header

app/addons/activetasks/routes.js
```diff
@@ -15,29 +15,24 @@ import FauxtonAPI from "../../core/api";
...
     'activetasks/:id': 'showActiveTasks',
     'activetasks': 'showActiveTasks'
   },
-  crumbs: [
-    {'name': 'Active Tasks'}
-  ],
-  apiUrl: function () {
-    var apiurl = window.location.origin + '/_active_tasks';
-    return [apiurl, FauxtonAPI.constants.DOC_URLS.ACTIVE_TASKS];
-  },
```
robertkowalski: wow cool! decluttering routeobjects! 👍

app/addons/activetasks/layout.js
garrensmith: Does it matter? Webpack doesn't care.

app/addons/activetasks/components.react.jsx
garrensmith: Sorry as I did that fix I knew you would be unhappy.

app/addons/activetasks/routes.js
garrensmith: ok I'll take a look see if I can remove this

robertkowalski:
+1 

great stuff!

garrensmith:
Merged

PR #799, opened 2016-11-02
redux: delete unused store

robertkowalski:
i was just about to convert our first module to redux, turns out
it doesn't need a store right now. so i deleted the store.


garrensmith:
+1 once tests pass

PR #808, opened 2016-11-16
Handle trailing params when building URLs for views

millayr:
In some circumstances, FauxtonAPI.urls() returns paths with a query parameter already appended to the end.  This results in URLs that contain multiple ?'s instead of one ? and multiple &'s.  The fix in this PR no longer assumes FauxtonAPI.urls() returns paths in a specific format.

### DIFF
--- a/app/addons/documents/resources.js
+++ b/app/addons/documents/resources.js
@@ -506,7 +506,7 @@ Documents.IndexCollection = PagingCollection.extend({
         view = app.utils.safeURLName(this.view),
         url = FauxtonAPI.urls('view', context, database, design, view);
 
-    return (url.indexOf("?") > -1) ? url + "&" + query : url + "?" + query;
+    return (url.indexOf("?") > -1) ? `${url}&${query}` : `${url}?${query}`;
   },
 
   url: function () {
--- a/app/addons/documents/resources.js
+++ b/app/addons/documents/resources.js
@@ -488,13 +488,13 @@ Documents.IndexCollection = PagingCollection.extend({
 
     if (params) {
       if (!_.isEmpty(params)) {
-        query = "?" + $.param(params);
+        query = $.param(params);
       } else {
         query = '';
       }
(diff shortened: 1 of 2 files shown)


app/addons/documents/resources.js
```diff
@@ -506,7 +506,7 @@ Documents.IndexCollection = PagingCollection.extend({
         view = app.utils.safeURLName(this.view),
         url = FauxtonAPI.urls('view', context, database, design, view);
 
-    return url + query;
+    return (url.indexOf("?") > -1) ? url + "&" + query : url + "?" + query;
```
garrensmith: Can you change this to use es6 template strings https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals

app/addons/documents/tests/resourcesSpec.js
```diff
@@ -460,5 +461,33 @@ describe('Bulk Delete', function () {
...
+    {
+      id:'myId2',
+      doc: 'num2'
+    }], {
+      database: {id: 'databaseId', safeID: function () { return this.id; }},
+      design: '_design/myDoc'
+    });
+  });
+
+  it('creates the right url with correct params when trailing question mark', function () {
```
garrensmith: Can you add another test to prove it works when you don't add the trailing `?`. 
Also we don't use `function ()` anymore with tests. You can use arrow functions http://exploringjs.com/es6/ch_arrow-functions.html

app/addons/documents/resources.js
millayr: Addressed in latest commit

app/addons/documents/tests/resourcesSpec.js
millayr: Addressed latest commit

robertkowalski:
+1

robertkowalski:
@millayr thank you. you can close the PR

PR #810, opened 2016-11-21
Remove one pane

garrensmith:
Remove the one_panel.html and use react instead


bin/create-animal-db
```diff
@@ -16,7 +16,7 @@
 // with conflicts for the zebra doc
 
 
-const url = 'http://localhost:5984/';
+const url = 'http://tester:testerpass@localhost:5984/';
```
robertkowalski: debug issue?

robertkowalski:
+1, merge it after removing the user/pass from the script :) 👍

bin/create-animal-db
garrensmith: I would like to change it to that. We always end up using the docker couchdb instance anyway. This is so much easier

garrensmith:
merging since this is green https://travis-ci.org/garrensmith/couchdb-fauxton/builds/177983807

PR #815, opened 2016-11-29
tests: add jest for testing

robertkowalski:
Adds Jest for testing. Right now it runs in parallel to our
old test setup.

The idea is that we can use the faster, simpler testing for all
new stuff and all refactored elements.

To run jest, simply run:

`npm run jest`


app/addons/replication/__tests__/helpers-tests.js
```diff
@@ -0,0 +1,34 @@
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License. You may obtain a copy of
```
garrensmith: do we have to have the __tests__ folder?

timmak: I have found <file>.test.js to work quite well, but I have seen the above pattern quite often.

garrensmith: @timmak I prefer something like that. I see we just have to edit the [testRegex](https://facebook.github.io/jest/docs/configuration.html#testregex-string). Since all our current tests have `Spec` i them. Any tests we want jest to run should have `.test` in them. I think that could be nicer than moving folders etc.

robertkowalski: hmmm... i like the new folder `__test__` -- it is jest standard and it separates the old system clearly from the new

robertkowalski: btw there will be also the folder `__mocks__` -- used by jest for mocking

garrensmith:
Ok I'm happy with this setup. +1

On Wed, Nov 30, 2016 at 3:34 PM +0200, "Tim Pinington" <notifications@github.com> wrote:












@timmak commented on this pull request.





In app/addons/replication/__tests__/helpers-tests.js:
> @@ -0,0 +1,34 @@
+// use this file except in compliance with the License. You may obtain a copy of



not a big fan of mocks so I am indifferent on this.

robertkowalski:
merged, thank you :)

PR #819, opened 2016-12-05
WIP - React component for clipboard

millayr:
This is a work in progress.
This PR will replace ZeroClipboard with a non-flash alternative, clipboard.js.
I still need to write up some tests as well.


app/addons/components/components/copy.js
```diff
@@ -0,0 +1,71 @@
...
+    );
+  }
+};
+
+Copy.defaultProps = {
+  displayType: 'icon',
+  textDisplay: 'Copy',
+  title: 'Copy to clipboard',
+  onClipboardClick: function () { },
+  clipboard: null
```
garrensmith: If its null, don't make it a defaultProp

app/addons/components/components/copy.js
```diff
@@ -0,0 +1,71 @@
...
+};
+
+Copy.propTypes = {
+  text: React.PropTypes.string.isRequired,
+  displayType: React.PropTypes.oneOf(['icon', 'text'])
```
garrensmith: Add the onClipboardClick callback here rather. And make it required

app/addons/components/components/copy.js
```diff
@@ -0,0 +1,71 @@
...
+
+import React from 'react';
+import ReactDOM from 'react-dom';
+import Clipboard from 'clipboard';
+
+let clipboard;
+
+export const initializeClipboard = (cb) => {
+  clipboard = new Clipboard('.copy');
+  if (cb != null) {
```
garrensmith: I don't really like this `cb` check. This basically means this will fail silently. Rather make sure that the react component always checks for a function via prop types that is passed into here.

app/addons/components/components/copy.js
millayr: heh, this was an artifact of an earlier implementation.  I removed this prop completely.

app/addons/components/components/copy.js
millayr: Done!

millayr:
Now to write some tests and remove the remaining zeroclipboard bits...

millayr:
This is now stale after some rookie git mistakes.  Please https://github.com/apache/couchdb-fauxton/pull/825 moving forward for reviews.

PR #822, opened 2016-12-14
Redux: update & fix

robertkowalski:
`getReducers` is mostly covered by our other tests.

no additional test coverage as we would need to build a separate build process (one without permissions) to test, so the cost for the test is to high in relation to the usefulness.


garrensmith:
+1 once tests pass

robertkowalski:
green!

PR #823, opened 2016-12-14
remove tray and use only traycontents

garrensmith:


app/addons/components/components/apibar.js
```diff
@@ -91,21 +88,8 @@ export const APIBar = React.createClass({
...
-      }
-
-      if ($(e.target).closest('.api-bar-tray,.control-toggle-api-url').length === 0) {
-        Actions.toggleApiBarVisibility(false);
-      }
-    }.bind(this));
-  },
-
-  componentWillUnmount () {
-    $('body').off('click.APIBar');
```
robertkowalski: one minor thing this hack was preventing:

1. open active tasks
2. open api url bar
3. go to other screen
4. bar reappears for 1-2 seconds

---

maybe you can try to solve the flickering, but when it takes far too long then we remove it anyway with the death of the api url bar?

robertkowalski:
+1

garrensmith:
merged

PR #824, opened 2016-12-16
Replication: hotfix url parsing for invalid urls

robertkowalski:
See COUCHDB-3257: sometimes the replicator returns invalid urls
which are not encoded. This makes standard conforming url parsers
choke.

This catches the exception, with the tradeoff of displaying the
password in the cases where the url is invalid.


millayr:
+1

garrensmith:
I made some changes here and its passing here https://travis-ci.org/robertkowalski/couchdb-fauxton/builds/204485099 

Apache travis is slow at the moment. So I'm going to merge.

PR #827, opened 2017-01-03
Allow hyphen in key name when matching edit line

Flamenco:
In my design doc, I can open a string editor for

```
"updates" : {
   "foobar" : "//comment"
   "foo_bar" : "//comment"
}
```
but not

```
"updates" : {
   "foo-bar" : "//comment"
}
```

An editor opens in the later case with no content.

This works however, as the key is map, not foo-bar when matching the edit line.
```
"views" : {
   "foo-bar" : {
        "map": "// a comment"
   }
}
```

### DIFF
--- a/app/addons/components/components/codeeditor.js
+++ b/app/addons/components/components/codeeditor.js
@@ -253,7 +253,7 @@ export const CodeEditor = React.createClass({
     // one JS(ON) string can't span more than one line - we edit one string, so ensure we don't select several lines
     if (selStart >= 0 && selEnd >= 0 && selStart === selEnd && this.isRowExpanded(selStart)) {
       var editLine = this.getLine(selStart),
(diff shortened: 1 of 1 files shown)


garrensmith:
Thanks @Flamenco this is a great find and a good fix. The changes have been merged into master

Flamenco:
Now if I could only edit the updates from the sidebar...  I tried to implement this, but it was not as simple as I had hoped.  After a couple of hours, I gave up and coded this beauty in about 15 minutes.  Perhaps I open source it and you integrated it?  It's react, but just not as *complicated* as your current approach.

screen shot 2017-01-12 at 10 31 04 am

garrensmith:
I won't be able to merge your work in. But we are working on a redesign for that section. I think @justin-mcdavid-ibm will be sending some some mockups for feedback once he is done

Flamenco:
Looking forward to that.  Hopefully there won't need to be as many 'clicks' to get the job done.

PR #828, opened 2017-01-05
76043 - tweaking hover color for attachment links

millayr:
The hover style for attachment links was not consistency with the rest of Fauxton.


garrensmith:
+1

PR #835, opened 2017-01-18
81371 - Unwanted scrollbars on notification center items

millayr:
divs in the notification center were scrollable even though content was not clipped.  I've removed the scollbars and have prevented unwanted scroll behavior.


garrensmith:
+1

PR #836, opened 2017-01-18
81574 - API Url test intermittently fails

millayr:
This test would periodically fail due to a timing issue.

Upon further investigation, the API Url modal would still be visible for a short amount of time after clicking on the Edit link.  Then when attempting to show the modal, the test framework would effectively be closing the modal instead.  This change essentially pauses the test suite until the old modal has time to unmount.


garrensmith:
+1

PR #837, opened 2017-01-19
Add an ETA field for active tasks tab

kevinjqiu:
- Show ETA for active tasks if possible (N/A if not)
- Show rate of change

Screenshot attached:

p


garrensmith:
@kevinjqiu thanks for this PR. Its great that you are contributing. The only thing is how accurate do you think that ETA is? Do we want to show users an ETA where we cannot guarantee the accuracy of. 

@rnewson and @janl I would love your thoughts on this.

kevinjqiu:
@garrensmith Thanks for the reply. It's accurate as of the moment the stats is pulled (from `/_active_tasks`) and is calculated. If the rate of change varies significantly from one moment to the other, then the estimate is going to fluctuate. As per my testing, when the rate of change stabilizes, the estimate is more accurate.

That said, the estimate is an estimate after all, and is as good as the data it can work with. As long as the users are aware of the fact the ETA is calculated given the data at that moment, I think it's a good piece of information to show :)

rnewson:
We had something like this in the old cloudant dashboard (quite some time ago). My experience is that it's initial ETA was wildly inaccurate, based on a single _active_tasks call, but warmed up as it saw more data. I suggest that this approach be followed in fauxton.

garrensmith:
@rnewson brings up a really good point. Since active tasks is polling in the background. Instead of calculating the ETA on one request. Could you rather wait for say 3 - 5 requests and then do an aggregate calculation for the ETA. I think that will stabilise the ETA a bit and make it a little more consistent between active task refreshes.

You could use local storage/pouchdb to keep a list of previous ETA's so that whenever the user views the active tasks page they immediately get an ETA for a long running task.

kevinjqiu:
@garrensmith Yeah that's a good idea.

garrensmith:
Awesome. Let me know if I can help in anyway

justin-mcdavid-ibm:
@kevinjqiu great idea and work. If we add ETA, though, I think it might make sense to re-order the display of information slightly.  Placing **Started on** and **Estimated Completion** side-by-side would make it easier to make quick relative comparisons.  And it might make sense to add the **Updated on** value to the value list in the **Status** column-- which would mean that the display on smaller screens (like mine) wouldn't be squished by the addition of a new column.

Thanks for the work on this.

garrensmith:
I'm closing since this has fallen behind quite a bit. Please feel free to reopen if you able to work on it again.

PR #840, opened 2017-01-24
Generic tab layout wrapper for Fauxton

millayr:
We're expecting to converge on a consistent tab layout thanks to many of Justin's fauxton designs.  This PR adds a new component that we can reuse for handling the tab content and logic.


millayr:
been merged

PR #847, opened 2017-02-09
lodash updates switching from _.contains to _.includes

millayr:
After a large set of dependency updates we were left with lodash misbehaving.  Switching to _.includes should resolve this problem.


garrensmith:
+1. I've merged. Can you close this PR

PR #849, opened 2017-02-14
Currently reducers seem to be handled quite strangely 

timmak:
~~this is a slightly better work around for now~~

Sorry @rnewson you were right I have add some comments to the code. The current solution seem to only allow for one large reducer this follow the more conventional way of adding reducers.


rnewson:
This description is not very helpful.

app/main.js
```diff
@@ -57,24 +57,10 @@ $(document).on("click", "a:not([data-bypass])", function (evt) {
...
-  }, {});
-}
-
-const reducer = getReducers(FauxtonAPI.reducers);
-
 const middlewares = [thunk];
 
 const store = createStore(
-  reducer,
+  combineReducers(FauxtonAPI.reducers),
```
timmak: Instead of having a single reducer we use redux combineReducer so we can name space the different reducers
http://redux.js.org/docs/api/combineReducers.html

app/core/base.js
```diff
@@ -157,6 +157,10 @@ FauxtonAPI.setSession = function (newSession) {
   return FauxtonAPI.session.fetchUser();
 };
 
-FauxtonAPI.reducers = [];
+FauxtonAPI.reducers = {};
+
+FauxtonAPI.addReducers = (reducers) => {
+  FauxtonAPI.reducers = Object.assign({}, FauxtonAPI.reducers, reducers);
+};
```
timmak: This not the most elegant solution but reducers are named space by an object so when we add reducer to the store we merge it with previous reducers object and reassign it

app/addons/permissions/container/PermissionsContainer.js
```diff
@@ -24,14 +24,14 @@ import {
 } from '../reducers';
 
 
-const mapStateToProps = (state) => {
+const mapStateToProps = ({permissions}) => {
   return {
```
timmak: Permission are now named space in the reducer so with need to pick permissions from the whole state

app/addons/permissions/base.js
```diff
@@ -12,11 +12,13 @@
...
-import reducer from './reducers';
+import reducers from './reducers';
 import "./assets/less/permissions.less";
 
 Permissions.initialize = function () {};
 
-FauxtonAPI.reducers.push(reducer);
+FauxtonAPI.addReducers({
+  permissions: reducers
+});
```
timmak: We declare the name and the reducer to use, using the add reducer api

rnewson:
thanks for the clarifying comments. I can't personally review the work, we'll need @garrensmith

app/core/base.js
garrensmith: This is really cool. What about if you used the [spread operator](https://babeljs.io/docs/plugins/transform-object-rest-spread/) here. Then this should be:

```
FauxtonAPI.addReducers = (reducers) => {
  FauxtonAPI.reducers = {
   ...FauxtonAPI.reducers,
   reducers
 };
```

garrensmith: I haven't tested if this will work or not. What do you think?

timmak: Unfortunately the build and linting isn't currently setup to whole object spread syntax, though this is a lot nicer

garrensmith: Can't we just add the npm package and it will work? I would be happy for you to do that in this PR.

timmak: Ok all done should be good to go once tests have passed

garrensmith:
+1 this is really cool. Could you rebase from master, I think we had an issue with the docker image. The tests should pass after that and then I can merge.

garrensmith:
Thanks @timmak I've merged this in. Can you close this PR. 

For future reference don't merge from master into your working branch. Rather rebase master into your PR branch. And then when getting your code into master do a merge then. Otherwise it makes a mess of the git history. I've fixed up the git history and squished your commits into one.

PR #850, opened 2017-02-15
extract css to separate file in production

nolanlawson:
With this change, we use extract-text-webpack-plugin to extract CSS from our JavaScript bundle into a separate CSS file.

There are a few performance benefits to this change:

1. JS bundle size is smaller (2.4MB -> 2.2MB)
2. CSS can apply before the JS executes, meaning the browser can start laying out styles faster, and users see a more progressive loading experience. In our case this is just our gray background instead of a white background, but at least it's something.

Before:

screenshot 2017-02-14 20 31 45

After:

screenshot 2017-02-14 20 31 48


garrensmith:
This is brilliant. Thanks @nolanlawson. I've merged it.

PR #851, opened 2017-02-15
use modular pouchdb

nolanlawson:
If we're only using PouchDB to do HTTP requests, we can just use pouchdb-core with pouchdb-adapter-http. This reduces the bundle size from 2.555MB to 2.477MB and also makes `npm install` faster because we don't have to install LevelDB anymore.

Looking at the code, we could probably just replace this all with a `fetch()`, but that can be done in a later PR.


garrensmith:
+1. Thanks @nolanlawson I've merged it in.

PR #852, opened 2017-02-15
trim moment locale using IgnorePlugin

nolanlawson:
This trims the release bundle.js size from 2554776 bytes to 2372243 bytes. Apparently there is a large part of the bundle taken up by moment.js, but AFAICT we're not actually using moment/locale anywhere, so this removes it.

See https://github.com/moment/moment/issues/2373 for why this seemed like a good idea.

### DIFF
--- a/webpack.config.release.js
+++ b/webpack.config.release.js
@@ -30,6 +30,9 @@ module.exports = {
         'NODE_ENV': JSON.stringify('production') // This has effect on the react lib size
       }
     }),
+    // moment doesn't offer a modular API, so manually remove locale
+    // see https://github.com/moment/moment/issues/2373
+    new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/),
     new webpack.optimize.DedupePlugin(),
     new webpack.optimize.UglifyJsPlugin({
       compress: {


nolanlawson:
Context: I'm using webpack-bundle-analyzer to try to figure out where we can trim the JS bundle:

screenshot 2017-02-15 00 35 54

garrensmith:
This is awesome. Managing the bundle size is definitely a neglected side of Fauxton. The builds are taking long to run today. But soon as they have built I will merge.

nolanlawson:
Yeah the long build times seem entirely related to Webpack, but I'm not super familiar with Webpack so I'm not sure how to optimize.

In any case there is some other stuff I wanted to do to improve build time but haven't gotten around to:
- split `brace` into separate code-split bundle using `require.ensure()`
- use modular lodash, e.g. with `babel-plugin-lodash` and `lodash-webpack-plugin`, or just stuff like `var map = require('lodash/map')`
- use rollup to compress the size of files we control, e.g. that whole app/ folder should be one big JS file rather than separate modules (seems [rollup-loader](https://github.com/egoist/rollup-loader) would work well for this)
- alternatively, upgrade to webpack 2 because it seems that would cut down the react-bootstrap size since it's exposed as an ES module and therefore the tree-shaking would automatically trim a lot of the cruft. potentially rollup-loader could also do this but needs investigation

garrensmith:
I've documented your ideas here https://issues.apache.org/jira/browse/COUCHDB-3303

garrensmith:
Thanks. I've merged it in.

PR #853, opened 2017-02-16
code-split "brace" into second bundle

nolanlawson:
I am not 100% confident in this PR but maybe it's a good starting point for discussion.

I noticed that `brace` makes up a lot of the default bundle. I also noticed that you don't really need `brace` until you open an editor. So it seems like a good opportunity to code-split.

Before:

```
2554776	bundle.js
```

After:

```
252701	1.bundle.js
2302814	bundle.js
```

So this removes 251962 bytes (~250KB) from the default bundle.

I've tested this and it does indeed work. When you open up an editor, `1.bundle.js` downloads and you can edit JavaScript/JSON. Open questions, though:

1. What is zenmodeoverlay? I'm not sure how to test it. I'm also not sure how to dynamically load `brace/theme/dawn` inside of it.
2. Should we move `1.bundle.js` to `dashboard.assets`? TBH I'm not sure why we even move `bundle.js` into a content-addressed file. It shouldn't be extremely valuable to make it content-addressed, unless folks are frequently upgrading their databases and we're worried that the cache might be stale.

### DIFF
--- a/app/addons/components/components/codeeditor.js
+++ b/app/addons/components/components/codeeditor.js
@@ -12,12 +12,17 @@
 import React from "react";
 import ReactDOM from "react-dom";
 import FauxtonAPI from "../../../core/api";
-import ace from "brace";
 import {StringEditModal} from './stringeditmodal';
 
-require('brace/mode/javascript');
-require('brace/mode/json');
-require('brace/theme/idle_fingers');
+function ensureAce(callback) {
+  // dynamically load brace because it's large
+  require.ensure([
+    'brace',
+    'brace/mode/javascript',
+    'brace/mode/json',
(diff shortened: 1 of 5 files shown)


garrensmith:
This is really interesting. I want to hold off merging it and do a bit of research on this. It also might be worth upgrading to webpack 2.

nolanlawson:
Definitely, this is a risky change and should be thoroughly tested. But potentially nice performance gains for first load!

nolanlawson:
This is probably too risky and doesn't help a whole heck of a lot. Maybe it would make more sense to trim the mono-bundle first (e.g. underscore to lodash, lodash-webpack-plugin, etc.) and then tackle this later.

PR #854, opened 2017-02-18
don't encode username/pass unless necessary

nolanlawson:
This along with https://github.com/pouchdb/pouchdb-server/pull/170 should fix https://github.com/pouchdb/express-pouchdb/issues/430.

The goal is just to avoid encoding the username/password if we're in admin party mode and neither is required. This avoids the `http://null:@localhost` bug.

Unfortunately this doesn't 100% fix the problem, because it relies on the URL of fauxton being the same as the URL of the CouchDB itself, which is not always the case. But it'll at least fix things for pouchdb-server as-is.

### DIFF
--- a/app/addons/replication/api.js
+++ b/app/addons/replication/api.js
@@ -40,6 +40,9 @@ export const getUsername = () => {
 };
 
 export const getAuthHeaders = (username, password) => {
+  if (!username || !password) {
+    return {};
+  }
   return {
     'Authorization': 'Basic ' + base64.encode(username + ':' + password)
   };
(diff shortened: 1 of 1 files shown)


nolanlawson:
One thing I was very confused about with this code is that, to replicate to or from local DBs, you don't need to specify the full URL; you can just specify the DB name. E.g.:

```bash
curl -X POST http://localhost:6984/_replicator \
  -H 'content-type:application/json' \
  -d '{"_id": "test_replication1", "source": "http://examples.cloudant.com/animaldb", "target": "animaldb", "create_target": true}'
```

This works fine; is there any reason to use the full URL for local DBs? It also breaks when you run Fauxton at a different root than the database itself.

garrensmith:
@nolanlawson I've added a test and continued this work here https://github.com/apache/couchdb-fauxton/pull/858

If the tests all pass I'll merge it all in.

garrensmith:
One thing, we need full urls because CouchDB 2.0 requires it.

nolanlawson:
okay, closing in favor of https://github.com/apache/couchdb-fauxton/pull/858

PR #855, opened 2017-02-18
Update CONTRIBUTING with quickstart instructions

nolanlawson:
These quickstart instructions would have helped me out a lot when I was getting started with Fauxton. Think of it as a "TL;DR, I just want to start contributing now." :smiley:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,6 +8,45 @@ with the framework you'd need to complete the task.
 
 The [Readme file](https://github.com/apache/couchdb-fauxton/blob/master/readme.md) has information about how to get the project running.
 
+## Contributor quick start
+
+Instructions to get a dev environment up and running as fast as possible:
+
+First, ensure that you have Node and npm installed. You should also have either CouchDB 2.0+ or PouchDB Server. The easiest to install is PouchDB Server:
+
+```
+npm install -g pouchdb-server
+pouchdb-server --port 5984
+```
+
+Now that we have a CouchDB (or PouchDB Server) up and running, check out the code:
+
+```
+git clone https://github.com/apache/couchdb-fauxton
+cd couchdb-fauxton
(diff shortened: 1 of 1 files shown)


nolanlawson:
One remaining issue, though, is apparently you can't disable Content-Security-Policy in dev without changing `devserver.js`. Seems like this ought to be configurable by users? Otherwise it's a bit painful, because you can't debug a separate CouchDB/PouchServer running at a different origin.

CONTRIBUTING.md
```diff
@@ -8,6 +8,45 @@ with the framework you'd need to complete the task.
...
+
+Now that we have a CouchDB (or PouchDB Server) up and running, check out the code:
+
+```
+git clone https://github.com/apache/couchdb-fauxton
+cd couchdb-fauxton
+npm install
+```
+
+Next, copy `settings.json.default.json` to `settings.json`. This will be our local settings file.
```
garrensmith: It will automatically read `settings.json.default.json` so you only need a `settings.json` if you have custom code.

CONTRIBUTING.md
```diff
@@ -8,6 +8,45 @@ with the framework you'd need to complete the task.
...
+
+In the `settings.json`, under `"development"` -> `"app"`, change the `"host"` to point to your local CouchDB or PouchDB Server, e.g.:
+
+```js
+"development": {
+  /* ... */
+  "app": {
+    /* ... */
+    "host": "http://localhost:5984"
```
garrensmith: Weird. I thought it would default to this port. Is that not the case?

nolanlawson:
I thought I should specify, because some folks may have CouchDB 1.x already running on 5984, in which case they’d need to run PouchDB Server on something else. This was the case for me, for instance.

garrensmith:
+1 thanks. I've merged it.

PR #858, opened 2017-02-20
Replication from local

garrensmith:
This improves the way we encode usernames and passwords. So if no
username and password are required it should still work and if the user
has special characters in the url it should still work.


nolanlawson:
LGTM, looking forward to replication working for pouchdb server!

millayr:
+1 from me

garrensmith:
Thanks merged.

nolanlawson:
Well, I probably should have tested this before saying :+1:. 😅 It actually breaks worse than before. Attempting to replicate from `http://examples.cloudant.com/animaldb` to a new local DB `animaldb` results in a replicator document looking like this:

```json
{
  "source": {
    "headers": {},
    "url": "http://examples.cloudant.com/animaldb"
  },
  "target": {
    "headers": {},
    "url": "http://localhost:6984/animaldb"
  },
  "create_target": true
}
```

I'll open a PR to fix it.

garrensmith:
Oh no. That's frustrating. 

On Mon, Feb 20, 2017 at 8:24 PM +0200, "Nolan Lawson" <notifications@github.com> wrote:












Well, I probably should have tested this before saying 👍. 😅 It actually breaks worse than before. Attempting to replicate from http://examples.cloudant.com/animaldb to a new local DB animaldb results in a replicator document looking like this:
{
  "source": {
    "headers": {},
    "url": "http://examples.cloudant.com/animaldb"
  },
  "target": {
    "headers": {},
    "url": "http://localhost:6984/animaldb"
  },
  "create_target": true
}


I'll open a PR to fix it.

PR #859, opened 2017-02-20
fix replication source/target url

nolanlawson:
Based on manual testing, this fixes the issue described in https://github.com/apache/couchdb-fauxton/pull/858#issuecomment-281150205 . I didn't run the Nightwatch tests out of laziness, though, so let's just see what Travis says.


nolanlawson:
Also I'll note that there's another issue somewhere in Fauxton where, when you run it in dev mode, it still posts replications to `localhost:5984` even if you try to point it to somewhere else... seems hardcoded somewhere. :confused:

garrensmith:
@nolanlawson could you close this since we have fixed it in pouchdb-server

PR #860, opened 2017-02-20
replace url construction with FauxtonAPI lookup

millayr:
By using `Fauxton.urls` we limit the number of times we determine/construct the URL and allow for it be overridden elsewhere.


garrensmith:
+1

garrensmith:
@millayr could you close this PR.

PR #861, opened 2017-02-21
fix deployment to work with styles.css

garrensmith:


Gruntfile.js
```diff
@@ -72,6 +72,7 @@ module.exports = function (grunt) {
       settings.template[key].app.i18n = i18n;
     });
 
+    console.log(settings.template);
```
millayr: if this was just for debugging, I'd recommend removing this

millayr:
+1

PR #862, opened 2017-02-22
handle username and passwords correctly with base64 encoding

garrensmith:


app/addons/replication/api.js
```diff
@@ -41,6 +42,38 @@ export const getAuthHeaders = (username, password) => {
...
+      username: '',
+      password: ''
+    };
+  }
+
+  const startIndex = url.startsWith("https") ? 8 : 7;
+  const rawCreds = url.slice(startIndex, index);
+  const colonIndex = rawCreds.indexOf(':');
+  const username = rawCreds.slice(0, colonIndex);
+  const password = rawCreds.slice(colonIndex + 1, rawCreds.length);
```
millayr: Minor note but I think you could combine all of these variable instantiations and assignments in to a single const statement.

app/addons/replication/api.js
```diff
@@ -41,6 +42,38 @@ export const getAuthHeaders = (username, password) => {
...
+  };
+};
+
+export const removeCredentialsFromUrl = (url) => {
+  const index = url.lastIndexOf('@');
+  if (index === -1) {
+    return url;
+  }
+
+  const protocal = url.startsWith("https") ? "https://" : 'http://';
```
millayr: protoc_o_l :)

millayr:
+1

garrensmith:
merged

PR #864, opened 2017-03-02
Update replication to work with scheduler api

garrensmith:
This is to support the new scheduler api that will land in CouchDB soon


justin-mcdavid-ibm:
1. I'm sorry for the confusion in the mockups, but I think we should keep the "New Replication" button in place for Replicator DB Activity, and add it to the _replicate Activity tab as well, even though both buttons will only be capable of triggering a replicator db job.  My mistake here; I think we want to try and maintain the cohesive pattern of "new" buttons above the table across the dashboard, and when it comes time to add "Advanced Replication" to the mix, we can do so still maintaining that pattern. Updated mockups below.

2. Hover help on the trashcan icon in the _replicate Activity section should not reference a document being deleted, but instead the job. “Delete job id 1342345234534534563456.”

3. Hovering over the state should result in the state time displaying.

4. _replicate Activity doesn’t seem to be picking up the type correctly. My continuous jobs are showing up as “one time.” 

5. The “Retrying” state should only be for for jobs that are actually being retried. Auth errors aren’t retried, are they? I’m seeing Retry for an error with message of “unauthorized: unauthorized to access or create database https://garrensmith.cloudant.com/tickle_production/”  which I believe is a “failed” job, right?

6. Not all Replicator DB jobs are displaying Start Times in the activity table, though the columns are still sorting on the start times in the replication docs.

7. I'd like to shorten our caveat, so that it remains a one-liner when we eventually constrain line-lengths for block text across the dashboard. Please change the caveat to:
"Active _replicate jobs are displayed.  Completed and failed jobs are not."

8. Are we going to be able to surface erroring/crashing states for _replicate Activity?  I see that in my earlier mockup, that I left the "State" column out, mostly because I figured that the "Start Time" one would suffice if state would only be showing running jobs. But perhaps I misunderstood. If we can pick up _troubled jobs' replication states, then we definitely should have the state column.




Here is the list of tweaks that I had in [FogBugz](https://cloudant.fogbugz.com/f/cases/81384/), if you'd like all comments consolidated here:

1. On the Replication form, make the URL-entry fields a consistent length. Default document IDs are longer than the current "Replication Document" field, as well, so let's make the field longer, consistent with the lenght of the URL fields. It looks like the remote URL fields are currenly longer (~346px) to match the length of the text string below. I'd be fine if all URL fields and the Rep Doc field were 400px.

2. On the Verify Deletion dialogue box, change the button text so that it accounts for singular and plural doc-deletions. Probably easiest would be to change the text to "Delete Document(s)", otherwise, you could have "Delete Document" for single deletions, and "Delete Documents" for multiple deletions.

3. Change embedded help pop-up text from center-justification to default (left) justification.

4. In the Replication-activity-section header, change "Replication" to "Replication Activity."

5. Breadcrumbs are unnecessary on the new/edit-replication pages, as clicking the main nav takes the user to the exact same place as clicking on "Replication" in the header. So, I think we should remove the "Replication" crumb.

6. Clicking on the edit-replication icon takes the user to the replication form page, where the header nav states "New Replication." "New" is a smidge confusing in the context of editing, but users are able to trigger entirely new replications from the page. So for both entirely "new" replications and replications being edited, lets change that header text to "Replication-Job Details".

7. Increase the margin between the two buttons on the "Fix Document Conflict" dialogue box to 20px. 

Thanks much for all your work on this.

replication - _replicate activity
replication - replicator db activity

garrensmith:
@justin-mcdavid-ibm I've made some of the changes. Others of them I have some comments:

> I'm sorry for the confusion in the mockups, but I think we should keep the "New Replication" button in place for Replicator DB Activity, and add it to the _replicate Activity tab as well, even though both buttons will only be capable of triggering a replicator db job. My mistake here; I think we want to try and maintain the cohesive pattern of "new" buttons above the table across the dashboard, and when it comes time to add "Advanced Replication" to the mix, we can do so still maintaining that pattern. Updated mockups below.

I'm a little confused on this. Do you still want a tab for the new replications? Can you do a quick mockup of how the `new replication page` now looks?

> Hovering over the state should result in the state time displaying.

We don't always get the state time returned. So its only displayed on hover if we get it. It seems that completed don't get the time.

> The “Retrying” state should only be for for jobs that are actually being retried. Auth errors aren’t retried, are they? I’m seeing Retry for an error with message of “unauthorized: unauthorized to access or create database https://garrensmith.cloudant.com/tickle_production/” which I believe is a “failed” job, right?

I agree. I followed the mock up comments of `"Retrying" shown for both "Error" state and "Crashing" state.`. The auth state is an Error state.

> Not all Replicator DB jobs are displaying Start Times in the activity table, though the columns are still sorting on the start times in the replication docs.

The start time isn't always returned.

justin-mcdavid-ibm:
Here's the mockup:

replication job configuration

So, no tab on this page. And instead of "Replication > New Replication" or the proposed "Replication-Job Details," I think "Job Configuration" in the header will be a better descriptor.  Because you can edit an existing job on this page, "New" isn't exactly correct (though I know an edited job is just a new job on the back of a stopped one).

I know you're mid-way through this, and the beta doesn't necessarily reflect the most recent work, so apologies if this is already on your radar, but the local source/target dropdown component doesn't yet match the length of the longer dropdown field.

Also, I realized that we can now remove some of the text from our Verify Deletion error messages, now that replicator and _replicate have separate tabs.  Also, we need to change some of the button text, as _replicate deletions don't involve deleting a document.

Here are the messages, varying on whether deletion is triggered through the bulk selector, or in the table-row trashcan icons: 

replication - verify single deletion dialogue
replication - verify deletions dialogue

Thanks for the work on this. It's looking great.

justin-mcdavid-ibm:
Hey Garren,

Lookin' good! That section is rather slick. I see a few lingering issues, and a handful of things to polish, and then I think we're good.

Something has changed broken alignment of the values in the replication form:
field alignment

And the width for the dropdown values is still not matching the width of the dropdown field (for both source and target local dbs):
dropdowns

The doc-conflict buttons could still use some extra space (20px) between them:
screen shot 2017-03-07 at 9 43 58 am

I noticed that the job-deleted banner message (and message in the notification center) is picking up more than just the replication number:

screen shot 2017-03-07 at 10 39 18 am

The length is such that I noticed a lack-of-wrapping bug in the notification center, which is something I'll open up a bug on.

screen shot 2017-03-07 at 11 00 45 am

Can we strip off the non-job-id information in that job name, and add a period to the end of the statement?


**Some polish and tweaks:**

With only one Action icon on the _replicate Activity table, I think it would be better to align it under the column header:

screen shot 2017-03-07 at 9 48 37 am

I realize that "everybody" permissions terminology is going to be confusing for those who've set permissions using the CL, where the user is "nobody." For the last paragraph in this hoverhelp, I'd like to change the statement to "If the remote database granted permissions to unauthenticated connections, you do not need to enter a username or password."
screen shot 2017-03-07 at 9 52 18 am

Let's align the (now shorter) caveats with the text in the first content column (Source) in the tables.  I think that would be 80px from the left.

screen shot 2017-03-07 at 9 59 16 am

screen shot 2017-03-07 at 10 02 08 am

It would be nice to swap out the white-check icon for the previously used replication icon:
screen shot 2017-03-07 at 10 04 27 am

screen shot 2017-03-07 at 10 06 30 am

And swap the white-check icon on the verify-deletion buttons for trashcan icons:

screen shot 2017-03-07 at 10 08 14 am

Thanks for your work on this.

garrensmith:
> Something has changed broken alignment of the values in the replication form:

This is actually a firefox issue. I've managed to make a work around.

> The length is such that I noticed a lack-of-wrapping bug in the notification center, which is something I'll open up a bug on.

I fixed it.

> Can we strip off the non-job-id information in that job name, and add a period to the end of the statement?

As far as I understand, that is the full replicate id. We cannot strip any part of it off.

justin-mcdavid-ibm:
Thanks for the work, Garren.  It looks great.

app/addons/replication/controller.js
```diff
@@ -186,25 +208,79 @@ export default class ReplicationController extends React.Component {
...
+      <TabElementWrapper>
+        {this.getTabElements()}
+      </TabElementWrapper>
+    );
+  }
+
   render () {
+    const { checkingAPI } = this.state;
+
+    if (checkingAPI) {
```
millayr: Is it worth switching this to `this.state.checkingAPI` to eliminate the previous line?  Doesn't look like we use `checkingAPI` anywhere else later on.

app/addons/replication/route.js
```diff
@@ -39,25 +41,31 @@ const ReplicationRouteObject = FauxtonAPI.RouteObject.extend({
   },
 
   defaultView: function (databaseName) {
-    const localSource = databaseName || '';
+    let localSource = databaseName || '';
```
millayr: can we switch this back to `const`?

app/addons/replication/stores.js
```diff
@@ -77,6 +110,23 @@ const ReplicationStore = FauxtonAPI.Store.extend({
...
   },
 
+  isReplicateInfoLoading () {
+    return this._fetchingReplicateInfo;
+  },
+
+  getReplicateInfo () {
+    return this._replicateInfo.filter(doc => {
+      return _.values(doc).filter(item => {
+        if (!item) {return null;}
```
millayr: Minor but can we change this to return `false`?  I feel like it'd be more readable when considering `filter` keeps/discards items based on boolean values.

app/addons/replication/actions.js
```diff
@@ -171,13 +298,20 @@ const deleteDocs = (docs) => {
   });
 };
 
-const getReplicationStateFrom = (id) => {
-  $.ajax({
-    url: `${app.host}/_replicator/${encodeURIComponent(id)}`,
-    contentType: 'application/json',
-    dataType: 'json',
+export const getReplicationStateFrom = (id) => {
```
garrensmith: No, that is correct. Its getting the replication state from a document with `id`

app/addons/replication/api.js
```diff
@@ -16,6 +16,30 @@ import app from '../../app';
...
+    newApiPromise = new FauxtonAPI.Promise((resolve) => {
+      fetch('/_scheduler/jobs', {
+        credentials: 'include',
+        headers: {
+            'Accept': 'application/json; charset=utf-8',
+          }
+        })
+      .then(resp => {
+        if (resp.status > 202) {
+          return resolve(false);
```
garrensmith: if we don't use the return here, then it once its outside the if statement it will then try and resolve the promise again with `true`. I could put the return after the resolve. It doesn't really matter

app/addons/replication/components/common-activity.js
```diff
@@ -0,0 +1,107 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+import React from 'react';
+import {DeleteModal} from './modals';
+
+export class BulkDeleteController extends React.Component {
```
garrensmith: This is used in the activity.js and replicateActivity.js

garrensmith: Actually good spot. This code is not used anymore. I've removed it.

app/addons/replication/controller.js
garrensmith: I think its a personal preference thing. I would prefer to leave as is.

millayr:
+1 after tests

PR #868, opened 2017-03-16
Update to webpack 2

garrensmith:
This upgrades to webpack 2. Webpack 2 makes it a little easier to do certain things. One of them is to split the bundle.js into three separate files:

* manifest.js -> webpack loader file
* vendor.js -> all required libraries
* bundle.js -> Fauxton code

Each file also has a a hash applied to it based on the file contents. This should improve caching and load times for Fauxton as the manifest file and vendor file will not change as often as the bundle file.

Webpack is now also generating the html file, this was previously done in grunt.


assets/index.underscore
```diff
@@ -19,11 +19,8 @@
   <meta name="viewport" content="width=device-width,initial-scale=1">
   <meta http-equiv="Content-Language" content="en" />
   <link rel="shortcut icon" type="image/png" href="dashboard.assets/img/couchdb-logo.png"/>
```
nolanlawson: awesome, can we make the favicon configurable too? this was another thing we needed for pouchdb-fauxton

webpack.config.release.js
```diff
@@ -49,51 +73,71 @@ module.exports = {
...
     {
-      test: /\.woff2(\?\S*)?$/,   loader: 'url?limit=10000&mimetype=application/font-woff2&name=dashboard.assets/fonts/[name].[ext]'
+      test: /\.woff2(\?\S*)?$/,   loader: 'url-loader?limit=10000&mimetype=application/font-woff2&name=dashboard.assets/fonts/[name].[ext]'
     },
     {
-      test: /\.ttf(\?v=\d+\.\d+\.\d+)?$/,    loader: 'url?limit=10000&mimetype=application/font-tff&name=dashboard.assets/fonts/[name].[ext]'
+      test: /\.ttf(\?v=\d+\.\d+\.\d+)?$/,    loader: 'url-loader?limit=10000&mimetype=application/font-tff&name=dashboard.assets/fonts/[name].[ext]'
     },
-    { test: /\.eot(\?v=\d+\.\d+\.\d+)?$/,    loader: 'file?name=dashboard.assets/fonts/[name].[ext]' },
-    { test: /\.swf(\?v=\d+\.\d+\.\d+)?$/,    loader: 'file?name=dashboard.assets/[name].[ext]' },
```
millayr: why no swf support in `webpack.config.release.js` but it is supported in `webpack.config.dev.js`?

millayr:
This is great!  Left a few comments @garrensmith

millayr:
+1 pending the minor comments/tweaks

test/test.config.underscore
```diff
@@ -12,7 +12,7 @@
 // License for the specific language governing permissions and limitations under
 // the License.
 //
-
+var Promise = require('bluebird');
```
garrensmith: Yes, it is required.

webpack.config.release.js
garrensmith: swf isn't actually needed in either. I'll remove it.

PR #880, opened 2017-03-22
filter out all design docs for replication activity

### LINKED ISSUE #879: Replication UI does not render when _replicator contains ddoc with filter function
If a user has created a ddoc in the `_replicator` db that defines a filter function for use with filtered replications, the UI fails to load.  It hangs with the loading bars animation.

Likely not that useful but here's the browser console output:
`Unhandled rejection TypeError: Failed to construct 'URL': Invalid URL`

garrensmith:
This fixes #879


millayr:
LGTM +1

PR #884, opened 2017-03-25
Variable Cleanup in Styles

justin-mcdavid-ibm:
I'm not entirely done with this effort, but I believe it's in a state where it can be merged in without breaking anything... Everything is compiling for me with no problems, and visual testing hasn't turned up anything terrible.

This effort is mainly around establishing consistent variables (but with varying values) between the two projects dashboard/fauxton, consistently using variables where appropriate, cleaning up unused variables, consistently styling buttons in a L1/L2/L3 scheme, and general changes for interaction consistency.  There are still some nails to hammer down on this, but I think it's currently in a state of mergability.


app/addons/documents/changes/components.react.jsx
```diff
@@ -270,7 +270,7 @@ class ChangeRow extends React.Component {
   render () {
     const { codeVisible } = this.state;
     const { change, databaseName } = this.props;
-    const jsonBtnClasses = 'btn btn-small' + (codeVisible ? ' btn-secondary' : ' btn-info');
+    const jsonBtnClasses = 'btn btn-small' + (codeVisible ? ' btn-secondary' : ' btn-secondary');
```
garrensmith: We can simplify this now:

`const jsonBtnClasses = 'btn btn-small btn-seconday'`

garrensmith:
This is merged in via #886

PR #889, opened 2017-03-28
Shown database doesn't exist notification

### LINKED ISSUE #888: Provide better behaviour with a database that doesn't exist
When a user tries to navigate to a database that doesn't exist provide them with the following issue:

A database that does not exist:
We display a "Database does not exist." error message in the banner and the notification panel, but the page appears to continue attempting to access the non existent database.

This needs to apply to query, changes, permissions, all_docs

garrensmith:
This fixes #888

### DIFF
--- a/app/addons/documents/index-results/actions.js
+++ b/app/addons/documents/index-results/actions.js
@@ -50,19 +50,22 @@ export default {
 
     if (!options.collection.fetch) { return; }
 
-    return options.collection.fetch({reset: true}).then(function () {
+    return options.collection.fetch({reset: true}).then(() => {
       this.resultsListReset();
       this.sendMessageNewResultList(options);
 
-    }.bind(this), function (collection, _xhr) {
+    }, function (collection, _xhr) {
       //Make this more robust as sometimes the colection is passed through here.
       var xhr = collection.responseText ? collection : _xhr;
       var errorMsg = 'Bad Request';
 
       try {
-        var responseText = JSON.parse(xhr.responseText);
+        const responseText = JSON.parse(xhr.responseText);
         if (responseText.reason) {
           errorMsg = responseText.reason;
+        }
+        if (responseText.error && responseText.error === 'not_found') {
+          return;
         }
       } catch (e) {
         console.log(e);
@@ -72,7 +75,8 @@ export default {
         msg: errorMsg,
         type: "error",
         clear:  true
-     });
+      });
+
     });
   },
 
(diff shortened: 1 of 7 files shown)


app/addons/documents/index-results/actions.js
```diff
@@ -50,20 +50,23 @@ export default {
 
     if (!options.collection.fetch) { return; }
 
-    return options.collection.fetch({reset: true}).then(function () {
+    return options.collection.fetch({reset: true}).then(() => {
       this.resultsListReset();
       this.sendMessageNewResultList(options);
 
-    }.bind(this), function (collection, _xhr) {
+    }, function (collection, _xhr) {
```
millayr: `(collection, _xhr) => {`

app/addons/documents/index-results/actions.js
```diff
@@ -50,20 +50,23 @@ export default {
...
       var errorMsg = 'Bad Request';
 
       try {
-        var responseText = JSON.parse(xhr.responseText);
+        const responseText = JSON.parse(xhr.responseText);
         if (responseText.reason) {
           errorMsg = responseText.reason;
         }
+        if (responseText.error && responseText.error === 'not_found') {
+          return;
```
millayr: Won't this bail out of the function and skip the `addNotification` call?  Is that your intent here?

app/addons/documents/mango/__tests__/mango.actions.test.js
```diff
@@ -0,0 +1,60 @@
...
+              error: 'not_found'
+            }
+          });
+        },
+      }
+    };
+
+    Actions.getIndexList(options);
+    process.nextTick(() => {
+      assert.ok(notificationSpy.calledOnce);
```
millayr: I've been trying to use Jest's `expect` when doing these checks.

app/addons/documents/sidebar/__tests__/sidebar.actions.test.js
```diff
@@ -0,0 +1,60 @@
...
+              error: 'not_found'
+            }
+          });
+        },
+      }
+    };
+
+    Actions.newOptions(options);
+    process.nextTick(() => {
+      assert.ok(notificationSpy.calledOnce);
```
millayr: similar to above, I've been using `expect` for these.  We should probably decide on a common approach here

app/addons/documents/index-results/actions.js
garrensmith: Yes, that is intentional. We use the check in the sidebar action to validate if the database exists or not. If this shows a message here we will get two notifications that say the same thing.

PR #893, opened 2017-03-30
consistent error message for views and doc

### LINKED ISSUE #890: Warning notification for missing document in database
A document that does not exist in a database that does exist:
We display an error message, and then do a redirect to the _all_docs of the existing database.

A document that does not exist in a database that does not exist:
We display an error message that the database does not exist, there is a redirect to the non-existent database's _all_docs page, where the secondary nav and doc-jump component fail to load, but the document list can be populated with documents from previously loaded databases.

If a user attempts to access anything in a document (including design documents) that doesn't exist:
Redirect to the _all_docs view for the database, and throw and error of "The <doc_name> document does not exist."

garrensmith:
This fixes #890

### DIFF
--- a/app/addons/documents/index-results/actions.js
+++ b/app/addons/documents/index-results/actions.js
@@ -49,6 +49,7 @@ export default {
     this.clearResults();
 
     if (!options.collection.fetch) { return; }
+    console.log(options);
 
     return options.collection.fetch({reset: true}).then(() => {
       this.resultsListReset();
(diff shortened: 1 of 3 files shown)


app/addons/documents/index-results/actions.js
```diff
@@ -63,8 +63,9 @@ export default {
         if (responseText.reason) {
           errorMsg = responseText.reason;
         }
-        if (responseText.error && responseText.error === 'not_found') {
-          return;
+
+        if (responseText.reason && responseText.reason === 'missing_named_view') {
+          errorMsg = 'The <index_name> <index_type> does not exist.';
```
millayr: I'd suggest coming up with a more readable error message for this

millayr:
+1 pending travis and addressing my comment

PR #895, opened 2017-04-10
Removing persistent scrollbars from navbar

millayr:
@justin-mcdavid-ibm reported his browser displayed vertical and horizontal scrollbars for the new navbar implementation even when there was no overflowed content.

This css tweak limits scrollbar rendering to vertical only if there is overflow.


rnewson:
+1

PR #897, opened 2017-04-10
Prevent navbar link underlines

millayr:
@justin-mcdavid-ibm noticed that the links in the lefthand navbar temporarily receive an underline when they enter a "focus" state.  This PR prevents the temporary underline.


justin-mcdavid-ibm:
+1

PR #904, opened 2017-04-24
Be explicit about version of Node/npm required

wohali:
Recently I traced many Jenkins CI failures to using an outdated version of Node/NPM (Node 4, npm 2). @garrensmith and @janl said this is outdated and that Node 6 / npm 3 was required.

This makes the dependency on these versions explicit.


garrensmith:
Thanks @wohali

PR #906, opened 2017-04-26
Toolbar redesign for all docs

millayr:
This is an incremental change to increase fauxton's info density per some design suggestions from @justin-mcdavid-ibm.

- We now default to a "Metadata" view when possible.  This condenses the results from a query (sans include_docs) in to a table layout.
- Table view automatically triggers `include_docs=true`.
- JSON view automatically triggers `include_docs=true`.
- A new "Create Document" is now included in the toolbar to be consistent with other design changes (see Replication for an example).

See new screenshots below:
image

image

image

Future plans include migrating the Query Options modal in to this new toolbar and creating a quick filter input (similar to Replication).  The secondary sidebar will also eventually be removed and replaced by tabs (again, see Replication).


garrensmith:
This is really cool. Some feedback:

* Keep the header section when clicking between sections and have the loader only for the content
* Make it a single click when I click on a document to take me to view that document
* Make the json view have `include_docs=true` by default

app/addons/documents/base.js
```diff
@@ -109,6 +109,10 @@ FauxtonAPI.registerUrls('view', {
 
   fragment: function (database, designDoc, viewName) {
     return 'database/' + database + designDoc + '/_view/' + viewName;
+  },
+
+  query: function (database, designDoc, viewName, query) {
+    return 'database/' + database + '/_design/' + designDoc + '/_view/' + viewName + getQueryParam(query);
```
garrensmith: Is the designDoc and view name encoded elsewhere? will it work with `/_designDoc/view/name`?

app/addons/documents/index-results/index-results.components.js
```diff
@@ -172,16 +160,19 @@ var TableRow = React.createClass({
...
+  },
+
   render: function () {
     var i = this.props.index;
     var docContent = this.props.el.content;
     var el = this.props.el;
 
     return (
-      <tr key={"tableview-content-row-" + i}>
+      <tr key={"tableview-content-row-" + i} onDoubleClick={this.onDoubleClick}>
```
garrensmith: I would prefer onClick here. DoubleClick doesn't feel intuitive

app/addons/documents/index-results/stores.js
```diff
@@ -472,19 +474,9 @@ Stores.IndexResultsStore = FauxtonAPI.Store.extend({
...
-    var hasBulkDeletableDoc;
-    var database = this.getDatabase();
-    var isView = !!this._collection.view;
-
     // softmigration remove backbone
-    var data;
-    var collectionType = this._collection.collectionType;
-    data = this._filteredCollection.map(function (el) {
+    const collectionType = this._collection.collectionType;
+    let data = this._filteredCollection.map(function (el) {
```
garrensmith: this could become `let data = this._filteredCollection.map(el => fixDocId.....)`

garrensmith:
I'm also not sure if we should have the `Create Document` button on the view results page or the Mango page. But I'll defer that to @justin-mcdavid-ibm

garrensmith:
This is making progress. I've picked up few bugs with pagination.
If I have a database with > 20 docs and I'm on the metadata page. I then click the next page button.
It then paginates correctly to the next list of documents. If I then click the json tab it shows me the documents from the first page instead of the second page. If I then click back to the metadata tab it shows the correct documents but the pagination arrows don't allow me to go back.
Same issue with I click on the table view

app/addons/documents/index-results/components/pagination/TableControls.js
```diff
@@ -0,0 +1,64 @@
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
```
garrensmith: All our other component files are all lowercase. So this should be `tablecontrols`. Can you change that for all the files. Otherwise its going to get a bit messy

app/addons/documents/index-results/components/results/IndexResults.js
```diff
@@ -0,0 +1,97 @@
...
+  constructor (props) {
+    super(props);
+    const {
+      fetchAllDocs,
+      fetchParams,
+      queryOptionsParams,
+    } = this.props;
+
+    // now get the docs!
+    fetchAllDocs(fetchParams, queryOptionsParams);
```
garrensmith: Isn't it better to call this on `componentDidMount`?

app/addons/documents/index-results/components/results/ResultsScreen.js
```diff
@@ -0,0 +1,132 @@
...
+    let mainView = null;
+
+    if (this.props.isLoading) {
+      mainView = <div className="loading-lines-wrapper"><LoadLines /></div>;
+    } else if (!this.props.hasResults) {
+      mainView = <NoResultsScreen text={this.props.textEmptyIndex}/>;
+    } else if (this.props.selectedLayout === Constants.LAYOUT_ORIENTATION.JSON) {
+      mainView = this.getDocumentStyleView();
+    } else {
+      mainView = this.getTableStyleView();
```
garrensmith: Instead of this else, rather do `let mainView = this.getTableStyleView();`

app/addons/documents/index-results/helpers/json-view.js
```diff
@@ -0,0 +1,34 @@
...
+export const getJsonViewData = (docs, { databaseName, typeOfIndex }) => {
+  const stagedResults = docs.map((doc) => {
+    return {
+      content: JSON.stringify(doc, null, ' '),
+      id: doc.id, //|| doc.key.toString(),
+      _rev: doc._rev || (doc.value && doc.value.rev),
+      header: doc.id, //|| doc.key.toString(),
+      keylabel: 'id', //doc.isFromView() ? 'key' : 'id',
+      url: doc.id ? getDocUrl('app', doc.id, databaseName) : null,
+      isDeletable: true, //TODO: determine logic here
```
garrensmith: Is this still a `TODO`?

app/addons/documents/index-results/helpers/shared-helpers.js
```diff
@@ -0,0 +1,75 @@
...
+  if (context === undefined) {
+    context = 'server';
+  }
+
+  // new without id make a POST to the DB and not a PUT on a DB
+  let safeId = app.utils.getSafeIdForDoc(id);
+  if (!safeId) {
+    safeId = '';
+  }
+  const safeDatabaseName = app.utils.safeURLName(databaseName);
```
garrensmith: Could we url encode it instead of using `safeURLName`?

app/addons/documents/index-results/helpers/shared-helpers.js
```diff
@@ -0,0 +1,75 @@
...
+  if (docType === 'MangoIndex') {
+    return doc.type !== 'special';
+  }
+  const result = (doc._id || doc.id) && (doc._rev || (doc.value && doc.value.rev));
+  return !!result;
+};
+
+const hasBulkDeletableDoc = (docs, docType) => {
+  // use a for loop here as we can end it once we found the first id
+  for (let i = 0; i < docs.length; i++) {
```
garrensmith: You can also use `array.find`

app/addons/documents/index-results/index-results.components.js
```diff
@@ -402,23 +356,30 @@ var ResultsScreen = React.createClass({
...
 
     if (this.props.isLoading) {
-      loadLines = <LoadLines />;
+      mainView = <div className="loading-lines-wrapper"><LoadLines /></div>;
+    } else if (!this.props.hasResults) {
+      mainView = <NoResultsScreen text={this.props.textEmptyIndex}/>;
+    } else if (this.props.selectedLayout === Constants.LAYOUT_ORIENTATION.JSON) {
+      mainView = this.getDocumentStyleView();
+    } else {
+      mainView = this.getTableStyleView();
```
garrensmith: Again this else can be removed and `let mainView = this.getTableStyleView();`

app/addons/documents/index-results/apis/base-api.js
```diff
@@ -0,0 +1,95 @@
...
+
+export const newSelectedDocs = (selectedDocs = []) => {
+  return {
+    type: ActionTypes.INDEX_RESULTS_REDUX_NEW_SELECTED_DOCS,
+    selectedDocs: selectedDocs
+  };
+};
+
+export const selectDoc = (doc, selectedDocs) => {
+  const indexInSelectedDocs = selectedDocs.findIndex((selectedDoc) => {
```
millayr: We need to remove docs from the array when a user has deselected it from their list of results.  I've added some comments better describe what the function is trying to achieve.

app/addons/documents/index-results/apis/base-api.js
```diff
@@ -0,0 +1,95 @@
...
+    selectedDocs.splice(indexInSelectedDocs, 1);
+  } else {
+    doc._deleted = true;
+    selectedDocs.push(doc);
+  }
+
+  return newSelectedDocs(selectedDocs);
+};
+
+export const bulkCheckOrUncheck = (docs, selectedDocs, allDocumentsSelected) => {
```
millayr: Pagination makes this tricky.  When you bulk select all docs on the page, we add those docs to the running `selectedDocs` array.  When you bulk deselect all docs on the page, we remove them BUT we keep all of the docs you've selected on previous pages of results.  So it's not really an all-or-nothing thing.

app/addons/documents/index-results/components/pagination/TableControls.js
millayr: I made this same comment to Robert back in the day when he did the Redux refactor for the permissions section.  He mentioned the style pattern in the JS community was shifting to uppercase for components.  I could go either way.  Feel strongly I switch it to lowercase?

app/addons/documents/index-results/components/results/ResultsScreen.js
millayr: I'd prefer to leave this as is.  By initializing `mainView` to `this.getTableStyleView()`, we do extra calculations when the end result is not the table view.

app/addons/documents/index-results/index-results.components.js
millayr: Same as my comment above ^

garrensmith:
+1
This is fantastic work. Thanks @millayr

Issue #909, opened 2017-04-28
Adding view to empty ddoc fails


millayr:
Fauxton throws an exception when attempting to save a new view to a ddoc that does not already have the 'views' attribute.  Fauxton incorrectly assumes that 'views' will exist if the doc is a ddoc.

This was originally reported in https://github.com/cloudant/dashboard/issues/774

millayr:
Fix included in https://github.com/apache/couchdb-fauxton/pull/911

PR #911, opened 2017-05-01
Handle view creation on empty ddoc

millayr:
Addresses: https://github.com/apache/couchdb-fauxton/issues/909

In this PR:
* Injecting 'views' and 'language' in to ddoc when it doesn't exist
* Fix overwrite bug when creating a second view in a ddoc.  I came across this while testing the fix above.  The new logic only tries to overwrite a view if and only if the user has decided to **edit** a view.

### DIFF
--- a/app/addons/documents/index-editor/__tests__/actions.test.js
+++ b/app/addons/documents/index-editor/__tests__/actions.test.js
@@ -58,15 +58,11 @@ describe('Index Editor Actions', function () {
 
     it('saves design doc if has other views', function () {
       designDoc.save = function () {
-        var promise = $.Deferred();
-        promise.resolve();
-        return promise;
+        return FauxtonAPI.Promise.resolve();
       };
       var saveSpy = sinon.spy(designDoc, 'save');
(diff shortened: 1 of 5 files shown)


app/addons/documents/index-editor/__tests__/actions.test.js
```diff
@@ -109,26 +101,48 @@ describe('Index Editor Actions', function () {
...
-      assert.ok(spy.calledOnce);
+    it('saves design doc if it has no view section', function () {
+      const ddoc = { _id: designDocId };
+      const ddocModel = new Documents.Doc(ddoc, { database: database });
+
+      ddocModel.setDdocView('testview', '() => {}', '() => {}');
+      assert.deepEqual(ddocModel.get('views'), {
+        testview: {
+          map: '() => {}',
+          reduce: '() => {}'
```
garrensmith: Could you add a check that it adds `language: javascript` as well.

millayr: done!

PR #833, opened 2017-01-16
remove backbone from auth

timmak:
In process of removing jQuery this also will require remove backbone rather make a few small PRs then on big one.

PR #923, opened 2017-05-10
Clean up Auth section

garrensmith:
Remove backbone and clean up auth. This is continued the work that @timmak did in #833 

I've updated the PR to work with the current codebase. I now want to make some extra changes and possibly move to redux.


app/addons/auth/actions.js
```diff
@@ -10,148 +10,157 @@
...
+} = ActionTypes;
 
-var errorHandler = function (xhr, type, msg) {
-  msg = xhr;
-  if (arguments.length === 3) {
-    msg = xhr.responseJSON.reason;
-  }
+const nodesStore = ClusterStore.nodesStore;
 
+function errorHandler({ message }) {
```
millayr: Consider arrow function style here

app/addons/auth/actions.js
```diff
@@ -10,148 +10,157 @@
...
   FauxtonAPI.addNotification({
-    msg: msg,
-    type: 'error'
+    msg: message,
+    type: "error"
   });
 };
 
+function validate(...predicates) {
```
millayr: same as above

app/addons/auth/actions.js
```diff
@@ -10,148 +10,157 @@
...
+};
 
-function createAdmin (username, password, loginAfter) {
+export const changePassword = (username, password, passwordConfirm) => dispatch => {
+  if (!validatePasswords(password, passwordConfirm)) {
+    return errorHandler({message: app.i18n.en_US['auth-passwords-not-matching']});
+  }
   var nodes = nodesStore.getNodes();
-  var promise = FauxtonAPI.session.createAdmin(username, password, loginAfter, nodes[0].node);
+  Api.createAdmin({
```
millayr: Why call `createAdmin` on a password change?  If this is by design, can we consider renaming this function or at least leave a comment explaining why we're doing this?

app/addons/auth/actions.js
```diff
@@ -10,148 +10,157 @@
...
-}
+};
 
-function selectPage (page) {
-  FauxtonAPI.dispatch({
-    type: ActionTypes.AUTH_SELECT_PAGE,
-    page: page
-  });
+// simple authentication method - does nothing other than check creds
+export function authenticate(username, password, onSuccess) {
```
millayr: Consider being consistent with arrow functions here.  I'm perhaps overly picky when it comes to consistency :)

app/addons/auth/stores/changepasswordstore.js
```diff
@@ -0,0 +1,58 @@
+// import FauxtonAPI from "../../../core/api";
```
millayr: I'd delete this file

app/addons/auth/components/changepasswordform.js
```diff
@@ -0,0 +1,97 @@
...
+          <p>
+            <button type="submit" className="btn btn-primary">Change</button>
+          </p>
+        </form>
+      </div>
+    );
+  }
+}
+
+export default connect(
```
millayr: I'd like to see us stick with the Container/Presentational Component pattern defined here:  http://redux.js.org/docs/basics/UsageWithReact.html#implementing-container-components.

It'll keep all of the fauxton modules consistent and a bit easier to piece together what parts of the state we're subscribing to/which events we want to dispatch.

app/addons/auth/components/createadminform.js
```diff
@@ -0,0 +1,116 @@
...
+CreateAdminForm.propTypes = {
+  loginAfter: React.PropTypes.bool.isRequired
+};
+
+CreateAdminForm.defaultProps = {
+  loginAfter: false
+};
+
+
+export default connect(
```
millayr: same as above

app/addons/auth/session.js
```diff
@@ -0,0 +1,135 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+
+import { uniqueId, each } from "lodash";
+import Api from './api';
+import Promise from 'bluebird';
+import app from '../../app';
+import {includes} from 'lodash';
```
millayr: this can be combine with the first import from `lodash` above

app/addons/auth/session.js
```diff
@@ -0,0 +1,135 @@
...
+  matchesRoles(roles = []) {
+    if (roles.length === 0) {
+      return true;
+    }
+
+    const numberMatchingRoles = _.intersection(this.userRoles(), roles).length;
+    return numberMatchingRoles > 0;
+  }
+
+  createAdmin(username, password, login, node) {
```
millayr: Where do we use this `createAdmin`?  It seems like the `this` context is expecting something different from where this function is defined.

app/addons/auth/session.js
```diff
@@ -0,0 +1,135 @@
...
+      .then(() => Api.createAdmin({ name: this._user.name, password, node }))
+      .then(
+        () =>
+          login
+            ? this.login(username, password)
+            : this.fetchUser({ forceFetch: true })
+      );
+  }
+
+  changePassword(password, confirmedPassword, node) {
```
millayr: same as above

app/addons/documents/changes/components.react.jsx
```diff
@@ -0,0 +1,352 @@
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
```
millayr: I'd rename the file so we don't introduce another `jsx` extension

app/addons/replication/__tests__/newreplication.test.js
```diff
@@ -21,7 +21,7 @@ const {assert, restore}  = utils;
 
 describe('New Replication Component', () => {
 
-  FauxtonAPI.session.triggerError = () => {};
+  //FauxtonAPI.session.triggerError = () => {};
```
millayr: remove this line

app/core/authentication.js
```diff
@@ -0,0 +1,43 @@
...
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+
+import FauxtonAPI from "./base";
+
+export const authenticate = (session, roles) => {
+  if (session.isAdminParty()) {
+    return true;
+  } else if (session.matchesRoles(roles)) {
```
millayr: Consider refactoring this `else if` in to an OR in the first `if` since both return true

app/core/router.js
```diff
@@ -10,8 +10,8 @@
 // License for the specific language governing permissions and limitations under
 // the License.
 
-import FauxtonAPI from "./base";
-import Auth from "./auth";
+//import FauxtonAPI from "./base";
```
millayr: remove this line

app/core/router.js
```diff
@@ -108,12 +103,12 @@ export default Backbone.Router.extend({
...
 
-  triggerRouteEvent: function (event, args) {
-    if (this.activeRouteObject) {
-      var eventArgs = [event].concat(args);
-
-      this.activeRouteObject.trigger.apply(this.activeRouteObject, eventArgs);
-      this.activeRouteObject.renderWith(eventArgs, FauxtonAPI.masterLayout, args);
-    }
-  }
+  // triggerRouteEvent: function (event, args) {
```
millayr: remove these lines

millayr:
Oh, other general comment is consistency with ES6 arrow functions :)

app/addons/auth/actions.js
garrensmith: Updating a password is the same as creating an new admin because its via the _config endpoint. I'll add a comment.

app/addons/auth/components/changepasswordform.js
garrensmith: I don't really follow what you mean here? Normally this component would be the container with more "dumb" components inside but because its a simple section I decided not to add any more components. In terms of using redux, I'm not using a store/reducer at all and rather keeping the state in the component. I gave this quite a bit of thought and couldn't find a good reason to use a reducer here since its just capturing some basic info.

app/addons/auth/session.js
garrensmith: Good catch. We are not using that function anymore.

app/core/authentication.js
garrensmith: Normally I would agree with you. But I would prefer to keep them separate since they involve different checks. I feel this makes it a little more obvious.

app/addons/auth/components/changepasswordform.js
millayr: Alright, I agree the state info is limited.  I noticed you have functions in `actions.js` that still return or dispatch events.  Without a reducer to handler those events, can we remove that bit of extra work?

millayr:
+1 from me.  nice work!

millayr:
+1

garrensmith:
Thanks @timmak for the initial work.

timmak:
thanks to @garrensmith and everyone that helped get this through sorry I wasn't able to finish this

dbgreen:
+1

PR #924, opened 2017-05-15
Fix table for url with special char in password

garrensmith:
Fixes an issue where the replication activity table wouldn't render if a
url had a password with a special character in it


millayr:
Nice, +1

PR #930, opened 2017-06-14
Json view

garrensmith:
This removes the api button and dropdown tray and simplifies it to use a json link and a button for help.


millayr:
This looks good!  I'm seeing some minor styling behavior after clicking on either buton, JSON or docs.  The button is still red/active.  Here are the steps:
1. Click on {} JSON
2. Return back to the fauxton tab
3. Note that the {} JSON button is still highlight red (and it has an underline)
4. Repeat for the docs icon.

Note I'm using Chrome 58.

image

millayr:
Nice +1

PR #933, opened 2017-07-17
Update for changes to _cluster_setup endpoint

### LINKED ISSUE #874: Single Node Setup on Fauxton forces non-127.0.0.1 IP
Single node setups might just want to bind to 127.0.0.1, this could be fore security on the db server, or on a dev machine.
Currently, the setup wizard for the single-node case forces an ip address other than 127.0.0.1, we should not do that 

Taken from https://issues.apache.org/jira/browse/COUCHDB-3139

wohali:
These changes add support for the new enable_single_node action,
which tolerates binding a node to 127.0.0.1.

They also add support for the new mandatory node_count parameter
when setting up a cluster. This parameter in the API call actually
sets the ini `[cluster] n` value, so it is set to min(3, <value provided>)
based on recommendations from @rnewson and personal experience.

Closes #874.

Addresses (but does not close) apache/couchdb#593

@garrensmith We're going to need a new tagged Fauxton release once this lands, if you please. 🤓 

## Checklist

- [X] Changes are covered by tests;


app/addons/setup/setup.actions.js
```diff
@@ -87,7 +88,13 @@ export default {
         return FauxtonAPI.session.login(username, password);
       })
       .then(function () {
```
michellephung: indent

app/addons/setup/setup.actions.js
```diff
@@ -87,7 +88,13 @@ export default {
...
-        return this.finishClusterSetup('CouchDB is set up!');
+      FauxtonAPI.addNotification({
```
michellephung: indent inside the function

app/addons/setup/setup.js
```diff
@@ -22,11 +22,41 @@ var ConfirmButton = ReactComponents.ConfirmButton;
...
 
   render: function () {
+    var nodetype;
+
+    if (this.state.clusterState === 'cluster_finished') {
+      nodetype = 'clustered';
+    } else if (this.state.clusterState === 'single_node_enabled') {
+      nodetype = 'single';
+    } else {
+      nodetype = 'unknown state';
```
michellephung: you can put this if-else in a function to clean up the render function.

like 

var nodetype = this.getNodeType();

michellephung:
code looks good, but i'm not sure how to test it

I usually run fauxton with a load balancer in front

wohali:
You need to run Fauxton without a load balancer, and the very latest CouchDB, to test it.

app/addons/setup/setup.stores.js
```diff
@@ -91,6 +92,14 @@ var SetupStore = FauxtonAPI.Store.extend({
     return this._setupNode.bindAddress;
   },
 
+  setNodeCountForSetupNode: function (options) {
+    this._setupNode.nodeCount = Math.min(options.value, 3);
```
wohali: This sets the cluster n value, so it should set n = 3.

app/addons/setup/setup.js
```diff
@@ -179,6 +238,10 @@ var SetupMultipleNodesController = React.createClass({
     SetupActions.setPortForSetupNode(e.target.value);
   },
 
+  alterNodeCount: function (e) {
+    SetupActions.setNodeCount(e.target.value);
```
wohali: I was just copying how all of the other functions were doing it. If I change this one, do I have to change all the others? Really out of my depth here

app/addons/setup/setup.js
```diff
@@ -66,6 +99,32 @@ var SetupCurrentAdminPassword = React.createClass({
...
+          onChange={this.handleNodeCountChange}
+          placeholder="Value of cluster n"
+          type="text" />
+      </div>
+    );
+  }
+});
 
 var SetupOptionalSettings = React.createClass({
   getInitialState: function () {
```
wohali: I didn't change this code, I can fix it but I really don't understand it

app/addons/setup/setup.js
```diff
@@ -66,6 +99,32 @@ var SetupCurrentAdminPassword = React.createClass({
...
 
+var SetupNodeCountSetting = React.createClass({
+  getInitialState: function () {
+    return {
+      nodeCountValue: this.props.nodeCountValue
+    };
+  },
+
+  handleNodeCountChange: function (event) {
+    this.props.onAlterNodeCount(event);
```
wohali: Again was just copying this from other functions in this function...will try to change it

wohali:
You probably need to update your deps if you're getting that error. From the root checkout directory: `rebar get-deps update-deps` should do it.

garrensmith:
So does that mean it's not possible to setup a cluster larger than 3 nodes?

________________________________

wohali:
No, it means that clusters with many more nodes will still get `local.ini` set to `[cluster] n = 3` which is appropriate. The setting doesn't limit the number of nodes you can add to a cluster, just caps the setting of the `n` variable which is appropriate in the vast majority of cases. I doubt very much that Cloudant has any clusters with n>3.

janl:
Trying to distill an IRC conversation between @wohali & @michellephung:

As far as I understand, the situation is this: fauxton runs on admin party couchdb, user does /_cluster_setup 'enable cluster', after which CouchDB doesn’t accept unauthenticated requests any more, in order to not have folks go to login right away, and since in the cluster setup page we have the plaintext adminuser/password, the cluster setup fauxton code does a session.login() with these credentials.

That’s before this PR, but subsequently, this PR also doesn't work as expected.

I haven’t had time to verify that current Fauxton on CouchDB master fails on cluster setup, but that’d be the litmus test.

janl:
> I haven’t had time to verify that current Fauxton on CouchDB master fails on cluster setup, but that’d be the litmus test.

If that’s broken, we need to fix that prior to 2.1. regardless of this PR.

michellephung:
i'd like to test this some more if possible. I don't mind if it goes out though, it works, it's just wonky when it logs you out unexpectedly.

wohali:
FYI I'm not getting the logout error testing locally with Firefox on Windows. Perhaps this is a Chrome vs. FF thing?

Also, we have to land this PR because it adds the node count parameter to clustered-mode, which is now a required parameter for the API endpoint. Without the change landing, the entire Setup wizard in Fauxton is broken.

wohali:
Here's how I test this PR locally:

1. Start with a brand new checkout of couchdb, to be 100% sure you don't have old dependencies lying around: `git checkout https://github.com/apache/couchdb`
1. `cd couchdb && ./configure -c`
1. `cd src/fauxton && git checkout 874-update-setup-wizard`
1. `cd ../.. && make couch` and wait about 6 minutes.
1. `dev/run -n 1 --with-admin-party-please --no-join` or, for a cluster, use `-n 3`. When running on a remote host, I port forward 15984/25984/35984 to my laptop. Testing requires the nodes not be joined and that no load balancer is used.

I'd like it if at least one other person can confirm this works in both cluster and single-node mode before we merge at a bare minimum.

wohali:
OK, I've fixed the re-login issue, which was indeed broken for both single-node and cluster setup paths. I've also validated that the wizard works for single-node and cluster-mode.

To test single node, after completing the steps above:

* Setup > Configure Single Node
* l/p admin/password
* Bind to 127.0.0.1
* Port 15984
* Press the green button - everything should work

To test a 3-node cluster setup:

Be sure to run `make devclean` before restarting `dev/run` to wipe the `_nodes` database.

* Setup > Configure a cluster
* l/p admin/password
* Bind to 0.0.0.0
* Port 15984
* 3 nodes

Then, below:
* Remote host 127.0.0.1
* Bind address 0.0.0.0
* Port 25984
* press Add Node
* Now change port to 35984
* press Add Node
* press Configure Cluster

michellephung:
+1

PR #940, opened 2017-07-26
Use application/x-www-form-urlencoded for _session

garrensmith:
Use x-www-form-urlencoded instead of application/json so that this will
work with older versions of CouchDB that don't support application/json


garrensmith:
Tests passing show that login/logout is working.

rnewson:
I don't know JS too well but +1 in the spirit of solidarity. :)

PR #946, opened 2017-08-02
Fixes documents per page issue with mango

garrensmith:
This fixes the issue if you try to change the documents per page on the
mango query page before you have done a query it would fail

## Testing recommendations

Click on the query button in a database. Click update documents per page and change the value. Then click `Run Query`. This should all work and return results.


Antonio-Maranhao:
+1. Looks good to me.

PR #948, opened 2017-08-04
Fix sidebar expand/collapse when design doc name has special chars

Antonio-Maranhao:
## Overview

Fixes two issues:
 - When a design doc name has special chars (e.g. $, #), the item on the side bar was showing the encoded value (e.g. `my-$doc` is displayed as `my-%23doc`).
 - Clicking the design doc item would throw and the item was not expanded.

## Testing recommendations

- In a database click to create a new view. Name the design doc `my-design-$-#doc`, and the view `my-$-#view`.
   - Both names should display correctly on the sidebar
   - Click the design doc to expand/collapse it
   - Make sure after clicking the item is highlighted (text turns red)
- Load the view using the URL
   - Type the URL directly in the browser (e.g. `http://<hsot>:8001/#/database/stuff/_design/my-design-%24-%23doc/_view/my-%24-%23view`) and press Enter.

## GitHub issue number

https://github.com/cloudant/dashboard/issues/822


Antonio-Maranhao:
@garrensmith ready for review.

app/helpers.js
```diff
@@ -57,4 +57,13 @@ Helpers.getDateFromNow = function (timestamp) {
   return moment(timestamp, 'X').fromNow();
 };
 
+/**
+ * Escapes any characters that have special meaning in JQuery selectors.
+ * @param {string} selector
+ * @return {string}
+ */
+Helpers.escapeJQuerySelector = function (selector) {
```
garrensmith: What is the advantage of using this versus `decodeURIComponent`?

app/addons/documents/routes-index-editor.js
```diff
@@ -53,7 +53,7 @@ const IndexEditorAndResults = BaseRoute.extend({
     const params = this.createParams(),
           urlParams = params.urlParams,
           docParams = params.docParams,
-          decodeDdoc = decodeURIComponent(ddoc),
+          decodeDdoc = ddoc,
```
garrensmith: This variable name is confusing since its now not the decoded design doc.

app/helpers.js
```diff
@@ -57,4 +57,13 @@ Helpers.getDateFromNow = function (timestamp) {
   return moment(timestamp, 'X').fromNow();
 };
 
+/**
+ * Escapes any characters that have special meaning in JQuery selectors.
```
garrensmith: We haven't used comments like this before. I'm not super convinced we need them. Could you take these out. 
We can have a bigger discussion around comments like this and then look to add them in multiple places.

app/helpers.js
Antonio-Maranhao: It handles CSS metachars and escapes them with `\` instead of `%`.
The doc ID is used as ID for some `<div>`. When the ID has chars like `#` or `.`,  JQuery fails because it considers those metachars. E.g. if you want to find element with ID `#foo.bar` your query would be `$('#\#foo\.bar')`

Antonio-Maranhao:
@garrensmith nightwatch test added.

PR #951, opened 2017-08-10
Set safeUrlName to always encode a name

garrensmith:
Sometimes a document can be double encoded before saved into CouchDB. We need to support that use case. That means we must always encode the name given to `safeUrlName`.


Antonio-Maranhao:
+1
I tested browsing/creating docs using IDs like `mydoc_,$,(,), ,-,and/doc` and its encoded version `mydoc_%2C%24%2C(%2C)%2C%20%2C-%2Cand%2Fdoc`. Everything seems to work.
@garrensmith

Issue #929, opened 2017-06-08
db security form does not populate with existing object,


wohali:
## Expected Behavior
When browsing to `http://<url>/_utils/#/database/dbname/permissions` , the page should reflect the content of the current database `_security` object. It should then allow incremental editing of the object through adding/deleting admin and member users and roles.

## Current Behaviour
When browsing to `http://<url>/_utils/#/database/dbname/permissions` , the page currently shows no data, as if there was no `_security` object in the database.

## Steps to Reproduce
1. `curl -X PUT http://<url>/dbname`
2. Create a user such as `wohali` in `_users` and make it an admin (e.g. by adding it to `[admins]` in `local.ini`)
3. `curl -X PUT http://wohali:password@<url>/dbname/_security -d '{"admins": {"names": [ "wohali" ], "roles": ["foo", "bar"]}, "members": { "names": ["ralph","frank"], "roles":["developers"]}}'` to create the correct db permissions.
4. `curl -X GET http://wohali:password@<url>/dbname/_security` and validate the output
4. Visit `https://<url>/_utils/#/database/dbname/permissions`. None of the populated data in the `_security` object is shown. 

**Note**: in my browser network console the `GET /_security` on that page returns a 401 error. Inspecting the headers for the request does not show the request being made with the correct security context.

## Your Environment
* Version used: CouchDB, master build
* Browser Name and version: Firefox 53.0.3
* Operating System and version (desktop or mobile): Windows 7 x64, desktop

lostnet:
I hit the same issue in Chrome after upgrading to 2.1.0-27147cd and I think it is a bug in the use of fetch: 

[missing credentials](https://github.com/apache/couchdb-fauxton/blob/14b0901df88a909513d8441fcc499023da3524a8/app/addons/permissions/actions.js?utf8=%E2%9C%93#L31) vs. [included credentials](https://github.com/apache/couchdb-fauxton/blob/14b0901df88a909513d8441fcc499023da3524a8/app/addons/permissions/actions.js?utf8=%E2%9C%93#L88) where the default is to [omit](https://fetch.spec.whatwg.org/#concept-request-credentials-mode).

Strangely, this seems to work in firefox which is probably a bug in the whatwg-fetch pollyfill?

gambolputty:
Seems to be related to [this issue](https://github.com/apache/couchdb/issues/741) over at couchdb.

PR #959, opened 2017-08-16
correct initial fetch() of _security to use credentials (#929)

lostnet:
## Overview

sends credentials in the fetch of a db's existing _security policy so that it is correctly displayed and can be amended with changes. (issue #929) 

## Testing recommendations

In chrome, update permissions on a db, reload fauxton, modify the permissions and save, reload fauxton, verify permissions based on both sessions are correctly displayed.

## GitHub issue number
 929

## Checklist

- [X] Documentation reflects the changes; **(fixes a regression, so presumably)**
- [X] Update [rebar.config.script](https://github.com/apache/couchdb/blob/master/rebar.config.script) with the correct tag once a new Fauxton release is made **(is this an action for me or a request from me?)**


lostnet:
This just fixes the issue of not sending credentials.

How to handle fetch failures is complicated by couchdb being insecure by default. It would be simple enough to add error/catch handling similar to [updatePermission](https://github.com/apache/couchdb-fauxton/blob/14b0901df88a909513d8441fcc499023da3524a8/app/addons/permissions/actions.js?utf8=%E2%9C%93#L94) to show an error notification, but `receivedPermissions()` still has to be called with something to let the user add permissions and `{}` is not a safe default.

Antonio-Maranhao:
@lostnet I agree with you that `{}` is not a safe default. I'd suggest adding the error handling anyway though. This way it prevents the user from inadvertently wiping out the current permissions since the form starts out blank.
Here's my suggestion:

```
return fetch(url, {
    headers: {'Accept': 'application/json' },
    credentials: 'include'
  })
  .then(res => res.json())
  .then(json => {
    if (!json.ok) {
      throw new Error(json.reason);
    }
    dispatch(receivedPermissions(json));
  })
  .catch((err) => {
    FauxtonAPI.addNotification({
      msg: 'Failed to retrieve permissions. Please try again. Reason:' + err.message,
      type: 'error'
    });
  });
```

lostnet:
@Antonio-Maranhao the trouble I have with ignoring the `{}` problem is that receivedPermissions is no longer called if we do normal throwing on the error, so the behavior is still wrong, but different (`isLoading` is then still true for example.) For some users that currently found a way to at least re-populate everything, the new behavior may be worse as we have now gone from wrongly initialized to uninitialized..

What do you think about swapping the error and dispatch?:
```
  .then(json => {
    dispatch(receivedPermissions(json));
    if (json.error && json.reason) {
      throw new Error(json.reason);
    }
  })
```
 This should keep the current bug of putting the error result into the permissions being edited, which I find marginally safer than iterating through wrong behaviors..

Antonio-Maranhao:
@lostnet right. I missed the fact `isLoading` would still be `true`. What about `dispatch(receivedPermissions({ }))` instead of `dispatch(receivedPermissions(json))` then throw the Error? 
The form would still be empty but the user would be notified of the issue (current permissions not loaded). Also, it would no longer update the permissions with the error fields.

lostnet:
@Antonio-Maranhao how about if I assume the db is [secure by default](https://git-wip-us.apache.org/repos/asf?p=couchdb-couch.git;a=blob;f=src/couch_db_updater.erl;h=78726358e2a2edfe7edafc9f60cf218252e54d55;hb=78726358e2a2edfe7edafc9f60cf218252e54d55#l1444), which I think in JS is:
```
if (json.error && json.reason) {
  dispatch(receivedPermissions({admins:{roles:["_admin"]},  users:{ roles:["_admin"]} }));
  throw new Error(json.reason);
}
```
The user can always try to maintain whatever access they need to keep in their first POST to _security, but should be less likely to open up a window of attack.

Antonio-Maranhao:
@lostnet Agree. That's better than the blank form.

lostnet:
@Antonio-Maranhao Excellent, I've added the error checking and the failure case form and verified it works as expected by blocking the connection in chrome devtools. Please review it and let me know what you think?

Antonio-Maranhao:
+1
Great. Worked for me as well.
@garrensmith

garrensmith:
Awesome thanks for the contribution @lostnet

PR #972, opened 2017-08-24
Replace Flux

### LINKED ISSUE #963: Remove Facebook flux from Fauxton
Facebook flux has the same BSD license. We need to remove it.

garrensmith:
We don't use use a lot of Flux's functionlity, so we can replace it with
a simple function that calls all stores with any messages.

Fixes #963


rnewson:
As the replacement code is original I think it would be good if we gave a distinct name rather than calling it flux. Does that break anything? Could aliasing it back to flux fix what it breaks?

I say this so that we're clear about provenance when we look back on this work.

rnewson:
I'm +1 on the change itself.

PR #979, opened 2017-09-06
React motion

garrensmith:
Use react-motion for all animation
This replaces react-addons-transitions and velocity with react-motion for animations. This is part of the work to move off React. This removes the react-addons-transition package dependancy.


Antonio-Maranhao:
+1
It looks good on Chrome, Safari and Firefox.

PR #984, opened 2017-09-20
Fix for Mango Query addon when db name has special characters

Antonio-Maranhao:
## Overview

Fix issues with Mango queries and deletion of indexes when database name contain special characters (e.g. `special/name`)
Also added error handling for `queryAllDocs`

## Testing recommendations

- Create database named `special/name` and create a few documents in it
- Execute a Mango query on this database
- Try to delete one or more docs
- Click on __Manage Indexes__ and create a new index
- Try to delete the new index


garrensmith:
+1 once tests pass

PR #986, opened 2017-09-28
Refactor map/reduce views to use Redux

### LINKED ISSUE #914: Fauxton: Error message missing when server error 401 in view edit
If nothing is done for several minutes in Fauxton, and then a view edit is submitted, the message "Saving view..." occurs on top, and lasts several seconds, but the final message "View saved." is missing. The server log says the return code was 401 (PUT /money/_design/Konten 401 ok 4). No (red) error message is given in this case, but the result is the old one, and if you click elsewhere and click edit that view again, the old code is shown. 
Reloading the page requests a logon, and -- in case nothing else was changed -- shows the new, changed and not saved contents, and save works properly.
This is rather nasty for a newcomer. 
Using Fauxton from couchDB 2.0.0 on port 5984 on a remote machine (not localhost).

Antonio-Maranhao:
## Overview

- [x] Updated map/reduce views to use Redux version of IndexResults components
- [x] Removed non-Redux components from `index-results` 
- [x] Removed Mango resources from `documents/resources.js` 
- [x] Query Options are now preserved when switching between different views, all docs, design docs. Also, if any of the options is enabled (except for All docs) the header is highlighted to indicate this fact.

Extra 
- [x] Bug fix: when creating a new view in an existing design doc, an error message is showed and the page doesn't change to the new view.
- [x] Bug fix: handle error when saving view. Closes #914 

## Testing recommendations

- Select a view and browse the results
  - Change display mode (Table/Metadata/JSON)
  - Click on next/previous page
  - Change the num of 
  - Set query options and click Run Query
    - Check that {}JSON link includes options from Query Options
    - Check that Reduce option only shows for views that contain a reduce function
       - When the Reduce option is enabled, only the Metadata layout is enabled
 - Create a new view in an existing design doc

## Checklist


Antonio-Maranhao:
@garrensmith could you please review it?

app/addons/documents/sidebar/SidebarControllerContainer.js
```diff
@@ -0,0 +1,126 @@
...
+//     hasSelectedItem: getHasDocsSelected(indexResults),
+//     numDocsSelected: getNumDocsSelected(indexResults),
+//     textEmptyIndex: getTextEmptyIndex(indexResults),
+//     docType: getDocType(indexResults),
+//     fetchParams: getFetchParams(indexResults),
+//     queryOptionsParams: getQueryOptionsParams(indexResults)
+//   };
+// };
+
+// const mapDispatchToProps = (dispatch, ownProps) => {
```
garrensmith: Can you remove these?

app/addons/documents/tests/document-test-helper.js
```diff
@@ -26,15 +26,6 @@ function createDocColumn (docs) {
...
-    return Documents.MangoIndex.prototype.parse(doc);
-  });
-
-  return new Documents.MangoIndexCollection(docs, opts);
-}
-
 export default {
-  createDocColumn: createDocColumn,
-  createMangoIndexDocColumn: createMangoIndexDocColumn
+  createDocColumn: createDocColumn
```
garrensmith: Could you make this:

```
export default {
   createDocColumn
}
```

Why repeat ourselves when we don't have too 😄

Antonio-Maranhao:
@garrensmith

Antonio-Maranhao:
@garrensmith finally got it to pass.
Fixed issue #914 as well since I was modifying that page.

garrensmith:
Great work @Antonio-Maranhao

PR #987, opened 2017-09-29
Update code editor's contents when a new document is loaded

### LINKED ISSUE #977: While viewing a document, changing the url and hitting enter doesn't change the document editor contents
Changing the URL of a document does not change to that document on first try, only if you subsequently hit enter (which results in the browser doing a full page refresh).

## Steps to Reproduce
 - Navigate to some document. The URL in your bar will be `http://blah/_utils/#database/db-name/document-name-1`. You can see the JSON of `document-name-1`
 - Change the URL to another document, say, `http://blah/_utils/#database/db-name/document-name-2`. Hit enter.

## Current Behaviour
 - The **title** changes, `database > document-id`
 - The content of the editor does not change, and is still the old doc

## Expected Behaviour 
 - Both the title and the content change.

Interestingly enough, if you give it an incorrect document id it does boot you back to the db screen telling you that it couldn't find the document, so it's definitely looking for it.

## Your Environment
* Version used: Build with CouchDB 2.1
* Browser Name and version: Chrome 60
* Operating System and version (desktop or mobile): MacOS

Antonio-Maranhao:
## Overview

 Code editor is not updated when switching to a different document by typing a new URL directly in the browser's address bar.

## Testing recommendations

- Navigate to some document. The URL in your bar will be `http://blah/_utils/#database/db-name/document-name-1`. You can see the JSON of `document-name-1`
- Change the URL to another document, say, `http://blah/_utils/#database/db-name/document-name-2`. Hit enter.
- The JSON of `document-name-2` should be displayed. The page header displays the new document ID.

## GitHub issue number

Closes #977

### DIFF
--- a/app/addons/documents/doc-editor/components.js
+++ b/app/addons/documents/doc-editor/components.js
@@ -84,9 +84,11 @@ var DocEditorController = React.createClass({
     store.off('change', this.onChange);
   },
 
-  // whenever a file is uploaded or a doc is cloned update the editor
   componentWillUpdate: function (nextProps, nextState) {
-    if (this.state.numFilesUploaded !== nextState.numFilesUploaded || this.state.doc && this.state.doc.hasChanged()) {
+    // Update the editor whenever a file is uploaded, a doc is cloned, or a new doc is loaded
+    if (this.state.numFilesUploaded !== nextState.numFilesUploaded ||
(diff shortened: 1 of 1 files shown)


Antonio-Maranhao:
@garrensmith could you please review it?

garrensmith:
+1 this looks great. Thanks

Issue #957, opened 2017-08-15
Feature Request: Auto quote Query keys if quotes missing


dharders:
When querying database documents through the Fauxton interface, `i.e. Clicking Options -> By Keys or Between Keys`, it should be possible to enter a string without having to surround it with quotes.

I've lost count of how many times I have entered key values, returning 0 results, even though the documents exist. I then have to go back and surround the values for the query to execute correctly. It's a jarring experience that could be easily solved by Fauxton doing a simple check for surrounding quotes and if not present add them.

## Expected Behavior

Add surrounding quotes to Query Key strings if missing
## Current Behavior

No results returned if quotes missing (which is technically correct, but not the best user experience)
## Possible Solution

Simple string check for surrounding quotes and if not present add them on submit. Could also make an option to 'Auto quote' etc, which is off by default.

## Your Environment

* Version used: 2.1.0
* Browser Name and version: Chrome 59.0.3071.115  
* Operating System and version (desktop or mobile):  Win10 x64 client, Debian (Jessie) x64 Server

PR #989, opened 2017-10-02
Basic validation of key values entered in the Query Options panel

Antonio-Maranhao:
## Overview

A common (newbiew) user mistake is to forget to include double quotes in the By Keys and Between Keys fields (Query Options panel) for String values. (See https://github.com/apache/couchdb-fauxton/issues/957)
Since CouchDB accepts arbitrary JSON values as key, automatically adding the quotes is not ideal. Instead, the panel does some basic validation and displays a warning when necessary.

## Testing recommendations

- Between Keys filter
  - It should display the warning for String values without quotes (e.g.: abc)
  - It should NOT display the warning for numbers and boolean.
- By Keys filter
  - It should display the warning if the value is not enclosed in square brackets (e.g.: "abc")

## GitHub issue number

Related to https://github.com/apache/couchdb-fauxton/issues/957


Antonio-Maranhao:
@willholley can you please review?

popojargo:
@Antonio-Maranhao Why don't we automatically add the double quotes when entering a string that doesn't start and end by  {} or [] ?

For example, I enter foo and it gets formated to => `"foo"`

Antonio-Maranhao:
@popojargo sorry for the late reply. My thinking at the time was to have a quick fix. Having the field auto-quote was a bit trickier to implement (what if the string itself has quotes, and supporting single/double quotes, etc). So I settled for showing a warning instead. 

I thought about adding the quotes behind the scenes, before sending the request, but the behavior would not be consistent. E.g. you don't use quotes for simple strings but still need them if you're using arrays. And even then if the user adds quotes themselves - do you just ignore, or are they part of the value?

Maybe I am just overthinking though. :)

wohali:
I'd love to see this land, but the PR needs rework, and I hate seeing PRs languish this long.

Please reopen when you have a chance to get back to this.

PR #990, opened 2017-10-03
Upgrade for react 16

garrensmith:
Upgrades to use the new prop-types library and converts all createClass to use React.Components.
I've used the react-codemods to do this https://github.com/reactjs/react-codemod


Antonio-Maranhao:
+1
Everything still works :)

PR #993, opened 2017-10-05
Display Mango execution statistics, when available

willholley:
## Overview

When running a Mango Query, display the execution stats and any warnings returned. The query execution time and presence of a warning is indicated in the query editor after a query
has been run. An overlay popup displays the detailed execution statistics and warning text.

Not all versions of CouchDB support execution statistics for Mango, so detect that the feature is supported before requesting them at query time.

In the case that execution statistics are not available, we still display any warnings that are returned.

execution_stats

execution_stats_popup

## Testing recommendations

Run some Mango queries against CouchDB master (where execution stats are supported) and against CouchDB 2.1 (which does not). Check that you can see execution statistics as in the screenshot. Ideally, try some queries which have index coverage (and so should not generate a warning) and some that do not.

## Checklist


app/addons/documents/mango/components/ExecutionStats.js
```diff
@@ -0,0 +1,127 @@
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
```
garrensmith: can you rename to make this file all lowercase

app/addons/documents/mango/components/ExecutionStats.js
```diff
@@ -0,0 +1,127 @@
...
+import { Popover, OverlayTrigger } from 'react-bootstrap';
+
+const TOO_MANY_DOCS_SCANNED_WARNING = "The number of documents examined is high in proportion to the number of results returned. Consider adding an index to improve this.";
+
+export default class ExecutionStats extends React.Component {
+  constructor (props) {
+    super(props);
+  }
+
+  humanizeDuration(milliseconds) {
```
garrensmith: We use moment.js. There should be a helper function to help with this, that might make this simpler

app/addons/documents/mango/components/ExecutionStats.js
```diff
@@ -0,0 +1,127 @@
...
+    const minutes = Math.floor(seconds / 60);
+    seconds = seconds - (minutes * 60);
+
+    return minutes + 'minute, ' + seconds + 'seconds';
+  }
+
+  getWarning(executionStats, warning) {
+    if (executionStats) {
+      // warn if many documents scanned in relation to results returned
+      if (executionStats && executionStats.results_returned) {
```
garrensmith: This can be `if (executionStats.results_returned) {`

garrensmith: Actually you can invert the first if statement:

`if (!executionStats) { return warning;}`

app/addons/documents/mango/components/ExecutionStats.js
```diff
@@ -0,0 +1,127 @@
...
+      );
+    } else if (!!warningText) {
+      executionStatsComponent = (
+        <span className="execution-stats-component">Warning</span>
+      );
+    }
+
+    const popup = this.popup(executionStats, warningText);
+    return (
+        <OverlayTrigger trigger={['hover', 'focus']} placement="right" overlay={popup}>
```
garrensmith: Can you also add click as a trigger. Just in case anyone uses this with a touch screen

app/addons/documents/mango/mango.api.js
```diff
@@ -84,6 +84,30 @@ export const fetchIndexes = (databaseName, params) => {
...
+          "_id": {"$gt": "a" }
+       },
+        execution_stats: true
+      }, {limit: 1})
+      .then(resp => {
+        supportsExecutionStatsCache = resp.status == 200;
+        resolve(supportsExecutionStatsCache);
+      });
+    });
+  } else {
```
garrensmith: You don'e need an else since you are returning in the if.

app/addons/documents/mango/mango.api.js
```diff
@@ -84,6 +84,30 @@ export const fetchIndexes = (databaseName, params) => {
...
+    return new FauxtonAPI.Promise((resolve) => {
```
garrensmith: You can use [Promise.resolve](http://bluebirdjs.com/docs/api/promise.resolve.html) instead

app/addons/documents/mango/mango.api.js
```diff
@@ -104,9 +128,10 @@ export const mergeFetchParams = (queryCode, fetchParams) => {
   };
 };
 
-export const mangoQueryDocs = (databaseName, queryCode, fetchParams) => {
-  const url = FauxtonAPI.urls('mango', 'query-server', encodeURIComponent(databaseName));
+export const mangoQuery = (databaseName, queryCode, fetchParams) => {
+  const url = FauxtonAPI.urls('mango', 'query-server', databaseName);
```
garrensmith: Can yo uadd the encodeURIComponent back here. Otherwise this doesn't work with any database with `/` in it

app/addons/documents/mango/mango.api.js
```diff
@@ -115,15 +140,26 @@ export const mangoQueryDocs = (databaseName, queryCode, fetchParams) => {
...
-        docType: Constants.INDEX_RESULTS_DOC_TYPE.MANGO_QUERY
-      };
-    });
+  });
+};
+
+export const mangoQueryDocs = (databaseName, queryCode, fetchParams) => {
+  return supportsExecutionStats(databaseName).then((shouldFetchExecutionStats) => {
+    if (shouldFetchExecutionStats) {
+      queryCode.execution_stats = true;
```
garrensmith: Can you add a note here that we can only add the execution_stats header if it is supported otherwise mango throws errors for unsupported json fields

app/addons/documents/mango/components/ExecutionStats.js
willholley: the other component files are camelcased - shouldn't we stick to convention?

app/addons/documents/mango/components/ExecutionStats.js
willholley: aha - I didn't realise that and didn't want to pull in the dependency. If it's there, I'll use it.

willholley: looking at it, the moment.js `humanize()` function rounds up milliseconds to seconds, so I don't think it's the right thing for this.

garrensmith: ok that makes sense

app/addons/documents/mango/components/ExecutionStats.js
garrensmith: Yes. We should have made them all lowercase. Lets leave it as is then. And we can lowercase them all in another PR

garrensmith:
+1 to merge once tests pass

willholley:
@garrensmith the tests are failing but I don't think the failures are related to this PR - any idea what might be wrong?

PR #1001, opened 2017-10-20
Fix table mode to show boolean values

Antonio-Maranhao:
## Overview

Fixes display of boolean values in Table mode. 

### Current behavior
Table mode won't display boolean values of top-level fields. E.g.: `{ "_id": "test", "f": false, "t": true }`

### Expected behavior
Table mode displays `true` or `false` for boolean fields. 

## Testing recommendations

 - Create a new doc `{ "_id": "test", "f": false, "t": true }`
 - In the all_docs list, switch to Table mode.
 - Values for `f` and `t` should be displayed

### DIFF
--- a/app/addons/documents/index-results/components/results/TableRow.js
+++ b/app/addons/documents/index-results/components/results/TableRow.js
@@ -37,7 +37,8 @@ export default class TableRow extends React.Component {
     const row = this.props.data.selectedFields.map(function (k, i) {
 
       const key = 'tableview-data-cell-' + rowNumber + k + i + el[k];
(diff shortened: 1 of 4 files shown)


app/addons/documents/index-results/components/results/TableRow.js
```diff
@@ -37,7 +37,8 @@ export default class TableRow extends React.Component {
     const row = this.props.data.selectedFields.map(function (k, i) {
 
       const key = 'tableview-data-cell-' + rowNumber + k + i + el[k];
-      const stringified = typeof el[k] === 'object' ? JSON.stringify(el[k], null, '  ') : el[k];
```
garrensmith: This looks good. Can you write a test that proves this works and so we don't ever have a regression on this.

PR #1004, opened 2017-10-23
Replication page renders even with missing URL

garrensmith:
Handle bad/missing urls by ignoring them in the formatUrl function.
In same cases a replication doc might not have a source or target and this causes the replication dashboard page to crash. We need to handle the missing url and continue to display the page.


Antonio-Maranhao:
+1 Looks good!

PR #1011, opened 2017-11-01
Fix could not bulk delete from table view

garrensmith:
Fixed an issue where the _rev was not getting added to the doc so table
view could not do bulk deletes

## Checklist


garrensmith:
@Antonio-Maranhao and @popojargo would you mind reviewing this for me?

Antonio-Maranhao:
+1
Thanks for fixing this!

Issue #949, opened 2017-08-08
No list of databases shown in IE 11 


gdelfino:
In CouchDB 2.1, when using Internet Explorer 11 (Win7) the included Fauxton does not show the list of databases nor a login screen, so it cannot be used. The only thing shown is the navigation bar on the left.

By having the built-in debugger break on all exceptions, I can see that IE complains about a syntax error on the expression "e.querySelectorAll("*,:x") included in vendor.js

wohali:
@garrensmith I can indeed confirm this is a bug as stated. Not sure how much effort we want to put into fixing it.

garrensmith:
If the error is from vendor.js it means that one of the libraries that we depend on is failing.

bagguley:
@wohali Are you serious? You're not sure how much effort to put into this? CouchDB is now unusable unless you install Firefox because Chrome doesn't even render anything on the page. It's appalling that such simple testing wasn't conducted before release. The exception that @gdelfino referred to is probably a red herring because jQuery (which I presume is where it's occurring) handles these exceptions, they are expected.

Chrome has around 60% of the browser share of desktop browser usage with Firefox bumbling around 6  to 10%. I suggest that this should be top priority.

garrensmith:
Hi @bagguley,

That's a pretty aggressive tone. You are the first person to report it doesn't work on Chrome. Could you provide more details around that. We do a lot of testing to make sure it works on chrome and Firefox. So I am surprised to hear it doesn't work on it.

Cheers
Garren

All misspelling thanks to my iPhone.
________________________________

bagguley:
Hi @garrensmith

It wasn't meant aggressively, but just in a completely shocked tone. It works in Edge and Firefox but not in IE with the Edge renderer. I'll try and dig out more details tomorrow.

Best regards.

wohali:
Can confirm everything is fine in Chrome. We don't do testing on IE so I'm unsurprised it's broken in IE11...as always we're happy to welcome PRs if you can help us get to the bottom of things.

gdelfino:
There are many enterprises still using Win7/IE 11. And installing an alternative browser is specifically forbidden. Fixing this IMHO is important in order to prevent excuses about why we should not be using CouchDB. We will try to look deeper into what is causing the issue.

bagguley:
After some investigation it looks like the issue could be one or more of the following:

popojargo:
[...]

PR #1013, opened 2017-11-01
Support fauxton on IE10,IE11 and Edge

popojargo:
## Overview

- Added babel-polyfill to support ES6 functions on browsers that doesn't support them
- Fixed layouts problems on IE11 and Edge

## Testing recommendations

Since we seems to mainly support Firefox and Chrome, there are not automated tests for IE.

With my changes, I fixed:
- EDGE: Fixed JSON API bar display
- IE11/IE10 : Fixed replication page that wasnt shown
- IE11/IE10 : We are now able to use React 👍 
- IE11/IE10 : Polling slider where not shown correctly (but they still don't work correctly on IE)

## GitHub issue number

#949

## Checklist


Antonio-Maranhao:
@popojargo @garrensmith 
Took me some time to get a VM with IE11 but now I'm set. I'll continue looking into this tomorrow.

webpack.config.dev.js
```diff
@@ -20,7 +20,7 @@ const settings = require('./tasks/helper')
 
 module.exports = {
   entry: {
-    bundle: './app/main.js' //Our starting point for our development.
+    bundle: ['babel-polyfill','./app/main.js'] //Our starting point for our development.
```
Antonio-Maranhao: To address a previous comment from @garrensmith we could be more granular using `core-js` and only polyfill only what's needed. I was able to launch the UI using `bundle: ['core-js/fn/array', 'core-js/fn/symbol', 'core-js/fn/promise', './app/main.js']`

Antonio-Maranhao: Side note... `webpack.config.release.js` needs to be updated as well.

app/addons/activetasks/components.js
```diff
@@ -388,8 +388,13 @@ export class ActiveTasksPollingWidgetController extends React.Component {
   state = this.getStoreState();
 
   render() {
+    const isIE1X = document.documentMode == 11 || document.documentMode == 10;
```
Antonio-Maranhao: Could you replace with the `Helpers.isIE1X()` function you created?

Antonio-Maranhao:
Btw... @popojargo are you planning on fixing Polling slider as well? If not we should open an issue so we don't lose track of it.

garrensmith:
We using Bluebird for promises so we might not need that polyfill.

________________________________

webpack.config.dev.js
popojargo: Do you have any idea how I can found which polyfill are missings?

popojargo:
@Antonio-Maranhao As for the polling, I think it could be for another issue. This is mostly a problem with the mouse events in IE and doesn't have any easy fix so far.

popojargo:
@Antonio-Maranhao I managed to fix the issue for the polling with a package: react-script. The onChange event for range inputs should be fixed on React16(but I can't confirm).

Antonio-Maranhao:
@garrensmith any concerns about the ISC license for `react-range`?
https://github.com/mapbox/react-range

popojargo:
@Antonio-Maranhao ISC is like MIT so it should pass right?

garrensmith:
From the list it looks fine https://www.apache.org/legal/resolved.html#category-x

________________________________

PR #1027, opened 2017-11-20
Use PUT endpoint to clone doc

garrensmith:
## Overview

The COPY doc api doesn't support cloning docs with ID's that need to be
encoded. So use the PUT command and just create a new doc instead

## Testing recommendations
Clone a document and test that it can clone a document. Ideally clone a document that would need to be encoded e.g `encoded/doc`.

## Checklist


Antonio-Maranhao:
+1

PR #1028, opened 2017-11-27
upgrade to selenium 3.4.0

garrensmith:
Use a more recent version of Selenium


garrensmith:
I'm going to wait for a newer version of nightwatch before doing this.

PR #1029, opened 2017-12-06
Feat/sidebar design fixes

lucidNTR:
## Overview

The sidebar requires too much brain overhead and i tried to improve the usability with common UI design patterns:
 
 - rule of 5s: clusters of items with more than 5 are perceived as "many" and individual items are hard to find. by seperating the menu entries into the top and bottom elements with a little spacer, the brain is releived a bit.
 - order of importance: the top elements of a list are generally more important then lower elements, but the sidebar toggle is still the highest element in the sidebar, even though it has the lowest importance of the entries. also the icons to show and hide the sidebar titles is non standard and hard to understand, and very different between the show and hide states (eg. logout button to hide the sidebar titles.) I tried to adopt as much to the standard way to represent hiding sidebar titles as possible, while having minimal impact on the overall design.

## Testing recommendations

Purely visual changes, please rebuild and manually inspect the design changes.

### DIFF
--- a/app/addons/documents/assets/less/header-docs-left.less
+++ b/app/addons/documents/assets/less/header-docs-left.less
@@ -58,6 +58,7 @@ button.faux-header__doc-header-dropdown-toggle:focus {
   border: none;
   background-color: transparent;
   border-right: 1px solid #ccc;
+  color: #333;
   width: 40px;
   padding: 11px;
 }
--- a/app/addons/fauxton/assets/less/navigation.less
+++ b/app/addons/fauxton/assets/less/navigation.less
@@ -38,7 +38,7 @@
 
 .faux-navbar__burger {
   background-color: @brandDark2;
-  padding: 19px 0 18px 18px;
+  padding: 19px 0 18px 25px;
 }
 
 .faux-navbar--narrow {
@@ -55,7 +55,7 @@
 
 .faux-navbar__burger__icon {
   color: @navIconColor;
-  font-size: 27px;
+  font-size: 19px;
 }
 
 .faux-navbar__burger__icon--flipped:before{
--- a/app/addons/fauxton/assets/less/navigation.less
+++ b/app/addons/fauxton/assets/less/navigation.less
@@ -201,6 +201,10 @@
   background-size: 40px;
 }
 
+.faux-navbar__spacer {
+  height: 6px;
+}
+
 .faux-navbar__footer {
   display: flex;
   flex-direction: column;
--- a/app/addons/fauxton/navigation/components/Burger.js
+++ b/app/addons/fauxton/navigation/components/Burger.js
@@ -25,8 +25,8 @@ const Burger = ({toggleMenu, isMinimized}) => {
   );
 
   const icon = isMinimized ?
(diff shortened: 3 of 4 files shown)


garrensmith:
@lucidNTR I'm not 100% sure about this change. The previous sidebar redesign was done by @justinmcdavid and @robertkowalski. They gave it quite a lot of thought and did some intentional changes to make it more obvious that the sidebar could expand.

garrensmith:
A screenshot to demo this:
screenshot 2017-12-07 08 30 50

popojargo:
I'm not a UX expert but It feels weird to have that expand/collapse button in the middle of the sidebar.

Also, why don't we use ☰ for the sidebar icon?

mikerhodes:
It would be good to see some citations for the rules presented. 

I do think that the separation between the database management bits and the docs/config ones is useful. I'm not sure if 5 is the right number here.

Antonio-Maranhao:
My two cents... 
- I also don't think the space between the top/bottom items is useful. At least for me the left nav bar is similar to top drop down menus. Many apps have way more than 5 menus, and the important distinction is they (usually) group functions unrelated to each other. 
- I am neutral on this one. I don't particularly dislike the arrow at the bottom. I used it once to minimize the menu and haven't touched it in ages. But Garren made a good point, for novice users, moving it to the bottom could make it harder to find at first.

justinmcdavid:
@lucidNTR Hick's Law does a decent job of explaining the issue I think you're trying to address: https://en.wikipedia.org/wiki/Hick%27s_law Small separators would help subdivision, though I think moving the the collapse/uncollapse control would be a mistake. It's best to avoid treating the collapse/uncollapse icon as if it's a menu item, and the serial primacy plus different visual treatment help to differentiate it.
 
@popojargo the ☰ icon serves a different role than that of uncollapsing multiple icons into multiple menu items. ☰ usually becomes an entire menu. While working with this same convention on Cloudant, we saw a lot of confusion during user testing.

lucidNTR:
im unsure how to react, this could easily evolve into bikeshedding, which i want to avoid by all means. Could we not just be at least consistent with cloudant? Users get used to it and now there are a few other services that use burgers for this "half collapsed menu"... I could live without the seperator for now, but the icon just feels amateurish, as it is used for "sign in" or "sign out" so has an official completely meaning.

justinmcdavid:
Cloudant users were also confused by the hamburger, so the design changes were for both Cloudant and Couch. Being consistent with Cloudant (and going with convention rather than against it) would mean not using the hamburger:

Here's what's on the current Cloudant splash page: https://1.cms.s81c.com/sites/default/files/styles/band_inline_image_standard/public/2017-12/leadspace.png?itok=cpKljhuH

robertkowalski:
death to the burger! :)

https://medium.com/the-ux-chap/how-the-hamburger-menu-almost-led-to-a-triple-digit-million-pound-loss-ef6cdfa34f1e

garrensmith:
I'm going to close. If anyone wants to lead a discussion around the sidebar lets open an issue where we can discuss it and plan.

PR #1042, opened 2018-01-02
pin babel-eslint to 8.0.1

garrensmith:
Fix the linting issues we are getting by pinning babel-eslint to 8.0.1


garrensmith:
@popojargo and @Antonio-Maranhao I've merged this in without a +1 because its a pretty straight forward change.

PR #1043, opened 2018-01-03
Upgrade to eslint v4 and enable recommended lint rules

Antonio-Maranhao:
## Overview

- Upgrade to ESLint 4.14 and eslint-plugin-react 7.5.1
- Use recommended ESLint rules
- Use recommended React lint rules

NOTE: Although the changes are quite small, there are quite a few of them. Probably easier to review each commit separately. 

## Testing recommendations

Run `npm run stylecheck`

## Checklist


app/addons/replication/components/common-table.js
```diff
@@ -24,7 +24,6 @@ export const formatUrl = (url) => {
     urlObj = new URL(removeCredentialsFromUrl(url));
     encoded = encodeURIComponent(urlObj.pathname.slice(1));
   } catch (error) {
```
garrensmith: You can remove the `error` variable if we not going to log it.

app/addons/replication/controller.js
```diff
@@ -257,8 +254,6 @@ export default class ReplicationController extends React.Component {
     } else {
```
garrensmith: Very minor but could you remove the `else` and just `return [{'name': 'Replication'}];` by default

.eslintrc
```diff
@@ -22,14 +21,17 @@
...
     "semi-spacing": [2, {"before": false, "after": true}],
-    "no-empty": 2,
     "space-in-parens": [2, "never"],
+    "space-unary-ops": 2,
     "no-unused-vars": [2, {"varsIgnorePattern": "ReactDOM"}],
-    "react/jsx-uses-vars": 2,
-    "react/jsx-uses-react": 2
+    "react/display-name": 0,
+    // disabling for now since there are too many components without PropTypes
+    "react/prop-types": 0
```
garrensmith: I'm not sure if this is all ready in the rules that we inherit from but should we set line spacing to 2?

Issue #1051, opened 2018-01-26
Failed to retrieve permissions.


gizocz:
Cannot set permissions on databases with '/' or '+' character in the database name.

## Steps to Reproduce
1. click "Create Database", enter "a/b", click "Create"
2. click "Permissions" => alert "Failed to retrieve permissions. Please try again. Reason:Database does not exist."; error on console: "https://localhost:6984/a/b/_security 404 ()". 
3. try to add a role => alert "Could not update permissions - reason: Error: Database does not exist."

## Environment
* Version used: Fauxton on Apache CouchDB v. 2.1.1 (from official docker hub repository)
* Browser Name and version: Google Chrome v 63.0.3239.132
* Operating System and version: Debian GNU/Linux 9.3

PR #1052, opened 2018-02-04
Encode database id in permissions

popojargo:
## Overview

When we were trying to edit permissions for a database with special characters, it was failing.
## GitHub issue number

#1051 

## Checklist


Antonio-Maranhao:
Good stuff @popojargo - thanks for the fix!

PR #1053, opened 2018-02-06
Throttle fetch requests triggered by key press events

Antonio-Maranhao:
## Overview

The search fields at the top fire one fetch request for each key pressed. Besides being a waste of resources, a user could overload the server by simply typing fast enough on these fields. 

This PR changes this behaviour so a fetch is only triggered 400ms after the last character was typed in.

## Testing recommendations

 - Open the network tab in the browser's devtools
 - In the database list page, type a few characters in the Database Name field at the top. 
 - If you typed fast enough (i.e. time between each key press is < 400ms) you should see only one request in the network tab
 - Repeat the steps above for the Document search in the All Documents page

## Checklist

### DIFF
--- a/app/addons/components/components/throttledreacselect.js
+++ b/app/addons/components/components/throttledreacselect.js
@@ -1,0 +1,49 @@
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License. You may obtain a copy of
+// the License at
+//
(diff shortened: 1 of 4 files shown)


Antonio-Maranhao:
@popojargo refactored it into a separate component. It eliminates the duplicated code and makes `lastCall` an instance var, which is better than having it in the global scope.

popojargo:
@Antonio-Maranhao  I'll review your pull requests tomorrow!

PR #1059, opened 2018-03-06
use npm ci on travis

garrensmith:
Use npm ci on travis for faster builds and stricter package management

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ git:
 
 before_script:
   - travis_retry ./bin/docker-up-and-check.sh
+  - npm install -g npm@latest
   - npm ci
   - curl http://127.0.0.1:5984
   - npm run stylecheck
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -9,7 +9,8 @@ git:
 
 before_script:
   - travis_retry ./bin/docker-up-and-check.sh
-  - npm install
+  - npm install -g npm@latest
+  - npm ci
```
Antonio-Maranhao: Travis is running `npm install` before getting to this part - it's automatic for Node projects that don't have a `install` section - which means `package-lock` could have been changed before `npm ci` runs.
The alternative is to override the default by moving these two lines to `install` section.
https://docs.travis-ci.com/user/customizing-the-build/#Customizing-the-Installation-Step

Antonio-Maranhao: Just saw https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Using-a-specific-npm-version which recommends `npm install -g npm@latest` should be on `before_install` section.

PR #1060, opened 2018-03-08
Remove all uses of JQuery from the verifyinstall addon

Antonio-Maranhao:
## Overview

Remove all uses of JQuery in the verifyinstall addon

## Testing recommendations

Run the 'Verify Installation' feature

## Checklist

### DIFF
--- a/app/addons/verifyinstall/resources.js
+++ b/app/addons/verifyinstall/resources.js
@@ -48,10 +48,7 @@ Verifyinstall.testProcess = {
   },
 
   saveDB: function () {
-    return FauxtonAPI.when([
-      db.save(),
-      dbReplicate.save()
-    ]);
+    return db.save();
   },
 
   setupDB: function (db) {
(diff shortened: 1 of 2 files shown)


app/addons/verifyinstall/resources.js
```diff
@@ -47,7 +48,10 @@ Verifyinstall.testProcess = {
   },
 
   saveDB: function () {
-    return db.save();
+    return FauxtonAPI.when([
```
garrensmith: Could you use `Fauxton.Promise.all` instead of FauxtonAPI.when.

app/addons/verifyinstall/resources.js
```diff
@@ -77,16 +81,21 @@ Verifyinstall.testProcess = {
   },
 
   testView: function () {
-    var deferred = FauxtonAPI.Deferred();
-    var promise = $.get(viewDoc.url() + '/_view/testview');
+    const deferred = FauxtonAPI.Deferred();
```
garrensmith: Could we use a promise instead of a Deferred

Antonio-Maranhao:
@garrensmith @popojargo Had to change the order of `testPassed` / `testFailed` calls because they didn't match the actual failures.

PR #1061, opened 2018-03-08
use latest chrome and selenium

garrensmith:
Use latest chrome and selenium for our nightwatch tests.

### DIFF
--- a/docker/dc.selenium-debug.yml
+++ b/docker/dc.selenium-debug.yml
@@ -5,7 +5,7 @@ services:
       file: './dc.selenium.yml'
       service: selenium
     container_name: selenium
-    image: selenium/standalone-firefox-debug:2.48.2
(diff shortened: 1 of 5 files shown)


app/addons/fauxton/tests/nightwatch/updatesUrlsSameRouteobject.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
     client
       .populateDatabase(newDatabaseName)
       .loginToGUI()
+      .pause(1000)
```
Antonio-Maranhao: Is this pause really needed?

garrensmith: Not that is a mistake. I'll remove it

PR #1062, opened 2018-03-08
Remove JQuery usage from config addon

Antonio-Maranhao:
## Overview

 - Remove JQuery usage from config addon
 - Also fixes a bug in the delete confirmation modal where the Cancel and X buttons were not closing the modal. 

## Testing recommendations

- Go to Config page
  - Add a new option
  - Edit the value of the new option
  - Delete the new option

## Checklist

### DIFF
--- a/app/addons/config/components.js
+++ b/app/addons/config/components.js
@@ -230,9 +230,13 @@ class ConfigOptionValue extends React.Component {
 }
 
 class ConfigOptionTrash extends React.Component {
-  state = {
-    show: false
-  };
+  constructor (props) {
+    super(props);
+    this.onDelete = this.onDelete.bind(this);
+    this.showModal = this.showModal.bind(this);
+    this.hideModal = this.hideModal.bind(this);
+    this.state = { show: false };
+  }
 
   onDelete = () => {
     this.props.onDelete();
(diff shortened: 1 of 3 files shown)


app/addons/config/components.js
```diff
@@ -248,13 +252,12 @@ class ConfigOptionTrash extends React.Component {
 
   render() {
     return (
-      <td className="text-center config-item-trash config-delete-value"
-        onClick={this.showModal.bind(this)}>
-        <i className="icon icon-trash"></i>
+      <td className="text-center config-item-trash config-delete-value">
+        <i className="icon icon-trash" onClick={this.showModal}></i>
```
popojargo: Any reason to move the onClick handler to <i> instead of <td>?

app/addons/config/resources.js
```diff
@@ -30,20 +31,20 @@ Config.OptionModel = Backbone.Model.extend({
...
-      params.type = 'DELETE';
-    } else {
-      params.type = 'PUT';
+      operation = deleteRequest;
     }
-    return $.ajax(params);
+
+    return operation(
+      model.url(),
+      model.get('value')
```
popojargo: Passing the value for the deleteRequest might have side effects. deleteRequest take the second parameter as options (not body)

app/addons/config/components.js
Antonio-Maranhao: @popojargo Yes. Having the onClick inside `<td>` was causing it to be fired if you clicked any where on the modal (even the gray background). Meaning, the modal couldn't be closed when clicking on Cancel or the X button because it trigged the event to show the modal again.

app/addons/config/resources.js
Antonio-Maranhao: Good catch. I changed the code since the body is not needed for the DELETE request.

PR #1065, opened 2018-03-09
Remove JQuery uses from documents addon

Antonio-Maranhao:
## Overview

Remove JQuery uses from documents addon.
Also as I was testing cancellation of a file upload, I noticed the 'Upload' button and 'Choose file' were still active when the upload is in progress. I fixed that so both will be disabled while upload is in progress.

## Testing recommendations

Test these features:
 - Bulk delete docs
 - Upload attachment to existing doc
   - Cancel an upload
   - While upload is in progress, user should only be able to press Cancel / Close the modal.
 - Delete document from Document Editor
 - Create new doc and check that `_id` field is set to a random UUID value
 - Go to All Docs page, then check that the Search combobox at the top works
 - List database changes

To test cancelling a file upload, it helps to throttle the network speed using in the Network tab of Chrome's devtools.

## Checklist

### DIFF
--- a/app/addons/documents/doc-editor/actions.js
+++ b/app/addons/documents/doc-editor/actions.js
@@ -13,9 +13,10 @@
 /* global FormData */
 
 import FauxtonAPI from "../../../core/api";
+import { deleteRequest } from "../../../core/ajax";
 import ActionTypes from "./actiontypes";
 
-var xhr;
+var currentUploadHttpRequest;
 
 function initDocEditor (params) {
   var doc = params.doc;
@@ -79,33 +80,24 @@ function hideDeleteDocModal () {
 }
 
 function deleteDoc (doc) {
-  var databaseName = doc.database.safeID();
-  var query = '?rev=' + doc.get('_rev');
-
-  $.ajax({
-    url: FauxtonAPI.urls('document', 'server', databaseName, doc.safeID(), query),
-    type: 'DELETE',
-    headers: {
-      'Accept': 'application/json',
-      'Content-Type': 'application/json',
-    },
-    xhrFields: {
-      withCredentials: true
-    },
-    success: function () {
-      FauxtonAPI.addNotification({
-        msg: 'Your document has been successfully deleted.',
-        clear: true
-      });
-      FauxtonAPI.navigate(FauxtonAPI.urls('allDocs', 'app', databaseName, ''));
-    },
-    error: function () {
-      FauxtonAPI.addNotification({
-        msg: 'Failed to delete your document!',
-        type: 'error',
-        clear: true
-      });
+  const databaseName = doc.database.safeID();
+  const query = '?rev=' + doc.get('_rev');
+  const url = FauxtonAPI.urls('document', 'server', databaseName, doc.safeID(), query);
+  deleteRequest(url).then(res => {
+    if (res.error) {
+      throw new Error(res.reason || res.error);
     }
+    FauxtonAPI.addNotification({
+      msg: 'Your document has been successfully deleted.',
+      clear: true
+    });
+    FauxtonAPI.navigate(FauxtonAPI.urls('allDocs', 'app', databaseName, ''));
+  }).catch(err => {
+    FauxtonAPI.addNotification({
+      msg: 'Failed to delete your document. Reason: ' + err.message,
+      type: 'error',
+      clear: true
+    });
   });
 }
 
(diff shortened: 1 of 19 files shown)


app/addons/documents/doc-editor/actions.js
```diff
@@ -79,33 +80,24 @@ function hideDeleteDocModal () {
...
-      });
+  const databaseName = doc.database.safeID();
+  const query = '?rev=' + doc.get('_rev');
+  const url = FauxtonAPI.urls('document', 'server', databaseName, doc.safeID(), query);
+  deleteRequest(url).then(res => {
+    if (res.error) {
+      throw new Error(res.reason || res.error);
     }
+    FauxtonAPI.addNotification({
+      msg: 'Your document has been successfully deleted.',
```
popojargo: It should be success right?

app/addons/documents/doc-editor/actions.js
```diff
@@ -79,33 +80,24 @@ function hideDeleteDocModal () {
...
-
-  $.ajax({
-    url: FauxtonAPI.urls('document', 'server', databaseName, doc.safeID(), query),
-    type: 'DELETE',
-    headers: {
-      'Accept': 'application/json',
-      'Content-Type': 'application/json',
-    },
-    xhrFields: {
-      withCredentials: true
```
popojargo: I think you have to include  something similar to that in fetch. I think it's:   credentials: 'include'

Antonio-Maranhao: The functions in `core/ajax` all use `credentials: 'include'`

app/addons/documents/doc-editor/actions.js
Antonio-Maranhao: Added the type

Antonio-Maranhao:
Pushed a small fix so the attachment's `content-type` is not empty when the file type is unknown.

app/addons/documents/doc-editor/actions.js
```diff
@@ -156,75 +149,88 @@ function uploadAttachment (params) {
...
+        }
+      }
+      onError(errorMsg);
+    }
+  };
+  httpRequest.open('PUT', url);
+  httpRequest.setRequestHeader('X-Requested-With', 'XMLHttpRequest');
+  httpRequest.setRequestHeader('Content-Type', file.type || `application/octet-stream`);
+  httpRequest.setRequestHeader('Accept', 'application/json');
+  httpRequest.send(file);
```
garrensmith: What is the advantage of using `XMLHttpRequest` over fetch to send the file?

app/addons/documents/resources.js
```diff
@@ -85,18 +86,15 @@ Documents.DdocInfo = FauxtonAPI.Model.extend({
...
-    var uuid = new Documents.UUID();
-    var deferred = this.deferred = $.Deferred();
-    var that = this;
-
-    uuid.fetch().done(function () {
-      that.set("_id", uuid.next());
-      deferred.resolve();
+    const uuid = new Documents.UUID();
+    const promise = new FauxtonAPI.Promise((resolve) => {
+      uuid.fetch().done(() => {
```
garrensmith: Could we use `then` instead of done?

garrensmith: Ideally, we need the UUID generation to be extracted out of the model and just be a nice function.

app/addons/documents/resources.js
```diff
@@ -120,31 +118,26 @@ Documents.BulkDeleteDocCollection = FauxtonAPI.Collection.extend({
...
-      that.handleResponse(res, promise);
-    }).fail(function () {
-      var ids = _.reduce(that.toArray(), function (acc, doc) {
-        acc.push(doc.id);
-        return acc;
-      }, []);
-      that.trigger('error', ids);
-      promise.reject(ids);
+    const payload = this.createPayload(this.toJSON());
+    const promise = new FauxtonAPI.Promise((resolve, reject) => {
```
garrensmith: I think you can just do `return new FauxtonAPI.Promise`...

app/addons/documents/resources.js
```diff
@@ -120,31 +118,26 @@ Documents.BulkDeleteDocCollection = FauxtonAPI.Collection.extend({
...
+      post(this.url(), payload).then(res => {
+        if (res.error) {
```
garrensmith: Do you think throwing our own error is the best way of handling this? Could we not just handle it here where we know the error has happened?

app/addons/documents/doc-editor/actions.js
Antonio-Maranhao: Reason was that we can't get the upload progress using fetch, so I had to use `XMLHttpRequest`.

app/addons/documents/resources.js
Antonio-Maranhao: The issue is unlike `$.ajax()`, `fetch` doesn't raise an error for 4xx, 5xx codes. I could handle errors inside `then()` as you suggested, but I should also add a `catch()` to handle `fetch()` errors as well (e.g. connection lost).
Considering both types of errors are handled pretty much in the same way, I found the easiest is to raise an Error and keep error handling inside `catch()` because it seems more logical.

Antonio-Maranhao:
@garrensmith I've addressed all comments. Can you please take a look ?

Antonio-Maranhao:
@garrensmith finally got the tests to pass.

PR #1066, opened 2018-03-10
Bring back compaction addon

popojargo:
## Overview

Bring back the compaction addong (revert + update code to use Redux)

### Database compaction
1. Create a database
2. Click on the database
3. Click on the clean & compact side bar button
4. Click Compact database
5. Click cleanup view

### Compact view

1. Create a new view
2. Edit the view
3. Compact the view (button at bottom right)

## Checklist

## Preview

compactview
compactview_


wohali:
-1.

CouchDB 2.1.x has automatic compaction enabled by default. Manual controls in Fauxton may encourage collision between the requested compaction, and the automatic compaction facility (which automatically determines the right priority and order for compactions to occur in, and ensures multiple compactions are not running at once).

I think the time for exposing this in a friendly UI is long past. Instead, we might want to find a way to hold recent compactions in Couch somewhere, along with the upcoming scheduled compactions, and allow these to be exposed within the Fauxton UI somehow.

PR #1070, opened 2018-04-03
Replication redux

garrensmith:
Convert the replication page to use Redux


app/addons/replication/controller.js
```diff
@@ -245,18 +194,18 @@ export default class ReplicationController extends React.Component {
   }
 
   onTabChange (section, url) {
-    Actions.changeTabSection(section, url);
+    this.props.changeTabSection(section, url);
```
Antonio-Maranhao: This line is throwing an error because the `changeTabSection` prop is not being set.

app/addons/replication/reducers.js
```diff
@@ -0,0 +1,381 @@
...
+    case ActionTypes.REPLICATION_FETCHING_REPLICATE_STATUS:
+      return {
+        ...state,
+        fetchingReplicateInfo: true,
+      };
+
+    case ActionTypes.REPLICATION_REPLICATE_STATUS:
+      return {
+        ...state,
+        fetchReplicateInfo: false,
```
Antonio-Maranhao: Typo. It should be `fetchingReplicateInfo`

app/addons/replication/actions.js
```diff
@@ -248,8 +233,8 @@ export const deleteDocs = (docs) => {
         clear: true
       });
 
-      clearSelectedDocs();
-      getReplicationActivity();
+      dispatch(clearSelectedDocs());
+      dispatch(getReplicationActivity());
     })
     .catch(resp => {
       resp.json()
```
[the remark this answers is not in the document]
garrensmith: It does in this case because I passed in `raw: true` as a param to the post. It then doesn't call the resp.json() for you.

PR #1071, opened 2018-04-03
Only use update and stable query params when set

Antonio-Maranhao:
## Overview

Document list doesn't work with older CouchDB versions that don't support the `update` and `stable` parameters because Fauxton always adds the parameters as part of the GET requests. 

The proposed fix is to only send these params when they're set in the Query Options panel to a value other than their default.

## Testing recommendations

- Go to the All Documents list
- Using the network tab in the browser's devtools, check the `update` and `stable` parameters are no longer being sent.
- Open the Query Options panel. Change the values of `update` and `stable` parameters.
- Check the params are now added to the GET requests

## Checklist

### DIFF
--- a/app/addons/documents/__tests__/reducers.test.js
+++ b/app/addons/documents/__tests__/reducers.test.js
@@ -482,5 +482,51 @@ describe('Docs Reducers', () => {
         group_level: 2
       });
     });
+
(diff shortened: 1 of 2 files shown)


rnewson:
+1

PR #1072, opened 2018-04-04
Fix query options reducer code

Antonio-Maranhao:
## Overview

For unknown reasons the code generated by the build task (`bundle.js`) behaves different than the original code. Using a build after #1071, the parameters `update` and `stable` were still being set. 

This PR submits an alternative implementation for the changes submitted in #1071, which from local tests generates a `bundle.js` that works correctly.

## Testing recommendations

Same as #1071 

## Checklist

### DIFF
--- a/app/addons/documents/index-results/reducers.js
+++ b/app/addons/documents/index-results/reducers.js
@@ -325,18 +325,11 @@ export const getQueryOptionsParams = (state) => {
   // Only add UPDATE and STABLE parameters when different than
   // their respective default values. This prevent errors in
   // older CouchDB versions that don't support these parameters.
-  if (queryOptionsPanel.update !== undefined) {
-    // Default value is 'true'
-    if (queryOptionsPanel.update !== 'true') {
-      params.update = queryOptionsPanel.update;
-    }
+  if (queryOptionsPanel.update !== undefined && queryOptionsPanel.update !== 'true') {
+    params.update = queryOptionsPanel.update;
   }
-
-  if (typeof queryOptionsPanel.stable === 'boolean') {
-    // Default value is false
-    if (queryOptionsPanel.stable === true) {
-      params.stable = queryOptionsPanel.stable;
-    }
+  if (queryOptionsPanel.stable === true) {
+    params.stable = queryOptionsPanel.stable;
   }
 
   return params;


popojargo:
1. There are a lot of useless files that I don't think you wanted to push
2. Maybe we should track down what is that "unknown" behavior before patching?

What do you think? @Antonio-Maranhao

Antonio-Maranhao:
Thanks @popojargo - that was a very bad rebase I did. It's fixed now. 

By 'unknown' I meant I don't know the reason why the minified code generated in `bundle.js` was not semantically equivalent to the original code.
The original `getQueryOptionsParams` in #1071 is :
```
export const getQueryOptionsParams = (state) => {
  const {queryOptionsPanel} = state;
  const params = {};

  if (queryOptionsPanel.includeDocs) {
    params.include_docs = queryOptionsPanel.includeDocs;
  }

  if (queryOptionsPanel.showBetweenKeys) {
    const betweenKeys = queryOptionsPanel.betweenKeys;
    params.inclusive_end = betweenKeys.include;
    if (betweenKeys.startkey && betweenKeys.startkey != '') {
      params.start_key = betweenKeys.startkey;
    }
    if (betweenKeys.endkey && betweenKeys.endkey != '') {
      params.end_key = betweenKeys.endkey;
    }
  } else if (queryOptionsPanel.showByKeys) {
    if (queryOptionsPanel.byKeys.trim()) {
      params.keys = queryOptionsPanel.byKeys.replace(/\r?\n/g, '');
    }
  }

  if (queryOptionsPanel.limit !== 'none') {
    params.limit = parseInt(queryOptionsPanel.limit, 10);
  }

  if (queryOptionsPanel.skip) {
    params.skip = parseInt(queryOptionsPanel.skip, 10);
  }

  if (queryOptionsPanel.descending) {
    params.descending = queryOptionsPanel.descending;
  }

  if (queryOptionsPanel.reduce) {
    params.reduce = true;

    if (queryOptionsPanel.groupLevel === 'exact') {
      params.group = true;
    } else {
      params.group_level = queryOptionsPanel.groupLevel;
    }
  }

  // Only add UPDATE and STABLE parameters when different than
  // their respective default values. This prevent errors in
  // older CouchDB versions that don't support these parameters.
  if (queryOptionsPanel.update !== undefined) {
    // Default value is 'true'
    if (queryOptionsPanel.update !== 'true') {
      params.update = queryOptionsPanel.update;
    }
  }

  if (typeof queryOptionsPanel.stable === 'boolean') {
    // Default value is false
    if (queryOptionsPanel.stable === true) {
      params.stable = queryOptionsPanel.stable;
    }
  }

  return params;
};
```

And the corresponding code in `bundle.js` is: 
```
getQueryOptionsParams=function(e){
  var t=e.queryOptionsPanel,a={};
  if(t.includeDocs&&(a.include_docs=t.includeDocs),t.showBetweenKeys){var n=t.betweenKeys;a.inclusive_end=n.include,n.startkey&&""!=n.startkey&&(a.start_key=n.startkey),n.endkey&&""!=n.endkey&&(a.end_key=n.endkey)} else t.showByKeys&&t.byKeys.trim()&&(a.keys=t.byKeys.replace(/\r?\n/g,""));return"none"!==t.limit&&(a.limit=parseInt(t.limit,10)),t.skip&&(a.skip=parseInt(t.skip,10)),t.descending&&(a.descending=t.descending),t.reduce&&(a.reduce=!0,"exact"===t.groupLevel?a.group=!0:a.group_level=t.groupLevel),void 0!==t.update&&(a.update=t.update),"boolean"==typeof t.stable&&(a.stable=t.stable),a}
```

Using `getQueryOptionsParams({queryOptionsPanel: {includeDocs: true, stable: false, update: "true"}})` as an example, the original code returns `{include_docs: true, limit: NaN}` but the version in `bundle.js` returns `{include_docs: true, limit: NaN, update: "true", stable: false}`.

Antonio-Maranhao:
@popojargo I opened #1073 to track the issue separately.

PR #1075, opened 2018-04-25
Replication paginate

garrensmith:
## Overview

This adds the ability to paginate through your replications in the replication tab. I've only added pagination to the Replication DB's tab since I can't think of a way of replicating through the _replicate replications. The _replicate replication uses the /jobs endpoint which is a mixture of replication DB and _replicate results.

## Testing recommendations

Create a few replications and then paginate.

## GitHub issue number

#1022


app/addons/replication/api.js
```diff
@@ -295,10 +297,16 @@ export const combineDocsAndScheduler = (docs, schedulerDocs) => {
...
+  const params = {
+    limit: docsPerPage + 1,
+    skip: (page - 1) * docsPerPage,
+    include_docs: true
+  };
+
   return supportNewApi()
     .then(newApi => {
-      const docsPromise = get('/_replicator/_all_docs?include_docs=true&limit=100')
+      const docsPromise = get(`/_replicator/_all_docs?startkey=%22_designZ%22&${app.utils.queryParams(params)}`)
```
popojargo: Can I know what is this suppose to do? When I run locally your branch, I don't have anything in my replicator activity and I suspect this line to the be the cause.

garrensmith: This fetches all the docs in the replicator db except the design doc.

Antonio-Maranhao: I agree with @popojargo - because of the `startkey` param, the query won't return any docs whose ID starts with a number or uppercased letter.

popojargo: The only viable solution I see is splitting the request in two API calls. One to fetch documents before the _design (?endkey="_design*") and another one to fetch documents after design docs (startkey="_designZ"

popojargo:
@garrensmith  Do you plan on finishing this or do you want someone else to complete the work?

wohali:
Hi @garrensmith @popojargo @Antonio-Maranhao , 

Just ran into this one. Any chance this PR can be freshened and merged for a new release of Fauxton?

popojargo:
I'll keep that in mine but my schedule is overloaded atm

Antonio-Maranhao:
Same for me, unfortunately. IIRC, it still requires reworking how the docs are retrieved because `/_replicator/_all_docs?startkey=%22_designZ%22` doesn't return any docs that start with lowercase or digits. 
@popojargo suggested above using two API calls but that will likely make it harder to paginate correctly. 
@garrensmith suggested using a Mango query instead. That's probably the best approach. Something like this would do:
```
{
   "selector": {
      "$or": [
         {
            "_id": {
               "$lt": "_design"
            }
         },
         {
            "_id": {
               "$gt": "_design0"
            }
         }
      ]
   }
}
```

garrensmith:
I don't remember suggesting a mango query but that does sound like the best option. Mango indexes automatically do not include _design docs so you can just do:
```
_id: {"$gte: null}
```

wohali:
Closing in favour of #1288 . If I have misunderstood, please re-open.

PR #1081, opened 2018-05-02
Extension for replication authentication methods

Antonio-Maranhao:
## Overview

Provide a mechanism to define extra authentication methods when creating/editing a replication. Separate add-ons can then use this mechanism to get/set custom authentication information to a new/existing replication document.

The Create/Edit replication form was also modified to include fields where the user can explicitly select the authentication method and corresponding credentials. As a consequence, the Enter Password modal was removed since the info is now capture directly on the form.

## Testing recommendations

- Create/edit replications using the built-in username/password method

## Checklist


app/addons/replication/actions.js
```diff
@@ -48,7 +49,7 @@ export const getDatabasesList = () => dispatch => {
     });
 };
 
-export const replicate = (params) => dispatch => {
+export const  replicate = (params) => dispatch => {
```
garrensmith: unnecessary space

PR #1085, opened 2018-05-28
Use a supported CouchDB docker image for selenium tests

wohali:
This change mirrors https://github.com/pouchdb/pouchdb/pull/6936 which moved PouchDB off of the obsolete `couchdbdev/debian-8-dev` image (which only has CouchDB 2.0.0 in it!)

I haven't tested this, but Travis should do so for me, if I read `.travis.yml` and `bin/docker-up-and-check.sh` and `package.json` correctly.

@garrensmith I know you're looking for an image that tracks CouchDB master, and that's coming, I promise! But before then I want to ensure you're at least tracking our latest public release. I also want to get rid of these unmaintained images that do no one any favours.

Once this lands I will remove the `couchdbdev/debian-8-dev` image from Docker Hub.

### DIFF
--- a/docker/dc.selenium.yml
+++ b/docker/dc.selenium.yml
@@ -8,7 +8,9 @@ services:
 
   couchdb:
     container_name: couchdb
-    image: couchdbdev/debian-8-dev
-    command: '--with-haproxy -a tester:testerpass -n 1'
+    image: apache/couchdb:latest
+    environment:
+      COUCHDB_USER: tester
+      COUCHDB_PASSWORD: testerpass
     ports:
       - "5984:5984"
--- a/test/nightwatch_tests/helpers/helpers.js
+++ b/test/nightwatch_tests/helpers/helpers.js
@@ -13,16 +13,30 @@
 var nano = require('nano');
 var async = require('async');
 
+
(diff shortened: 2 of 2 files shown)


wohali:
Need some help from someone to diagnose the build failure here. It _should_ just work... (famous last words, I know!)

garrensmith:
I've just restarted the tests. The changes look good. The tests can be a little flaky

popojargo:
I ran the tests locally and the same test failed. There's probably something to fix

popojargo:
I think that the errors is caused by the missing _users table. With the current configuration, a full cluster is already configured. This actually let us test the setup view but it has impacts on all other tests.

@garrensmith @Antonio-Maranhao Do we simply check for the required databases in the **Before** of each nightwatch test ?

Or do you prefer to setup automatically couchdb on startup?

wohali:
@popojargo got it! Can your setup script do a PUT http://localhost:5984/_users and http://localhost:5984/_replicator (ignoring if they're already there) before it proceeds? That should solve the problem.

The current Docker image doesn't pre-create these because it's designed to work either standalone or in a cluster.

PR #1095, opened 2018-06-22
Fix - Create replication fails when db doesn't exist

Antonio-Maranhao:
## Overview

The `_replicator` db was not being automatically created when a new replication is created.

The test was also fixed because it was failing for this scenario.

## Testing recommendations

- Delete `_replicator` db
- Create a new replication
- Operation succeeds

## Checklist


popojargo:
Question: Why should it be the responsability of Fauxton to create the database? Usually when you setup your Couch, it should automatically create the databases. I don't think we can't just add the _replicator database and omit the _users and other system databases.

As for Fauxton dev environment, we switched to docker recently which force use to create system databases manually. Maybe you just want to tweak this instead?

rnewson:
This PR mixes substantive changes with cosmetic (whitespace) ones. Can you update to remove the latter and rebase it to remove the merge from master too pls.

rnewson:
I'm +1 on the intention of the patch but I can't verify that your change is the right way to do it.

popojargo:
What's the reason for closing this? @Antonio-Maranhao

Antonio-Maranhao:
@popojargo sorry, my brain must have short-circuited. I meant to merge, not close it.

PR #1103, opened 2018-07-10
Customize style of the results list

Antonio-Maranhao:
## Overview

Adds a new menu to customize the style of the results list.

 - Truncated/Show full values
 - Font size: small, medium, large

image

## Testing recommendations

 - Go to documents list and switch to the Metadata view
 - Click the new button to show the menu
 - Click on `Show full values`
   - The full text value is displayed for each row
   - Menu label changes to 'Truncate values'
 - Click on `Small`, `Medium`, `Large` and verify the font size is updated
 - Switch to the Table view, and verify the same options also work
 - Switch to the JSON view, the menu should only show the 'Font size' options

## Checklist


wohali:
Love the idea, not crazy about the verbose checkbox for "Truncate values."

I'd suggest inverting the sense of the checkbox, making it "Expand values" instead?

Antonio-Maranhao:
@wohali @popojargo I had some time to rework this. What do you think?

"The good, old Fauxton that we know and love..."
image

-"But it'd be great to see the full values"
-"Sure thing"
image

-"Too much scrolling... MORE DENSITY"
-"Okie dokie"
image

-"Argh... I need new glasses. Magnify, pls"
-"aye aye captain"
image

popojargo:
Wow looks good! 👍

Antonio-Maranhao:
@popojargo I pushed all the changes, could you pls review it?

app/addons/documents/index-results/reducers.js
```diff
@@ -61,9 +62,34 @@ const initialState = {
...
   switch (action.type) {
 
+    case ActionTypes.INDEX_RESULTS_SET_STYLE:
+      const newStyle = {
+        ...state.resultsStyle,
+        ...action.resultsStyle
+      };
+      storeStyle(newStyle);
+      return Object.assign({}, state, {
+        resultsStyle: newStyle
```
popojargo: Why don't you use the spread operator here as well?

app/addons/documents/index-results/containers/IndexResultsContainer.js
```diff
@@ -56,7 +58,8 @@ const mapStateToProps = ({indexResults}, ownProps) => {
     docType: getDocType(indexResults),
     fetchParams: getFetchParams(indexResults),
     queryOptionsParams: getQueryOptionsParams(indexResults),
-    partitionKey: ownProps.partitionKey
+    partitionKey: ownProps.partitionKey,
```
popojargo: This should be in another PR right?

app/addons/documents/components/results-toolbar.js
```diff
@@ -83,5 +97,7 @@ ResultsToolBar.propTypes = {
   isLoading: PropTypes.bool.isRequired,
   hasResults: PropTypes.bool.isRequired,
   isListDeletable: PropTypes.bool,
-  partitionKey: PropTypes.string
+  partitionKey: PropTypes.string,
```
popojargo: Unrelated?

app/addons/documents/components/results-toolbar.js
```diff
@@ -13,12 +13,18 @@ import PropTypes from 'prop-types';
...
 import BulkDocumentHeaderController from "../header/header";
+import ResultsOptions from './results-options';
 import Components from "../../components/react-components";
+import Constants from '../constants';
 import Helpers from '../helpers';
 
 const {BulkActionComponent} = Components;
 
 export class ResultsToolBar extends React.Component {
+  constructor (props) {
```
popojargo: Useless constructor

app/addons/documents/components/results-options.js
```diff
@@ -0,0 +1,96 @@
...
+      fontSize: size
+    });
+  }
+
+  getDensitySection() {
+    const densityItems = [{
+      title: 'Show full values',
+      onClick: this.toggleTextOverflow
+    }];
+    if (this.props.resultsStyle.textOverflow === Constants.INDEX_RESULTS_STYLE.TEXT_OVERFLOW_FULL) {
```
popojargo: I think you should define the title before creating the densityItem. It's more error prone if you mutate the array at a certain index after its creation.

app/addons/documents/assets/less/index-results.less
```diff
@@ -34,6 +34,39 @@
...
   }
+
+  .text-overflow-switch {
+    min-width: 145px;
+    min-height: 26px;
+    padding: 8px 0;
+  }
+
+  .toolbar-dropdown {    
+    // padding-top: 10px;
```
popojargo: I guess we can remove those comments

app/addons/documents/__tests__/table-row.test.js
```diff
@@ -54,4 +56,34 @@ describe('Docs Table Row', () => {
...
+      docChecked={sinon.stub()}
+      el={elem}
+      data={data}
+      index={0}
+      docIdentifier={elem.id}
+      isSelected={false}
+      textOverflow={Constants.INDEX_RESULTS_STYLE.TEXT_OVERFLOW_TRUNCATED}
+    />);
+
+    expect(wrapper.find('td.showall').exists()).toBe(false);
```
popojargo: can this be replaced by `expect(wrapper.find('td.showall')).to.not.exist;` ?

Antonio-Maranhao:
Thanks @popojargo. 
There's no special config. The button should always be visible from All Docs or any view. Do you see any errors in the console?

popojargo:
@Antonio-Maranhao My bad, I had a problem with my local setup.

The icon doesn't look centered (there's a tiny margin): https://puu.sh/BXtqQ/d938cf448b.gif 

Also, I find it a bit weird that the button keeps moving from right to left depending on what's the view.

Ex: https://puu.sh/BXtnP/a251f74de6.png

The feature works fine on Chrome, Edge and IE11 just in case you're wondering ^^

app/addons/documents/__tests__/table-row.test.js
Antonio-Maranhao: No because `find()` will always return a Wrapper even when there is no match. You need to either call `exists()` or check if `length` is zero.

app/addons/documents/components/results-toolbar.js
Antonio-Maranhao: Removed

app/addons/documents/components/results-options.js
Antonio-Maranhao: Agree... it's changed.

app/addons/documents/assets/less/index-results.less
Antonio-Maranhao: Removed

app/addons/documents/components/results-toolbar.js
Antonio-Maranhao: Yes, the `partitionKey` was introduced in another PR

app/addons/documents/index-results/containers/IndexResultsContainer.js
Antonio-Maranhao: It was added in another PR. It's only highlighted because of the added comma at the end of the line.

app/addons/documents/index-results/reducers.js
Antonio-Maranhao: Updated... using the spread op in all actions

PR #1106, opened 2018-07-18
Fix how local creds are checked when creating replications

Antonio-Maranhao:
## Overview

When creating/editing replications, the code for credentials check is using `app.host` directly which is not a problem in the dev environment, but it produces errors in the release build.

## Testing recommendations

- Create a new replication using local a database with user/password as either source or target.
- The replication is created successfully

## Checklist

### DIFF
--- a/app/addons/replication/components/newreplication.js
+++ b/app/addons/replication/components/newreplication.js
@@ -12,7 +12,7 @@
 
 import base64 from 'base-64';
 import React from 'react';
-import app from '../../../app';
+import Helpers from '../../../helpers';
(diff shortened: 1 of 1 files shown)


popojargo:
👍  if travis pass. I should have paid more attention to the PR after my relative url change.

Antonio-Maranhao:
@popojargo no worries - my mistake as well for not double checking other uses of `app.host`. 
There's only one left in `addons/setup/base.js` which I assume it's related to the issue you already mentioned.
Thanks!

Issue #1117, opened 2018-08-03
Can only log on in IE11 when deubgger options is open


RSchade:
Another IE11 specific issue...

On 2.2.0RC3 I can log in fine if I have the debugger console open, but it will just refresh the page and not log me in when the debugger console is closed. Logging out doesn't work regardless. I don't see any errors in the debug console but I'm using Fauxton in production mode.

popojargo:
I'm having the same issue on my side. I tried to debug this lately but didn't the specific source of the problem. I can guarantee that I will have time to fix this in the next days. If anyone else can take this issue it would be appreciated.

wohali:
Closed by #1119 .

PR #1119, opened 2018-08-05
Refactor to use Ajax helper everywhere + no-cache

popojargo:
## Overview

For IE11, every API requests were cached. This cause a lot of issues as it's impossible to log in, sign out or get updated values when navigating through Fauxton.

I refactored the code to use the ajax helper for every API call.
I also added the no-cache headers in the helpers so the API calls are not cached.

## GitHub issue number

#1117 

## Checklist

### DIFF
--- a/app/addons/auth/api.js
+++ b/app/addons/auth/api.js
@@ -12,43 +12,12 @@
 
 import app from './../../app';
 import Helpers from "../../helpers";
-import { defaultsDeep } from "lodash";
-
-export const json = (url, opts = {}) => fetch(
-  url,
-  defaultsDeep(
-    {
-      credentials: "include",
-      headers: {
-        accept: "application/json",
-        "Content-Type": "application/json"
-      }
-    },
-    opts
-  )
-).then(resp => resp.json());
-
-export const formEncoded = (url, opts = {}) => fetch(
-  url,
-  defaultsDeep(
-    {
-      credentials: "include",
-      headers: {
-        accept: "application/json",
-        "Content-Type": 'application/x-www-form-urlencoded;charset=UTF-8'
-      }
-    },
-    opts
-  )
-).then(resp => resp.json());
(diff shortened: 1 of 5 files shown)


wohali:
Do you think this is worth respinning 2.2.0 for? I don't know if IE even worked for 2.1.2....

popojargo:
@wohali With 2.1.2, it was partially running. I know that we want to make a release soon but we have known issues with Fauxton that can be easily fixed (replication relative url and IE11 support). 

Users can still update manually Fauxton after the release but this is not documented/recommended.

popojargo:
@wohali Should we fix the relative url issue with the replication before creating a new tag?

wohali:
@popojargo do you have a fix? also can you come to IRC, we can discuss this faster :)

popojargo:
@wohali I can probably fix that tomorrow and my IRC isn't working. Do you have slack?

wohali:
@popojargo no, dev chat is on IRC only. sorry :( Guess I'll have to kill this release, I'll email the list... :( :(

wohali:
@popojargo Hope your exam is going well! Neither Garren nor Antonio were able to get to this today. 

You're my only hope. Any chance of this being fixed in the next 12 hours?

gdelfino:
IE11 user here (not by choice). I just wast to point out that we use Photon as an alternative front end for CouchDB. I find that the most useful feature of it is that with a click we can update it. So in other words, when a bug is found, it can be solved and distributed very quickly. On the other hand, if a bug is found in Fauxton after CouchDB 2.2.0 is released, then we need to wait a long time for an update. I am just voting here in favor of being able to update the database and the front-end separately.

popojargo:
@wohali Will do this pm

wohali:
@gdelfino this isn't the place, please file a new ticket, but this should be possible if you use npm.

janl:
#offtopice @gdelfino you can run `npm install -g fauxton` any time and get new bits.

PR #1128, opened 2018-09-28
Partitioned dbs partition key selector

Antonio-Maranhao:
## Overview

Second in a series of PRs that will be submitted in support of the new user-defined partitioned databases feature (https://github.com/apache/couchdb/pull/1605).

Summary:
 - Add new PartitionKeySelector component that allows setting a partition key for filtering query results.
 - The selected partition key is represented in the application URL. E.g.: `/#/database/mydb/_partition/part1/_design/ddoc1/_view/new-view`
 - Although the URL is being updated by the Selector, the query results are not affected yet. This will come in a future PR

## Testing recommendations

 - Create a new partitioned database
 - Navigate to 'All Documents'. The Selector should be inactive (gray) and say 'No partition selected'
 - Click the Selector and type the value `part1` then press Enter
 - The URL should change to `/#/database/<DB>/_partition/part1/_all_docs`
 - Click the Selector's (filter) icon to switch back to 'global' mode. The URL should change to back to `/#/database/<DB>/_all_docs`
 - Navigate to a view and repeat the same steps as for 'All Documents'

## Related Pull Requests

https://github.com/apache/couchdb/pull/1605

## Checklist

### DIFF
--- a/app/addons/documents/assets/less/partition-key.less
+++ b/app/addons/documents/assets/less/partition-key.less
@@ -1,0 +1,59 @@
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License. You may obtain a copy of
+// the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+
+@import "../../../../../assets/less/variables.less";
+
+.partition-selector {
+  display: flex;
+  align-items: center;
+  padding: 0;
+  min-height: 30px;
+}
+
+// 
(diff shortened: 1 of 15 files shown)


app/addons/documents/assets/less/partition-key.less
```diff
@@ -0,0 +1,59 @@
...
+
+.partition-selector {
+  display: flex;
+  align-items: center;
+  padding: 0;
+  min-height: 30px;
+}
+
+// 
+//   line-height: 16px;
```
garrensmith: Can you remove these?

app/addons/documents/base.js
```diff
@@ -49,6 +51,12 @@ FauxtonAPI.registerUrls('allDocs', {
   }
 });
 
+FauxtonAPI.registerUrls('partitioned_allDocs', {
+  app: function (databaseName, partitionKey, query) {
+    return 'database/' + databaseName + '/_partition/' + partitionKey + '/_all_docs' + getQueryParam(query);
```
garrensmith: We don't support partitions for _all_docs. So we don't need this.

garrensmith:
+1 nice work.

Issue #1098, opened 2018-06-27
Auto-check Mango $regexes for regex syntax validity


wohali:
See https://github.com/apache/couchdb/issues/1415 for an example. The response from CouchDB is a bit opaque.

Perhaps we can do better? All we need to do is cram the value of the `$regex` into a RegExp and trap on errors (though groups and escaping will need a bit of massaging...)

```
>> var re = new RegExp("F4647868\+\+\+");
SyntaxError: nothing to repeat
>> var re = new RegExp("F4647868\\+\\+\\+");
undefined
>> var re = new RegExp("(?i)F4647868\\+\\+\\+");
SyntaxError: invalid regexp group
```

PR #1137, opened 2018-10-10
Validate regex on client side.

popojargo:
## Overview

I added client side validation for the regex. So I extract all the $regex pairs through a recursive function and create them. I catch any errors and show it to the user.

I also added some JSON parsing validation. While testing, I found that the editor was not recognizing invalid patterns such as "\_"  -> The \_ is illegal.

Here's the output when an invalid regex is evaluated:
image

## GitHub issue number

#1098 

## Checklist

### DIFF
--- a/app/addons/documents/mango/components/MangoQueryEditor.js
+++ b/app/addons/documents/mango/components/MangoQueryEditor.js
@@ -19,6 +19,7 @@ import app from "../../../../app";
 import FauxtonAPI from "../../../../core/api";
 import ReactComponents from "../../../components/react-components";
 import ExecutionStats from './ExecutionStats';
+import Helper from '../mango.helper';
 
 const PaddedBorderedBox = ReactComponents.PaddedBorderedBox;
 const CodeEditorPanel = ReactComponents.CodeEditorPanel;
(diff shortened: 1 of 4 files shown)


app/addons/documents/mango/components/MangoQueryEditor.js
```diff
@@ -130,16 +131,54 @@ export default class MangoQueryEditor extends Component {
...
+        clear: true
+      });
+    }
+  }
+
+  isRegexValid(selector = {}) {
+    const regexes = Helper.getSelectorRegexes(selector);
+    const errors = _.reduce(regexes, (acc, val) => {
+      try {
+        new RegExp(val);
```
willholley: it's worth bearing in mind that Query uses [Regex engine](http://erlang.org/doc/man/re.html) that is not perfectly compatible with JavaScript's. This check will allow through regex strings that will not work in Query and there may be regexes that are valid in Query but not valid JS regexes.

Is there a particular use case that led to this PR? For instance, do we need better server-side errors in the event that a bad regex is passed in a selector?

popojargo: See the linked issue. I guess the best solution would be to have more human friendly error message when having regex errors. If that's not possible, I could allow to send wrong regex to the server and use my code to return more detailed error message in case of failure

popojargo:
@wohali Will pointed out a good point for the regex validation. Should we fix the error formatting at the server side instead?

app/addons/documents/mango/components/MangoQueryEditor.js
wohali: @willholley @popojargo Right, the idea is that we need much better error messages from Couch when a regex fails, and that many ultra-simple syntax errors can be caught in client-side directly - for instance having sufficient `/` count.

I appreciate the "impedence mismatch" between Erlang re and JavaScript regexes being a concern. The ultimate goal is to enable people working in Fauxton to find their errors more readily, and for the editor provide some sort of syntax highlighting or in-line guide to what might be wrong vs. having to read a likely truncated, obscure error message in the warning bar at the top after submission.

Suggestions are very welcome here :) Personally, I'm reacting to requests for improvement from IRC/Slack support for basic things like regex syntax highlighting and error parsing, which seem like low hanging fruit, even if we take a naive approach. What do you think?

popojargo: Adding a custom highlighter for the $regex key-value pairs in the editor looks complicated to me. Although, even with a good highlighter, the error will still be vague.

So I guess the first thing could be to find a way to have more expressive error messages.

My two suggestions:

- Reformat the Regex error message (display the letters as char, not charCode).
- On a query error, we can check the server error and display an appropriate error message if it's caused by regex errors (I can reverse engineer the request and detect Regex errors with the JavaScript regex engine.

wohali:
@popojargo Why did you close this? is a new PR coming for this?

popojargo:
@wohali Not planning on doing another PR. I think the error message should be more detailed at the backend level since the regex errors are related to the Erlang Regex engine.

wohali:
Then I suggest you file an issue in apache/couchdb, because otherwise we'll lose track of this :)

popojargo:
There's still that open issue here: https://github.com/apache/couchdb-fauxton/issues/1098

I'm taking the little time I have free to fix complete Fauxton pagination and fix the build :)

PR #1141, opened 2018-10-15
Convert activetasks to redux

garrensmith:
## Overview

Convert active tasks to use redux

## Testing recommendations
Create some active tasks. 
Checkout the active tasks tab. Make sure everything works.

## Checklist


app/addons/activetasks/actions.js
```diff
@@ -9,72 +9,79 @@
...
     });
-  },
-  runPollingUpdate (collection) {
-    collection.pollingFetch().then(() => {
-      FauxtonAPI.dispatch({
-        type: ActionTypes.ACTIVE_TASKS_POLLING_COLLECTION,
-        options: collection.table
+};
+
+export const fetchAndSetActiveTasks = (collection, backboneCollection) => {
```
Antonio-Maranhao: `fetchAndSetActiveTasks` is not being used anywhere, so it can be removed

Issue #871, opened 2017-03-16
Copied & pasted string with newlines should show JSON syntax error in Ace editors


garrensmith:
Copying & pasting in text with newlines into a string in the full page doc editor (and probably any Ace editor field) should show an error icon in the gutter, but it doesn't.

Taken from https://issues.apache.org/jira/browse/COUCHDB-2757

micheleriva:
Is help still needed here? What's the proposed behavior? Should `\n`s to be removed or encoded?

Antonio-Maranhao:
@micheleriva The issue still happens. I don't think we should convert the pasted text though. The editor itself should detect the invalid syntax and highlight it. 
For the time being, the workaround is to open the String Editor and paste the text with newlines there. Fauxton will convert any white space chars to the corresponding `\n`, `\t`, etc.

FYI, the latest Ace (v1.4.0) shows the error - you can try at https://ace.c9.io/build/kitchen-sink.html.
Unfortunately Fauxton uses `brace` which still relies on Ace v1.2.9 (latest is 1.4.0). Hopefully it gets updated soon and it'll solve the issue here.

micheleriva:
Ok thanks for the quick response. I am investigating but it seems that Ace editors creates a new `div` for each pasted line of code, so it is really difficult to detect that kind of behavior.
One ugly solution would be to add an [onpaste](https://www.w3schools.com/jsref/event_onpaste.asp) event and then detect any `\n`, `\t` etc, but I personally don't like that kind of approach.
If you think that some more investigation is worth the time, I can keep on. But if you think that it is better to wait until Ace will be updated, I can stop here 😊

Antonio-Maranhao:
No longer happens.

PR #1145, opened 2018-10-19
#871 added error message on 'paste' event if json schema is not valid

micheleriva:
## Overview

Added a quick fix for issue #871


wohali:
@micheleriva Hey, given how stale this PR is, and the confict, I'm going to close it. If you are interested in reviving it, please open a new one. Thanks!

Issue #1086, opened 2018-05-29
Enhancement: Apply filters on _all_dbs


LyteFM:
When using superlogin or the `couch_peruser` - option, it is often necessary to investigate a small subset of a large set of databases. My scripts access `_all_dbs` and then filter by name or by specific documents or keys on documents that exist in many databases. 
In Fauxton, it's only possible to select one database by name or to browse all by page. It would be great if some of the following was possible in future releases:

- Regular expression filtering on the database name

- Filtering based on the existence of a document (by name/ regex on name) in each database

- Filtering based on document keys (if the doc exists) in each database

The first point would already help a lot. And many thanks for your work on CouchDB/ Fauxton!

popojargo:
I think the only viable propositions so far is:

1. Allow prefix wilcards filtering (with startkey param) (We could also do suffix but meh?)

Other solutions won't scale altough they could be a nice addon.

popojargo:
@Antonio-Maranhao We recently discussed about this feature and I would like to work on it. Do you have any idea how I could integrate the search feature into the actual select? 

Here's some ideas:

- Support wildcards (eg: database_prefix*)
- Keyboard shortcuts. Let's say you enter client/, by pressing Enter, it opens the database "client/". Otherwise, if you press Ctrl+Enter, it list databases prefixed by client/.

Antonio-Maranhao:
@popojargo Merging the two actions in the same component will likely make things a bit confusing for the user. What about we split it like in the screenshots below: 

The search we have today is active by default:
![image](https://user-images.githubusercontent.com/30349380/47727211-5fe32f80-dc32-11e8-887b-db5ecc51ba86.png)

Clicking the filter icon would switches to a text input - hopefully with a nice animation to show/hide them 😉 - then you type the db prefix and press enter to filter the db list below
![image](https://user-images.githubusercontent.com/30349380/47730116-e51d1300-dc37-11e8-858c-67444106f2ca.png)

kripper:
The "Database name" search box (to select a single database) can be replaced by a database search box (to filter databases). This is the google "standard". 
This feature would also solve the requirement described in https://github.com/apache/couchdb-fauxton/issues/869.

PR #1146, opened 2018-10-21
Add grouping to databases with IDs containing '/'

### LINKED ISSUE #869: Group databases with / char
Currently databases with / char in name showed as flat list. While it's ok, but the main trick of having this char in db name is to group the databases on filesystem. It would be awesome if Fauxton will also group them in UI, providing fold/unfold buttons.

Taken from https://issues.apache.org/jira/browse/COUCHDB-2108

Georift:
## Overview

Attempt at grouping database that share the same 'prefix/' as described in #869. I don't think this is quite done I have a few more ideas but I would appreciate some feedback before I go much further.

Some points:
 - Should `prefix/again/db` be contained within `prefix/`'s fold? Right now all folds exist at the top level like so: _(also notice that databases with only one of that prefix are not folded)_
folding-single-level
- There is quite a bit of space in the prefix name row. I was thinking we could put a count of folded database.
 - At the moment it defaults to all folded. Would it be worth adding a default fold all on and off user preference?
 - I made use of ramda as I'm comfortable working with it. I noticed afterwards lodash already exists. If you would prefer I can convert my code to that.

## GitHub issue number

Fixes #869

## Checklist

### DIFF
--- a/app/addons/databases/components.js
+++ b/app/addons/databases/components.js
@@ -23,7 +23,7 @@ import FauxtonComponentsReact from "..//fauxton/components";
 import Stores from "./stores";
 import Actions from "./actions";
 
-import * as R from 'ramda';
+import * as _ from 'lodash';
 import {Tooltip, OverlayTrigger} from 'react-bootstrap';
 
 const databasesStore = Stores.databasesStore;
@@ -177,7 +177,7 @@ class DatabaseTable extends React.Component {
     }
 
     /* take all but the last part and add slashes */
-    return `${R.init(idParts).join('/')}/`;
+    return `${_.initial(idParts).join('/')}/`;
   };
 
   /**
@@ -195,10 +195,11 @@ class DatabaseTable extends React.Component {
         ? array : array[0];
     };
 
-    return R.compose(
-      R.map(unwrapSingleItemArrays),
-      R.groupBy(databaseGroupOrId)
-    )(databaseList);
+    /* key an object by the single database ID OR the group prefix */
+    const groupedDatabases = _.groupBy(databaseList, databaseGroupOrId);
+
+    /* remove the array from unprefixed databases OR single group db groups */
+    return _.mapValues(groupedDatabases, unwrapSingleItemArrays);
   }
 
   render() {
--- a/app/addons/databases/components.js
+++ b/app/addons/databases/components.js
@@ -23,6 +23,7 @@ import FauxtonComponentsReact from "..//fauxton/components";
 import Stores from "./stores";
 import Actions from "./actions";
 
+import * as R from 'ramda';
 import {Tooltip, OverlayTrigger} from 'react-bootstrap';
 
 const databasesStore = Stores.databasesStore;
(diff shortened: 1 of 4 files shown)


app/addons/databases/components.js
```diff
@@ -78,14 +79,74 @@ class DatabaseTable extends React.Component {
...
+            showPartitionedColumn={this.props.showPartitionedColumn}
+          />
+        );
     });
   };
 
+  createGroup = (items, nestLevel) => {
+    /* cannot create an empty group */
+    if (items.length === 0) {
+      /* @TODO add some error logging */
```
Georift: Is there any error logging, besides `console.error()`, we should do here?

app/addons/databases/components.js
```diff
@@ -78,14 +79,74 @@ class DatabaseTable extends React.Component {
...
+
+    const arrowIconClass = isOpen ? 'fonticon-up-dir' : 'fonticon-down-dir';
+
+    return [
+      <tr
+        key={`${groupName}-group-title`}
+        className="database-group-title"
+        onClick={() => this.toggleGroupVisibility(groupName)}
+      >
+        <td colSpan="4">
```
Georift: I copied this from the database load error row, but I think it could be an issue with many`getExtensionColumns()`.

app/addons/databases/components.js
```diff
@@ -179,12 +283,17 @@ class DatabaseRow extends React.Component {
         </tr>
       );
     }
+
     const partitionedCol = this.props.showPartitionedColumn ?
       (<td>{isPartitioned ? 'Yes' : 'No'}</td>) :
       null;
+
+    const DEFAULT_PADDING = 8;
```
Georift: I felt a little bad copying this from the css, any thoughts on a nicer way to handle the indenting?

popojargo:
Yeah. The best solution would be group paging but as you said, it adds a lot of complexity.

Georift:
Perhaps we could instead just push the sorting of the group's to the very end? Then at least this pagination issue is lessened?

popojargo:
I don't think that would be a good idea. At the moment, databases are listed/ordered by their names. If a database is grouped, you will need to go at the last pages of pagination to find the database. It's not very user friendly as you don't know which databases are prefixed/grouped.

With the new _dbs_info, we could easily do server paging. 

I see the following workflow for this feature:

1. List all database names
2. Group by prefixes
3. Get the first 10 top groups (render them)
4. (Optional) For the 10 top groups selected, fetch the information for their immediate children. On expand, the information will already be fetched which enhance user experience.
5. On expand, fetch the information of the sub children.  

This is somehow a big feature addon and I can give you some help if required. 

I would like to have the input of other people. What do you think @wohali @garrensmith @Antonio-Maranhao

Antonio-Maranhao:
IMO the two concepts (grouping & pagination) don't work well together. Maybe we should step back a bit and answer why do we even need grouping here - i.e. in which scenario(s) grouping makes easier to browse the db list?

popojargo:
For example,  we have a SAS application. For each client, we have 3 different databases.  So we prefix all of them. When we lookup for the database, there's a lot of entries to filter out and grouping would definitely make it cleaner.

Antonio-Maranhao:
Thanks for the example @popojargo, but if I understood correctly, you could just type the client name in the search box at the top and quickly access the three DBs for any client.
If what you need is to browse the list of prefixes (i.e. client names) and then expand only the one(s) you're interested on, then I agree that grouping would be helpful. Alternatively just providing a prefix filter would do the trick and works nicely with pagination (something you suggested in #1086)

In any case... Grouping could work with the workflow you proposed given 1) pagination is updated so a group only counts as one entry and 2) fetching the children is done in batches (e.g. with a 'load more' link). That would be a safeguard against firing 100s or 1000s of requests for db metadata.

Georift:
@Antonio-Maranhao The way you described is exactly how I'd imagine it'd need to be implemented. One place I'm not entirely clear is in (1), how would we remove the prefixed databases from the count cleanly without having to iterate it to determine if it should be part of a group.

Antonio-Maranhao:
@Georift I don't think there's a way around it. You'd need to iterate through all db names and group them until you end up with 20 entries (group or single items). The good thing is `_all_dbs` already returns all names sorted, so the only extra work is to group them.

That said, it'd probably make things easier to move the grouping logic to `actions.js` since it already handles the pagination aspect, which btw will need to be updated since jumping to another page is not as simple as `(page_number * items_per_page)`.

Given there's a bit of complexity with mixing grouping & pagination... I'd be ok if we split them into separate view modes. I.e. keep the paginated approach as is, and introduce a new one that would show the complete list (i.e. no pagination) with the groupings.

Georift:
Ah ok, I didn't realise we got a list of all dbs in a single request. I think I'm across how you're thinking.

One thing I'm not entirely clear on is how we will activate the new view? From your comment on changing the pagination logic I'd imagine it involves click on the folded prefix row which would swap for the grouped listing view. Does this sound right?

Antonio-Maranhao:
@Georift Sorry for not being clear. I was suggesting having two separate modes - similar to how you can switch between table/metadata/json in the documents list. The default mode would be the paginated list with no groupings we have today. The new mode would show the complete list (no pagination) with dbs grouped.
Something like a 'Show All / Paginate' link at the top would toggle between the two modes.

But again, this is an alternative approach to avoid making the pagination logic too complex. Feel free to completely ignore it :)

kripper:
Any chance to finish this pull request?

wohali:
@Georift We'd love to have this PR, but given how stale this is, I'm going to close this for now. Please re-open if you find the time to address the remaining requests above.

PR #1148, opened 2018-10-24
[partitioned-dbs] Updates to sidebar

Antonio-Maranhao:
## Overview

Forth in a series of PRs that will be submitted in support of the new user-defined partitioned databases feature (apache/couchdb#1605).

Summary:

Update links in the sidebar to include the selected partition key

## Testing recommendations

**Note:** disregard the error `'partition' parameter is mandatory for queries to this view`. It will be addressed in a separate PR

Execute the steps below in a partitioned db:
- Create a new view `p1` in a new ddoc called `ddocPartitioned`
- Create a new view `g1` in a new ddoc called `ddocGlobal`
  - edit `ddocGlobal` and add `"options": { "partitioned": false }`
  - verify that `ddocPartitioned` and `ddocGlobal` have different icons on the sidebar
- Go back to 'All Documents' and select a partition key at the top. Press Enter to confirm.
- Click on each of the items below and verify 1) page opens correctly and 2) selected partition key is preserved.
  - All Documents
  - Create Document, then cancel to go back to All Documents
  - Run A Query with Mango, then go back to All Documents
  - Design Documents, then back to All Documents
  - `ddocGlobal` > `Metadata`
  - `ddocGlobal` > `g1`
  - `ddocPartitioned` > `Metadata`
  - `ddocPartitioned` > `p1` (it'll show an error)
  - (+) button besides 'Design Documents'
    - New Doc, then cancel
    - New View
    - New Mango Indexes
  - (+) button besides `ddocGlobal`, then click New View
  - (+) button besides `ddocPartitioned`, then click New View

## Related Pull Requests

apache/couchdb#1605

## Checklist


garrensmith:
Two issues while testing:
* I struggle to clear the partition. If I type in foo in the top header and then clear it and press enter it stays there.
* If I create a partition in the top header and then click on changes and then back the partition is cleared.

app/addons/documents/base.js
```diff
@@ -42,14 +42,18 @@ function getQueryParam (query) {
   return query;
 }
 
+function partitionUrlComponent(partitionKey) {
+  return partitionKey ? '/_partition/' + partitionKey : '';
```
garrensmith: Can you rather do `return partitionKey ? "/_partition/${partitionKey}" : '';`
I've had to add in the wrong " at the top because of markdown.

app/addons/documents/base.js
```diff
@@ -42,14 +42,18 @@ function getQueryParam (query) {
...
+}
+
 FauxtonAPI.registerUrls('allDocs', {
   server: function (id, query) {
     /** XXX DEPRECATED: use allDocsSanitized **/
     return Helpers.getServerUrl('/' + id + '/_all_docs' + getQueryParam(query));
   },
-  app: function (id, query) {
+  app: function (id, partitionKey, query) {
     /** XXX DEPRECATED: use allDocsSanitized **/
```
garrensmith: Are these truly deprecated. If so we should remove them and use the correct ones. Or remove the comment.

app/addons/documents/helpers.js
```diff
@@ -21,36 +21,43 @@ const getSeqNum = (val) => {
...
   const addLinks = FauxtonAPI.getExtensions('sidebar:links');
   const newUrlPrefix = '#' + FauxtonAPI.urls('databaseBaseURL', 'app', FauxtonAPI.url.encode(databaseName));
+  let partitionKeyComponent = '';
+  let partitionKeyQueryParam = '';
+  if (partitionKey) {
+    partitionKeyComponent = '/_partition/' + encodeURIComponent(partitionKey);
+    partitionKeyQueryParam = '?partitionKey=' + encodeURIComponent(partitionKey);
+  }
 
   const addNewLinks = _.reduce(addLinks, function (menuLinks, link) {
```
garrensmith: You can do `addLinks.reduce`

app/addons/documents/index-editor/actions.js
```diff
@@ -127,7 +127,7 @@ function deleteView (options) {
 
     // if the user was on the index that was just deleted, redirect them back to all docs
     if (options.isOnIndex) {
-      var url = FauxtonAPI.urls('allDocs', 'app', options.database.safeID(), '?limit=' + FauxtonAPI.constants.DATABASES.DOCUMENT_LIMIT);
```
garrensmith: You have removed the limit here. Are we adding it in elsewhere? Or don't we need it?

Antonio-Maranhao: It's not needed because the query is limited by the num of docs per page.

app/addons/documents/base.js
Antonio-Maranhao: It's still used in several places so I'm removing the comment

Antonio-Maranhao:
@garrensmith I've made the changes. The Selector will now change to global mode if you clear the key and press Enter.

garrensmith:
@Antonio-Maranhao If I add a partition field to the top. And don't hit enter but rather click. The partition field isn't added. Same for if there is a partition at the top and I clear it and then click away from the text box. Then the partition field is not removed.

PR #1151, opened 2018-11-02
Update documents/index-editor to use Redux

Antonio-Maranhao:
## Overview

Update documents/index-editor to use redux.

## Testing recommendations

- Create a new view
- Update a view
   - without Reduce
   - with a built-in Reduce
   - with a CUSTOM Reduce
- Update a view, change its name and the Design Doc
   - Verify the old view is deleted and the new one is created in the target ddoc

## Checklist


app/addons/documents/index-editor/reducers.js
```diff
@@ -0,0 +1,196 @@
...
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+
+import ActionTypes from './actiontypes';
+import Resources from "../resources";
+
+const defaultMap = 'function (doc) {\n  emit(doc._id, 1);\n}';
+const defaultReduce = 'function (keys, values, rereduce) {\n  if (rereduce) {\n    return sum(values);\n  } else {\n    return values.length;\n  }\n}';
+const builtInReducers = ['_sum', '_count', '_stats'];
```
garrensmith: We have a new built in reduce `_approx_count_distinct` http://docs.couchdb.org/en/stable/ddocs/ddocs.html?highlight=reduce#built-in-reduce-functions

Antonio-Maranhao: I'll do this in a separate PR because it should only allow this option when the server version is >= 2.2.0.

PR #1155, opened 2018-11-09
[partitioned dbs] Support partitioned mango queries and creating partitioned indexes

Antonio-Maranhao:
## Overview

Sixth in a series of PRs that will be submitted in support of the new user-defined partitioned databases feature (apache/couchdb#1605).

Summary:

 - Show results for partitioned mango queries
 - Create partitioned indexes

## Testing recommendations

On a partitioned database: 
- Go to **Run A Query with Mango**
  - Verify the Partition Selector is displayed at the top
- Click on **Run Query**
  - Verify all docs that match are returned
- Set a partition key
  - Verify the results are updated with only those that match the key
- Click on **Explain**
  - Verify the results include the selected partition under `"opts"`

- Click on **manage indexes**
  - Verify the **partitioned** checkbox is displayed and it is checked
- Click on **Create Index**
  - Verify the results include the new index, and the index code includes `"design_opts": { "partitioned": true }`
- Uncheck the **partitioned** checkbox and click on **Create Index**
  - Verify the results include the new index, and the index code includes `"design_opts": { "partitioned": false }`
- Add `partitioned: true` to the index code in the editor, make sure the checkbox is unchecked, then click on **Create Index** 
  - Verify the results include the new index, and the index code includes `"design_opts": { "partitioned": true }` - which means the value in the editor takes precedence over the checkbox.

## Related Pull Requests

apache/couchdb#1605

## Checklist

### DIFF
--- a/app/addons/databases/reducers.js
+++ b/app/addons/databases/reducers.js
@@ -13,7 +13,7 @@
 import ActionTypes from './actiontypes';
 const initialState = {
   partitionedDatabasesAvailable: false,
-  isLoadingDbInfo: false,
+  isLoadingDbInfo: true,
   dbInfo: undefined,
   isDbPartitioned: false
 };
--- a/app/addons/documents/base.js
+++ b/app/addons/documents/base.js
@@ -222,40 +222,40 @@ FauxtonAPI.registerUrls('mango', {
     return Helpers.getServerUrl('/' + db + '/_index' + query);
   },
 
-  'index-apiurl': function (db, query) {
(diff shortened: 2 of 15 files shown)


garrensmith:
When we send a mango query we end up setting the r value (Number of reads from a replica) before returning the result. And if I look at the _explain for some reason its set to 49
```
"r": [
   49
  ],
```

Once we add in the partition restrictions this is going to cause an issue. r must always be equal to 1 fro partitions.

PR #1171, opened 2019-01-09
New replace backbone model

Antonio-Maranhao:
Reopening because Travis build is stuck in PR #1165 

## Overview

Introduces new classes `Model` and `Collection` to directly replace Backbone's `Model` and `Collection`. The new classes implement the main Backbone's methods (e.g. `save()` and  `destroy()`) using `fetch`. This provides a quick(er) way to remove the dependency with Backbone since most of the Fauxton code/logic that relies on Backbone models and collections can remain unchanged.

Summary of changes:
 - replaced all references to Backbone.Model/Collection with Model/Collection
 - it now uses `class extends Model` / `class extends Collection` instead of `Backbone.Model.extend` / `Backbone.Collection.extend`
 - callers now use `then/catch` (Promises) instead of `done/fail` (jQuery deferred).

 - Misc:
   - Replaced `Fauxton.VersionInfo` with a `fetch` call.
   - Removed unused classes: `BulkDeleteDocModel`, `BulkDeleteDocCollection`, `MangoBulkDeleteDocCollection`

## Testing recommendations

- Verify you can create, list, delete databases
- Verify you can list all docs, list view results
- Edit and save docs
- Edit and save views

## Checklist


wohali:
@Antonio-Maranhao reminder, you can actually log into travis-ci.org and force a build to be killed/restarted. no need to close one PR and open a new one.

Antonio-Maranhao:
thanks @wohali. I tried that several times but the job just wouldn't start running.

willholley:
@Antonio-Maranhao is this still relevant? Thinking we should merge or close the outstanding PRs before CouchDB 3.

PR #1186, opened 2019-02-13
Don't change the browser default for Ctrl + L

MasseGuillaume:
Ctrl + L or Cmd + L is the shortcut to jump to the address bar. Codemirror shows a jump to line prompt. I personally prefer to jump to the address bar.

## Checklist

### DIFF
--- a/app/addons/components/components/codeeditor.js
+++ b/app/addons/components/components/codeeditor.js
@@ -45,6 +45,11 @@ export class CodeEditor extends React.Component {
 
     // optional editor key commands (e.g. specific save action)
     editorCommands: [],
+    
+    extraKeys: {
+      "Cmd+L" -> null,
+      "Ctrl+L" -> null
+    },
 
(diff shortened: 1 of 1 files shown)


wohali:
@MasseGuillaume What does Ctrl-L bind to without your PR?

Would it make sense to bind that ctrl-command to a different keybinding, instead of completely unmapping it?

MasseGuillaume:
@wohali it binds to a jump to line javascript prompt() window. Maybe we could re-map to `ctrl/cmd+shift+L` ?

wohali:
@MasseGuillaume That sounds like a good idea!

wohali:
@MasseGuillaume Also, you are failing a test on invalid syntax in that file:

https://travis-ci.org/apache/couchdb-fauxton/builds/492929671#L1165-L1176

MasseGuillaume:
Here you go: https://github.com/apache/couchdb-fauxton/pull/1187

PR #1198, opened 2019-05-05
Improve Accessibility for screen reader users

### LINKED ISSUE #1197: Improve accessibility for screen reader users
## Expected Behavior

Screen reader users should have full access to the interface.

## Current Behavior

Too many to list, but generally, there are missing labels, and use of `<div>` or `<span>` instead of a more semantically appropriate element. 

## Possible Solution

See above

## Steps to Reproduce (for bugs)

Encountered while following the tutorials:

1. When minimized, navigation links have no labels, so are just read as "link" for each item. The control that toggles minimize/maximize has no role and label.
2. Incoming notifications are not read automatically.

## Your Environment

Latest Fauxton on Apache CouchDB v. 2.3.1
Browser: Firefox 67
Other software used: NVDA screen reader v2019.2

Neurrone:
Fixes #1197.

## Overview

Improve accessibility by adding missing labels and using appropriate roles.

* [x] Navigation menu: ensure that nav links are always labelled for screen readers, regardless of whether they are visually minimized. Add labels and roles for the minimize toggle button.
* [x] Make screen readers read incoming notifications, add some missing labels in the notifications panel
* [x] render notifications in a labelled `<aside>` instead of a `<span>`
* [x] Narrow the scope of `main` so it excludes the navigation sidebar
* [x] `DatabaseRow`: Add missing labels to controls

## Testing recommendations

Run NVDA with Firefox and connect to a fauxton instance.

## Checklist

- [x] Documentation reflects the changes;

### DIFF
--- a/app/addons/components/components/copy.js
+++ b/app/addons/components/components/copy.js
@@ -49,7 +49,7 @@ export class Copy extends React.Component {
 
   getClipboardElement () {
     if (this.props.displayType === 'icon') {
-      return (<i className="fontawesome icon-paste"></i>);
+      return (<i aria-hidden="true" className="fontawesome icon-paste"></i>);
     }
     return this.props.textDisplay;
   }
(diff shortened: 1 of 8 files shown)


Antonio-Maranhao:
Thanks for the contribution @Neurrone

Issue #1205, opened 2019-07-09
Consider adding an .editorconfig to enforce aspects of the style guide


iamgollum:
When I was working on issue #1202, I noticed in my PR that the format was off. I went to check the repo and noticed there was no `.editorconfig`. This can help standardize styling across os / contributors.

https://github.com/apache/couchdb-fauxton/blob/master/styleguide.md

popojargo:
I would be +1 on this. We already have eslint for code formatting but a `.editorconfig` would also helps being consistent in the different files.

PR #1206, opened 2019-07-09
Added base editor config based on styleguide - #1205

iamgollum:
Fulfills #1205


iamgollum:
Should I also add the following?

`tab_width`
`end_of_line`
`charset`
`trim_trailing_whitespace`
`insert_final_newline`
`max_line_length`

popojargo:
- charset = utf-8
- trim_trailing_whitespace = true
- insert_final_new_line = true
- end_of_line = lf

As for the max_line_length, we suggest line no longer than 120 but its required in some cases...

iamgollum:
@popojargo I am on it

iamgollum:
@popojargo not sure if I should be force pushing? Are commits squashed on merge?

garrensmith:
This looks cool. What is the value of having this versus setting up prettier that would just run at each commit?

popojargo:
Having prettier running as a commit hook would be a better solution indeed.

popojargo:
Also, no need to squash on your branch, you can do it at the merge time. Altough, you should always rebase your branch before merging

iamgollum:
@popojargo @garrensmith Ill add in prettier. I usually use `husky` for commit hooks. Can I proceed with prettier?

iamgollum:
I figured out how to combine linting with everything else, so I installed `lint-staged` and applied a regex similar to how you guys did it for the `stylecheck` script

iamgollum:
Did a little experiment, `pretty-quick` will only trigger once a file has already been previously formatted in prettier. So first time files - unless applying the VSCode prettier extension - will need to be formatted manually. After that, it will work.

iamgollum:
@popojargo @garrensmith Done with prettier setup. Let me know what you guys think.

.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
```
popojargo: Do we still need that file with prettier?

package.json
```diff
@@ -15,16 +15,34 @@
...
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.{js,jsx}": [
+      "pretty-quick --staged",
+      "eslint . --fix",
+      "git add"
```
popojargo: Why the `git add` here?

package.json
```diff
@@ -15,16 +15,34 @@
       "react"
     ]
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.{js,jsx}": [
```
popojargo: We don't want to lint JSON files or any other files?

iamgollum: @popojargo thats up to you guys ... if there is a lot of JSON, sure I can add that in.

.editorconfig
iamgollum: @popojargo the point of .editorconfig is to specify a common editor configuration for all developers working on the same project. EditorConfig is supported by many editors and is thus widely used. You don't have to duplicate overlapping settings between the two since editorconfig will work jointly with prettier.

The editor can use prettier to reformat a file at the point configured (aka. for this project on staged), but I don't think that prettier can be used by the editor each time it is opening a new line, so the editor cannot rely entirely on prettier for formatting thus the use of .editorconfig. Prettier is a plugin while the config is more "native".

The .editorconfig is also required for GitHub code views to render your tabs (not spaces) with any desired unit of spaces instead of the 8 that is default, so even a code base that is not shared between multiple developers may still need to define this file to look "prettier" on GitHub.

package.json
iamgollum: @popojargo Excellent question. Its because if eslint produces any fixes it will pipe in those files through `git add` automatically.

popojargo:
@iamgollum Can you rebase on master?

iamgollum:
@popojargo I will attempt to do that now

iamgollum:
@popojargo applied merge. Hopefully this is good to go ...

popojargo:
@iamgollum You need to update the package-lock.json as well. Running `npm i` should solve that

popojargo:
@garrensmith @Antonio-Maranhao  Any objections to use prettier?

wohali:
@iamgollum @popojargo This PR seems stale, any chance to revive it?

Issue #998, opened 2017-10-12
UI: Display bug and UI feedback


zachsa:
The Fauxton interface is great visually! Compared to Futon sometimes elements of the UI get in the way, however, and make for a slower workflow and provide less useful visual feedback than Futon. A few examples that I can think of offhand:

1. Notifications cover controls and last too long
2. When a notification tells you something useful (like an error) it disappears too quickly
3. On the all-documents page there is no total documents count
4. MapReduce view displays aren't useful
5. Errors in the UI

## Notifications
The green notification screen takes forever to disappear, meaning that I usually have to drag my mouse across the top screen to close it. However, when it shows an error that I want to read there are two problems:

1. It truncates the error message
2. You can't make the message hang around longer

I recommend that OK (green and blue) notification messages are transparent and disappear on a click, and that error messages (the black with red writing) don't disappear when hovered over. Or that something else is used instead. The green notifications are irritating and the error notifications are exasperating for now.

## DB metrics
Where I worked previously, we were using Couchbase-lite on an Android phone and the most basic test in our development workflow was 'is it syncing'. It was great that we could refresh the Futon page and see if the doc count incremented. On Fauxton, so far as I know, you would have to use a cURL command or navigate away from the Fauxton interface to the _all_docs endpoint. I'd prefer to stay on Fauxton if possible

I recommend putting this metric next to 'showing document 1 - 20'.

## MapReduce index results view
Futon showed a key: condensedJsonDoc output. this was much better than the current options in Fauxton:

1. The JSON view shows the first two documents of the index. And doesn't display the key. 
2. The Metadata view shows data in a MUCH less desnse way than Futon. Also. I seem to have two ID fields. (_id and id). the view (just a test DB) emits doc._id as a key. Using GUIDs as IDs, I don't know that I would ever need to see which document ID produced which map index.
3. On Futon it was nice to be able to edit the view function whilst seeing the result of the previously run view function below the editing. This is not possible on Fauxton.
4. The Table view includes the doc, which I can't imagine doing anything other than truncating after the first JSON property. So it's not useful to look at.

## A bug
This doesn't really effect usability, but is clearly an error. There is a display bug in the {json} button in the menu bar. I've attached a photo showing the error:

![json](https://user-images.githubusercontent.com/6336500/31486234-0f148686-af37-11e7-8833-3ffa8a6c1938.PNG)

I'm using a recent version of chrome on a relatively old 21" VGA monitor.

[...]

PR #1211, opened 2019-07-17
Fixed z-index btn group hover/focus/active relative to reselect dropdown

iamgollum:
There is a general misunderstanding of the z-index [Stacking Context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context)

```html
<div class="Select ... etc...   is-open  ...."> // Relative Position
   ... etc....
   <div Select-outer-menu /> ... dynamic
</div>
```

The following div will always be above the Select, no matter how high you set the z-index on the outer menu ....

```html
<div class="two-sides-toggle-button btn-group"> // Relative Position, Z-Index: 2
   <button type="button" class="btn btn-default"><i class="fonticon-table"></i> Table</button>
  <button type="button" class="active btn btn-default">Metadata</button>
  <button type="button" class="btn btn-default"><i class="fonticon-json"></i> JSON</button>
</div>
```

The fix is to adjust the Z-index on the button effects: hover,focus,active.
You can also make the Select not-relative and redesign that component.

See #998 in the thread with the dropdown on the reselect component going underneath the btn group on effect.

Screen Shot 2019-07-16 at 6 07 22 PM


assets/less/bootstrap/button-groups.less
```diff
@@ -85,7 +84,7 @@
 .btn-group > .btn:focus,
 .btn-group > .btn:active,
 .btn-group > .btn.active {
-  z-index: 2;
+  z-index: 1;
```
Antonio-Maranhao: I think it'd be better to set the `z-index` at https://github.com/apache/couchdb-fauxton/blob/master/assets/less/formstyles.less#L350 instead.
This way it's obvious what we are overriding whenever we upgrade bootstrap.

iamgollum: @Antonio-Maranhao sounds good. I will do this tonight!

iamgollum: @Antonio-Maranhao required a `!important` since bootstrap took hire precedence with more focused scope. Based on specificity: https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity ... its okay practice in this case to use it since Bootstrap is a "3rd party" stylesheet.

assets/less/formstyles.less
```diff
@@ -345,13 +353,12 @@ input.errorHighlight {
       background-color: @hoverHighlight;
       color: #fff;
       .box-shadow(none);
-
     }
     &.active {
+      z-index: 1 !important;
```
popojargo: Maybe it would be relevant to add a comment here to justify the !important

iamgollum: you got it

PR #1213, opened 2019-07-30
New design for Create Database panel

Antonio-Maranhao:
## Overview

Update UX of the Create Database dialog
 - The user has now to explicitly select between 'Partitioned' and 'Not partitioned' (when enabled)
 - Show help text to help user decide when to use a partitioned db.

### Without Partitioned feature
image

### With Partitioned feature
image

## Testing recommendations

- Try to create a new db providing only the name. It should show an error message that either 'Partitioned' and 'Not partitioned' must be selected.

## Checklist

### DIFF
--- a/app/addons/components/components/accordion.js
+++ b/app/addons/components/components/accordion.js
@@ -1,0 +1,74 @@
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License. You may obtain a copy of
+// the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+
+import PropTypes from 'prop-types';
+import React from 'react';
+import { Collapse } from 'react-bootstrap';
+import classnames from 'classnames';
+
+export const Accordion = (props) => {
+  const { children, className, style } = props;
+  const classNames = classnames('faux--accordion', className);
+  return (
+    <ul className={classNames} style={style}>
+      {children}
+    </ul>
+  );
+};
+
+export class AccordionItem extends React.Component {
+  static propTypes = {
+    title: PropTypes.string.isRequired
+  };
+
+  static defaultProps = {
+    onClick: () => { }
+  };
+
+  state = {
+    open: false
+  };
+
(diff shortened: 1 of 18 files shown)


app/addons/components/components/accordion.js
```diff
@@ -0,0 +1,74 @@
...
+
+  static defaultProps = {
+    onClick: () => { }
+  };
+
+  state = {
+    open: false
+  };
+
+  handleClick = (event) => {
```
garrensmith: This is all fine. But the new hooks functionality would also work quite nice here if you wanted to move this to a functional component instead of a class.

i18n.json.default.json
```diff
@@ -21,6 +21,7 @@
     "auth-admin-created": "CouchDB admin created",
     "auth-change-password": "Your password has been updated.",
     "auth-admin-creation-failed-prefix": "Could not create admin.",
-    "auth-passwords-not-matching": "Passwords do not match."
+    "auth-passwords-not-matching": "Passwords do not match.",
+    "create-db-partitioned-help": "A partitioned database requires a partition key for every document, where the document _id format is '&lt;partition_key&gt;:&lt;doc_key&gt;'. A partition is a logical grouping of documents. Use of partition queries can significantly improve response times over global queries."
```
garrensmith: `Use of partition queries can significantly improve response times over global queries.` is a bit confusing to me.
I think we need to reword that.

test/nightwatch_tests/custom-commands/helper.js
```diff
@@ -21,11 +21,13 @@ exports.checkForDocumentCreated = function checkForDocumentCreated (url, timeout
 
 exports.checkForDatabaseCreated = function checkForDatabaseCreated (couchUrl, databaseName, timeout, cb) {
   const timeOutId = setTimeout(() => {
-    throw new Error('timeout waiting for db to appear');
+    throw new Error('Timeout ('+timeout+') waiting for db to appear');
   }, timeout);
 
   const intervalId = setInterval(() => {
+    console.log(`Checking if db ${databaseName} exists on ${couchUrl}`);
```
garrensmith: You can remove these

popojargo:
Why not use a checkbox or a yes/no toggle with a single label? 

Partionned: yes/no 

Also, why having a full section with the partitionning details instead of having a tooltip next to the "Partionning" title?

Antonio-Maranhao:
@garrensmith Updated the help text and removed the `console.log` lines

Antonio-Maranhao:
@popojargo The idea was to make the partitioned/global choice more prominent, based on user feedback we got in Cloudant. Hence why the full section vs the previous single checkbox.

PR #1218, opened 2019-08-28
Applied css fix to align create doc btn text when it becomes two lines

iamgollum:
# Overview

On a certain screen size the create document button text gets cut off when it goes from one line to two lines. This is a fix to center it appropriately. 

# Github Issue

Addresses one of many issues in #998 

# Checklist

- [x] - Code works correct;
- [x] - Changes are covered by tests (styling issue, so not needed in this project)

### DIFF
--- a/app/addons/documents/assets/less/index-results.less
+++ b/app/addons/documents/assets/less/index-results.less
@@ -71,14 +71,9 @@
 }
 
 .document-result-screen__toolbar-create-btn {
+  display: flex;
+  align-items: center;
   height: 42px;
-}
-
-// Fixes text centering when it goes from a one line to two
-@media (max-width: 965px) {
-  .document-result-screen__toolbar-create-btn {
-    line-height: 12px;
-  }
 }
 
 a.document-result-screen__toolbar-create-btn:active,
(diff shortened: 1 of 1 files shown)


app/addons/documents/assets/less/index-results.less
```diff
@@ -76,6 +73,13 @@
   height: 42px;
 }
 
+// Fixes text centering when it goes from a one line to two
+@media (max-width: 965px) {
```
popojargo: I'm not a css expert but wouldn't it be better to have something like this:

`.document-result-screen__toolbar-create-btn {
  height: 42px;
  
  @media (max-width: 965px) {
      line-height: 12px;
  }
}`

popojargo:
With a width of 981, I got the same behavior: 
image

Maybe use the standard 1024 width?

app/addons/documents/assets/less/index-results.less
iamgollum: That makes sense.  I will do that

iamgollum:
@popojargo found a better fix that is responsive without need for media query.

PR #1222, opened 2019-09-05
Remove deprecated size properties from .info()

popojargo:
## Overview

Replace deprecated size properties with the new ones.

## Checklist

### DIFF
--- a/app/addons/databases/stores.js
+++ b/app/addons/databases/stores.js
@@ -113,7 +113,7 @@ const DatabasesStoreConstructor = FauxtonAPI.Store.extend({
       return {};
     }
     const {sizes} = details;
-    const dataSize = sizes.active || details.file || 0;
+    const dataSize = sizes.active || 0;
 
     return {
       dataSize: Helpers.formatSize(dataSize),
--- a/app/addons/databases/stores.js
+++ b/app/addons/databases/stores.js
@@ -112,7 +112,8 @@ const DatabasesStoreConstructor = FauxtonAPI.Store.extend({
     if (!details) {
       return {};
     }
-    const dataSize = details.data_size || details.disk_size || 0;
+    const {sizes} = details;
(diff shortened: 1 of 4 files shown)


app/addons/databases/stores.js
```diff
@@ -112,7 +112,8 @@ const DatabasesStoreConstructor = FauxtonAPI.Store.extend({
     if (!details) {
       return {};
     }
-    const dataSize = details.data_size || details.disk_size || 0;
+    const {sizes} = details;
+    const dataSize = sizes.active || details.file || 0;
```
Antonio-Maranhao: I think `details.file` should be removed here. It doesn't make sense to show different values under the same column.

app/addons/documents/designdocinfo/components/DesignDocInfo.js
```diff
@@ -56,8 +56,9 @@ export default class DesignDocInfo extends React.Component {
       return <LoadLines />;
     }
     const viewIndex = this.props.viewIndex;
-    const actualSize = (viewIndex.data_size) ? viewIndex.data_size.toLocaleString('en') : 0;
-    const dataSize = (viewIndex.disk_size) ? viewIndex.disk_size.toLocaleString('en') : 0;
+    const {sizes} = viewIndex;
+    const actualSize = (sizes.active) ? sizes.active.toLocaleString('en') : 0;
+    const dataSize = (sizes.file) ? sizes.file.toLocaleString('en') : 0;
```
Antonio-Maranhao: The equivalent here is `sizes.external` according to https://github.com/apache/couchdb/pull/2163/files#diff-b8437e2de7ccbb213a215be755cb6935

app/addons/databases/stores.js
popojargo: I kept the same "type" of values as it was before but if we want to only stick to one I can change it.

PR #1241, opened 2019-10-30
Extend fauxton to support IAM for IBM

rols2015:
## Overview
Fauxton can handle basic authentication only.

It has been changed so that it can connect to Cloudant instance with IAM (service call to get token and then use the Bearer authentication)

The process for the user is:

.) Generate API key which (User needs to be enabled for the Cloudant Dashboard).

.) Use the API key for getting a token.

.) Subsequent API calls to cloudant with Bearer authentication.

related article:
https://cloud.ibm.com/docs/services/watson?topic=watson-iam

## Testing recommendations
image

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ test/templates.js
 test/dashboard.assets
 # test coverage dir
 coverage
+.idea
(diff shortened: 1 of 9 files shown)


Antonio-Maranhao:
Hi @rols2015, the Cloudant Dashboard already supports IAM logins so why add this capability to Fauxton? 
Just trying to understand the motivation behind this change, particularly because I don't see any advantages on using Fauxton instead of the Cloudant Dashboard.

garrensmith:
Hi, @rols2015 thanks for being involved. Unfortunately I'm -1 on this feature. Fauxton is purely for CouchDB features. IAM is IBM only. If CouchDB ever adds a more generic authentication system that would work with IBM's IAM we consider something like this.

PR #1245, opened 2020-01-07
Mango: support multiple server-side warnings

willholley:
## Overview

Changes the handling of mango warnings to support multiple warnings
delimited by a newline, as added in https://github.com/apache/couchdb/pull/2410.

This also removes the client-side warning for too many docs scanned as that is now generated on the server.

## Testing recommendations

When https://github.com/apache/couchdb/pull/2410 is merged, run some Mango queries with bad  / no indexes to generate warnings.

## Checklist

### DIFF
--- a/app/addons/documents/mango/components/ExecutionStats.js
+++ b/app/addons/documents/mango/components/ExecutionStats.js
@@ -13,8 +13,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Popover, OverlayTrigger } from 'react-bootstrap';
 
-const TOO_MANY_DOCS_SCANNED_WARNING = "The number of documents examined is high in proportion to the number of results returned. Consider adding an index to improve this.";
-
 export default class ExecutionStats extends React.Component {
   constructor (props) {
     super(props);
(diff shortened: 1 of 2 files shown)


app/addons/documents/mango/components/ExecutionStats.js
```diff
@@ -37,18 +35,10 @@ export default class ExecutionStats extends React.Component {
...
-
-    // warn if many documents scanned in relation to results returned
-    if (!warning && executionStats.results_returned) {
-      const docsExamined = executionStats.total_docs_examined || executionStats.total_quorum_docs_examined;
-      if (docsExamined / executionStats.results_returned > 10) {
-        return TOO_MANY_DOCS_SCANNED_WARNING;
-      }
+  getWarning(warning) {
+    if (warning) {
+      return warning.split('\n').join('<br>');
```
Antonio-Maranhao: This will print `<br>` instead of adding a line break. You can use this instead:
```
getWarning(warnings) {
    if (warnings) {
      return (
        <span>
          {warnings.split('\n').map(warnText => <>{warnText}<br/></>)}
        </span>
      );
    }
  }
```

willholley: thanks @Antonio-Maranhao - updated with those changes

willholley:
hmm - looks like CI against `couchdb:dev` is broken...

PR #1246, opened 2020-01-14
Nav: Fix potential ID selector collisions

mhamann:
## Overview

If a design doc is created with a name that conflicts with an existing HTML element's ID, it will receive the styling for that element in addition to causing invalid HTML.

In my case, I'd created a design doc named `dashboard`, which then produced a navigation item with an ID of `#dashboard`. This conflicts with the existing `#dashboard` element, which has several CSS styles applied. It then breaks the navigation layout (see screenshot). It becomes impossible to edit the views within the design doc without manually editing the UI.

image

This change prefixes those IDs in order to ensure they
are unique.

## Testing recommendations

Tested the UI locally to ensure the rendering was fixed and that I could still access my docs, views, etc.

image

## GitHub issue number

N/A

## Related Pull Requests

N/A

## Checklist

### DIFF
--- a/app/addons/documents/sidebar/components/DesignDoc.js
+++ b/app/addons/documents/sidebar/components/DesignDoc.js
@@ -157,7 +157,7 @@ export default class DesignDoc extends React.Component {
           </div>
         </div>
         <Collapse in={this.props.isExpanded}>
-          <ul className={toggleBodyClassNames} id={this.props.designDocName}>
+          <ul className={toggleBodyClassNames} id={`design-doc-menu-${this.props.designDocName}`}>
             <li className={metadataRowClass}>
               <a href={"#/" + designDocMetaUrl} className="toggle-view accordion-header">
                 Metadata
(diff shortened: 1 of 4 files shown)


willholley:
nice catch - thanks @mhamann.

+1

willholley:
something's not quite right in the CI here - it's consistently failing when search is enabled. I'm running the Jenkins build against master to verify that it's not a general issue.

mhamann:
@willholley I had some trouble running the nightwatch tests locally, so I had to update a couple of test cases based on failures I noticed in Travis. It's possible I may still have missed something...

willholley:
@mhamann https://travis-ci.org/apache/couchdb-fauxton/jobs/637256303 (test suite running against couchdb + search) fails consistently with this branch but passes on master (after some retries). It seems likely this is just general flakiness; @Antonio-Maranhao do you have any suggestions?

willholley:
@mhamann the nightwatch tests [for search indexes](https://github.com/apache/couchdb-fauxton/tree/master/app/addons/search/tests/nightwatch). Can you update your PR to patch them similarly to the view tests and then we can merge?

mhamann:
@willholley got a clean build after my last commit.

willholley:
@mhamann perfect -thanks!

PR #1267, opened 2020-04-25
Add "None" analyzer option for the Search View editor

popojargo:
Prevents from adding an analyzer if none were defined in the design document

## GitHub issue number

#1256

## Checklist

### DIFF
--- a/app/addons/search/components/Analyzer.js
+++ b/app/addons/search/components/Analyzer.js
@@ -34,7 +34,7 @@ export default class Analyzer extends React.Component {
   };
 
   validate = () => {
-    if (this.props.analyzerType === Constants.ANALYZER_SINGLE) {
+    if (this.props.analyzerType !== Constants.ANALYZER_MULTIPLE) {
       return true;
     }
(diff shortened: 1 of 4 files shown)


garrensmith:
I'm not very familiar with search, but I've just spoken to @rnewson and I don't think this makes sense. An analyzer will always have to used to create an index. So if someone selects none in Fauxton then `standard` would actually be used by Lucene. So I think its better to rather just have the default analyzer set in the beginning and then a user can choose a different one if they know what they are doing.

Antonio-Maranhao:
@popojargo based on Garren's explanation, having the `None` option could be misleading, so I'm -1 on this change.

In #1256 it expects `It should not add any analyzer if not selected (furthermore a wrong analyzer)`. That behaviour (setting `single` instead of `standard`) was fixed in https://github.com/apache/couchdb-fauxton/pull/1261

popojargo:
@Antonio-Maranhao Gotcha. I will advise the issuer and close that issue!

PR #1284, opened 2020-05-22
[DISCUSS] feat: prototype news page

janl:
Hey all,

I thought it would be neat to be able to show the blog inline with fauxton, that way folks can see project updates in a more timely manner.

What do you think?

Screenshot 2020-05-22 at 16 31 42

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ app/addons/*
 !app/addons/cors
 !app/addons/setup
 !app/addons/search
+!app/addons/news
 settings.json*
 i18n.json
 !settings.json.default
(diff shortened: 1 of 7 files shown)


ermouth:
What about security? Loading remote iframe does expose at least IP of a particular CouchDB instance.

big-r81:
Can you explain why this might be a problem? Every web-server expose an ip address...

ermouth:
@big-r81 You don’t expect your admin panel to go outside and log the presence of your CouchDB instance without any notice, just because you clicked a button at nav panel. New XSS vulnerabilities are discovered in browsers  too regularly, and may I say much, much more frequently if compared to critical security vulnerabilities in CouchDB.

Also, as for me, the idea of creating a news outlet inside any admin panel might be good if there exist a decent, regularly updated and interesting news feed, which is absolutely not the case.

PR #1288, opened 2020-06-09
Feature/replication paginate

popojargo:
## Overview

Add support for pagination in Replicatory activity panel.

## Related Pull Requests

Supersed  #1075 

## Checklist


popojargo:
@garrensmith @Antonio-Maranhao Do I have to make some magic for partitioned database?

app/addons/replication/reducers.js
```diff
@@ -348,6 +371,38 @@ const replication = (state = initialState, {type, options}) => {
...
+          docsPerPage: options,
+          page: 1,
+          canShowNext: false
+        }
+      };
+
+    case ActionTypes.REPLICATION_NEXT_PAGE:
+      return {
+        ...state,
+        pagination: {
```
Antonio-Maranhao, line 391: Shouldn't you add `...state.pagination` to preserve the other values?

app/addons/replication/reducers.js
```diff
@@ -348,6 +371,38 @@ const replication = (state = initialState, {type, options}) => {
...
+          docsPerPage: state.pagination.docsPerPage,
+          page: state.pagination.page + 1
+        }
+      };
+
+    case ActionTypes.REPLICATION_PREVIOUS_PAGE:
+      return {
+        ...state,
+        pagination: {
```
Antonio-Maranhao, line 400: Same here re: `...state.pagination`

app/addons/replication/components/common-table.js
```diff
@@ -381,7 +381,7 @@ export class ReplicationTable extends React.Component {
   render () {
 
     return (
-      <Table striped>
+      <Table striped style={{marginBottom: '100px'}}>
         <thead>
           <tr>
             <th className="replication__table-bulk-select">
```
Antonio-Maranhao, line 387: Unrelated to your changes... but I noticed on lines 403 and 405, it should be `startTime` and not `statusTime` so the sorting matches the values displayed for that column.

brianewilkins:
In inspected the commit https://github.com/apache/couchdb-fauxton/pull/1288/commits/a35a657a75d2e046b6699da09be0044be3170e68
and, insofar as I understand it, it all looks good to me.
I do see that the commit contains fixes for the specific faults pointed out by @Antonio-Maranhao at https://github.com/apache/couchdb-fauxton/pull/1288#pullrequestreview-429124811 

I have played with it a bit (I created 110 continuous replications to do so) and did not find anything wrong with it.

So +1 from me, if that counts for anything.

Antonio-Maranhao:
@popojargo sorry for the delay. I'll make time to review this today or tomorrow.

app/addons/replication/reducers.js
```diff
@@ -406,4 +463,10 @@ export const getReplicateInfo = (state) => {
   });
 };
 
+export const getPagination = (state) => state.pagination;
+export const getPageStart = (state) => 1 + (state.pagination.page - 1) * state.pagination.docsPerPage;
+export const getPageEnd = (state) => state.pagination.docsPerPage + (state.pagination.page - 1) * state.pagination.docsPerPage;
```
Antonio-Maranhao, line 468: Can you use this so it shows the correct range on the last page?

```suggestion
export const getPageEnd = (state) => {
  if (state.isLoading || !state.statusDocs || state.statusDocs.length === 0) {
    return false;
  }
  const pageStart = (state.pagination.page - 1) * state.pagination.docsPerPage;
  return pageStart + state.statusDocs.length;
};
```

Antonio-Maranhao (changes requested):
Besides my suggestion, just wanted to point out that column sorting and filtering only affect the docs in the current page, which I'm sure it'll mislead users.

Maybe it'd be worth adding a note at the top @popojargo @brianewilkins ?

popojargo:
> Besides my suggestion, just wanted to point out that column sorting and filtering only affect the docs in the current page, which I'm sure it'll mislead users.
> 
> Maybe it'd be worth adding a note at the top @popojargo @brianewilkins ?

Brainstorming here:

- We could maybe add a tooltip to the sort buttons? That said, that would be a bit repetitive.<
- Another variant of the tooltip could be to add a tooltip to the right of the action bar. We would only have one tooltip for all sort buttons.
- We could add a note above the table. That would take some height some something that might not be important?

Antonio-Maranhao:
@popojargo I'd go with a note above the table. The tooltips are probably too repetitive as you said and less obvious.

PR #1317, opened 2021-06-21
Add extension point for database size

rhyshort:
## Overview

Add extension point for database size called `DatabaseList:dbSizeReportFieldName`,
this enables consumers to override the displayed value to meet
their deployment needs. For example, in the case of Cloudant's usage,
the default `sizes.active` field is meaningless to their customers
since they bill using the value from `sizes.external`, using the
extension it is possible to customize the view to provide a better experiance.

The extension point `DatabaseList:dbSizeReportFieldName` only supports
the single extentsion, and assumes the field name is in the `sizes`
object returned from db_info.

## Testing recommendations

Regular CI tests passing. It can be further tested by providing an extension
to use one of the other fields in the `sizes` object from the `{db}/` endpoint.

## GitHub issue number

n/a

## Related Pull Requests

n/a

## Checklist

### DIFF
--- a/app/addons/databases/stores.js
+++ b/app/addons/databases/stores.js
@@ -112,8 +112,15 @@ const DatabasesStoreConstructor = FauxtonAPI.Store.extend({
     if (!details) {
       return {};
     }
+
+    const sizeReporter = FauxtonAPI.getExtensions('DatabaseList:dbSizeReportFieldName');
+    let fieldName = 'active';
+    if (sizeReporter.length > 0) {
+      fieldName = sizeReporter[0];
+    }
+
     const {sizes} = details;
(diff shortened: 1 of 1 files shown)


app/addons/databases/stores.js
```diff
@@ -112,8 +112,15 @@ const DatabasesStoreConstructor = FauxtonAPI.Store.extend({
     if (!details) {
       return {};
     }
+
+    const sizeReporter = FauxtonAPI.getExtensions('DatabaseList:dbSizeReporter');
```
garrensmith, line 116: This is a real nitpick but I think it should be `dbSizeReportFieldName` or `dbSizeReporterName`

garrensmith (approved):

PR #1331, opened 2022-01-19
Replace brace with ace-build and react-ace

rhyshort:
## Overview

To enable users to remove the CSP directive `style-src unsafe-inline`, the
ace editor version needs to be upgraded to at least 1.4.13 which
includes the global option `useStrictCSP`. In order to do this
brace has had to be replaced since it hasn't had an update in 4 years.
react-ace is chosen as a replacement as it uses ace-builds and wraps
ace in a react component. The option `useStrictCSP` is enabled,
with the required style sheets imported, where possible properties
that can be handled by react-ace are passed directly to `AceEditor`
as part of the `render()` call, otherwise customizations are left
to be handled by the codeeditor component.

## Testing recommendations

The changes should be covered by the existing nightwatch tests and from an end user perspective the changes should be not be noticable.

## GitHub issue number

n/a
## Related Pull Requests

n/a

## Checklist


Antonio-Maranhao (changes requested):

app/addons/components/components/zenmodeoverlay.js
```diff
@@ -15,7 +15,7 @@ import app from "../../../app";
 import {CodeEditor} from './codeeditor';
 import {Tooltip, OverlayTrigger} from 'react-bootstrap';
 
-require('brace/theme/dawn');
+require('ace-builds/src-min-noconflict/theme-dawn');
```
Antonio-Maranhao, line 18: Dawn theme was not getting applied. I had to add `import 'ace-builds/css/theme/dawn.css';` here to make it work.

Antonio-Maranhao (changes requested):
Found a couple of issues.

app/addons/components/components/zenmodeoverlay.js
rhyshort: done

app/addons/documents/rev-browser/components/splitscreenarea.js
```diff
@@ -12,16 +12,13 @@
...
-import ace from "brace";
 
-require('brace/ext/static_highlight');
-const highlight = ace.acequire('ace/ext/static_highlight');
+const highlight  = require('ace-builds/src-min-noconflict/ext-static_highlight');
 
-require('brace/mode/json');
-const JavaScriptMode = ace.acequire('ace/mode/json').Mode;
+require('ace-builds/src-min-noconflict/mode-json');
+const JavaScriptMode = require('ace-builds/src-min-noconflict/mode-javascript').Mode;
```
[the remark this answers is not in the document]
rhyshort: fixed.

Antonio-Maranhao (approved):

PR #1362, opened 2022-08-01
Fix wrong config section for `enable_cors`

big-r81:
Since CouchDB 3.2 `enable_cors` moved from section `[httpd]` to
section `[chttpd]`. Fixes apache/couchdb-pkg#96


nickva (approved):
+1

Thanks for fixing it @big-r81 !

PR #1431, opened 2024-03-24
Optimize getParams + Prettify utils

RedYetiDev:
## Overview
This PR optimizes the `getParams` function to prevent the use of needless statements (such as while loops). With less statements, the function will block the event loop less, and run smoother.
> Note: The above is the primary reason for this PR. While I was implementing this, I also ran the code through the [prettier](https://tinyurl.com/5hbkwxaa) tool.

## Testing recommendations

This function should not require any changes to the original codebase, and will not provide any front-end change. To test these, try running it with varying formats of query parameters.

## GitHub issue number

This fixes no issues, but does end up resolving the comment `// I think this could be combined into one if`.

## Related Pull Requests

N/A

## Checklist

> This shouldn't require any changes to existing tests, but I've left this unchecked until you have confirmed
> This shouldn't require any changes to existing documentation, but I've left this unchecked until you have confirmed

### DIFF
--- a/app/core/utils.js
+++ b/app/core/utils.js
@@ -17,131 +17,119 @@
 // want to change this later, but for now this should be thought of as a
 // "purely functional" helper system.
 
-
 import _ from 'lodash';
 import param from 'jquery-param-fn';
 
 const utils = {
+	getParams: function (queryString) {
+		const getUrlParams = (str) => {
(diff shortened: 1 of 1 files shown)


RedYetiDev:
Hi team, any update?

PR #1396, opened 2023-04-09
docs: Mastodon Information to replace twitter

reconbot:
## Overview

Updating the docs to have mastodon and remove twitter.

image

PR #1435, opened 2024-05-21
Fix checksDocsPage tests

big-r81:
- update GH actions to remove deprecation warning
- fix Twitter -> Mastodon link #1396 
- update links to use https


jaydoane (approved):
I would probably put these changes into three separate commits, but otherwise LGTM 👍

PR #1482, opened 2025-06-27
Fix empty _rev parameter from dropdown select list

big-r81:
If there are multiple conflict versions in the dropdown field, the http get request was called with an empty rev parameter and failed. Use the right child object to get the value.

Example:

Before:
GET /b/1?rev= HTTP/1.1

After:
GET /b/1?rev=1-03a8cbc32975d2f58295a1ea2617d965 HTTP/1.1


big-r81:
@janl friendly nudging. The question is, if `target.value` is the correct child object for this. It seems to work ...

PR #1525, opened 2026-04-24
Remove uuid package

Antonio-Maranhao:
## Overview

Fauxton only uses v4 uuid which can be generated with the built-in crypto.randomUUID(), so using the external uuid package is no longer necessary.
Also, the uuid package has been ESM-only since v12 which causes problems with the Jest tests.
Removing it addresses https://github.com/advisories/GHSA-w5hq-g745-h8pq.

## Testing recommendations

No change in behavior so it's enough the CI tests pass.

## GitHub issue number

n/a

## Related Pull Requests

n/a

## Checklist


fryjordan:
LGTM!

PR #1527, opened 2026-04-27
fix: crypto.randomUUID() not available in insecure envs

Antonio-Maranhao:
## Overview

Add an alternative implementation for when crypto.randomUUID() is not available in the environment. This can happen because crypto.randomUUID() is only available in the browser under a secure context (HTTPS or localhost), so if Fauxton is accessed via HTTP the function won't exist and cause the app to fail.

## Testing recommendations

No change in behavior - should be enough that CI passes.

## GitHub issue number

n/a

## Related Pull Requests

n/a

## Checklist

### DIFF
--- a/app/core/__tests__/api.test.js
+++ b/app/core/__tests__/api.test.js
@@ -42,3 +42,13 @@ describe('URLs', () => {
   });
 
 });
+
+describe('generateUUID', () => {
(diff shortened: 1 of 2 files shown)


fryjordan:
LGTM!

